PR #14, opened 2017-07-08
tools: Add virtualenv support.

aero31aero:
Testing instructions: I recommend doing a fresh clone just in case there are conflicts otherwise:

```sh
git clone https://github.com/aero31aero/python-zulip-api.git test-repo`
cd test-repo
git checkout virtualenv
source tools/provision
# Installation should be complete
bots_api/test-bots
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# virtualenv
+env
+
 *.py[cod]
 
 # C extensions
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,7 @@ nosetests.xml
 .mr.developer.cfg
 .project
 .pydevproject
+
+# mypy
+mypy-cache
+
--- a/tools/provision
+++ b/tools/provision
@@ -1,0 +1,15 @@
+#! /usr/bin/env bash
+
+echo "Setup: Setting up Virtual Env"
+
+BASEDIR=`dirname $0`/..
+
+if [ ! -d "$BASEDIR/env" ]; then
+    virtualenv $BASEDIR/env --no-site-packages
+    echo "Virtualenv created."
+fi
+
+source $BASEDIR/env/bin/activate
+python $BASEDIR/zulip_bots/setup.py install
+python $BASEDIR/zulip_api/setup.py install
+echo "Provisioning Complete."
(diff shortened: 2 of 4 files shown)


tools/provision
```diff
@@ -0,0 +1,15 @@
+#! /usr/bin/env bash
+
+echo "Setup: Setting up Virtual Env"
+
+BASEDIR=`dirname $0`/..
+
+if [ ! -d "$BASEDIR/env" ]; then
+    virtualenv $BASEDIR/env --no-site-packages
```
rht: If `env`'s use case is equivalent to `node_modules`, then it should be at `$BASEDIR`, but [pipenv](https://github.com/kennethreitz/pipenv) puts project-specific envs at `~/.local/share/virtualenvs/` with an autogenerated virtualenv name. I don't know why it is this way in pipenv, but I guess this has the advantage that no relative pathing is required, should the `activate` script need to be accessed by another user, e.g. `zulip` or a bot process.

aero31aero: I used a local virtualenv here since this is only for the developers to develop in an isolated environment so that they do not clutter their system and in general for a clean environment. It is not meant for use by external scripts.

The users would install the packages using `pip install zulip-api` and `pip install `zulip-bots`. Here, users includes any part in the Zulip server that needs to run a bot or its scripts.

aero31aero:
@timabbott rebased this PR with master.

Pinging @eeshangarg @abhijeetkaur for testing. This probably wouldn't take much time.

.gitignore
```diff
@@ -31,3 +34,6 @@ nosetests.xml
 .mr.developer.cfg
 .project
 .pydevproject
+
+# mypy
+mypy-cache
```
derAnfaenger: Git complains here about a missing newline. Not sure what to do about it though.

aero31aero: Ah. Yes, I should have ended the file with a newline. Check back in a few minutes.

aero31aero:
This works on my machine from anywhere, as in running `tools/provision` or even `gitclones/python-zulip-api/tools/provision`. Could you add an `echo $BASEDIR` to the script and tell the output?

aero31aero:
Superceded by #52 . @eeshangarg take note that the provision script was failing on Windows, as pointed out above, because of some inconsistencies in the POSIX implementation perhaps. I wasn't able to set the $BASEDIR variable properly there.

PR #15, opened 2017-07-10
zulip-api: Package our bots and API bindings for PyPI.

eeshangarg:


eeshangarg:
@neiljp: It does, but you can run specific tests using `python -m unittest` too, just have to pass in the directory where the tests for a particular bot live (for example `python -m unittest discover bots/followup`). We don't need to do all of the path handling and argument handling ourselves when `unittest` already does it for us. I'll be refining the documentation for future contributors to make it easier to find out how to run specific tests.

eeshangarg:
Also, using a well-tested, widely-used test suite is always better than using scripts that only work for a subset of the tests. `unittest discover` finds all tests within a package recursively!

bots/john/assets/var/jokes.json
```diff
@@ -1,86 +0,0 @@
-[
-   {
-      "joke":"Did you hear about the guy whose whole left side was cut off? He's all right now."
```
aero31aero: Was this file intentionally removed?

eeshangarg: Actually I am not sure. I didn't remove this, I just renamed it (like I did everything else). I don't know why it is showing up as a removal.

eeshangarg: I think this is one of GitHub's quirks, this has happened before.

zulip_bots/setup.py
```diff
@@ -0,0 +1,80 @@
...
+        'Development Status :: 4 - Beta',
+        'Environment :: Web Environment',
+        'Intended Audience :: Developers',
+        'License :: OSI Approved :: MIT License',
+        'Topic :: Communications :: Chat',
+    ],
+    url='https://www.zulip.org/',
+    entry_points={
+        'console_scripts': [
+            'zulip-bot-server=server.bot_server:main',
```
aero31aero: To Discuss: Should we have `run.py` be exposed similarly?

eeshangarg: We already do, I am pretty sure

eeshangarg: Okay, yes I did, its right below line 25

bots/john/assets/var/jokes.json
aero31aero: I'm pretty sure its not there in the PR. I can't seem to find it locally.

zulip_bots/setup.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from __future__ import print_function
+
+import sys
+
+# We should be installable with either setuptools or distutils.
+package_info = dict(
+    name='zulipbots',
```
aero31aero: `s/zulipbots/zulip-bots` ?

eeshangarg: I know I am not sure what we should name it yet. I think I'll wait on feedback from everyone else and see what's the general consensus on name choices. :) Thanks for pointing this out!

zulip_bots/setup.py
aero31aero: Yeah I was viewing th partial diff. Found it.

bots/john/assets/var/jokes.json
eeshangarg: Cool, I'll take a look. Thanks for pointing this out! :)

zulip_api/setup.py
```diff
@@ -59,28 +59,22 @@ def recur_expand(target_root, dir):
...
 
 setuptools_info = dict(
     install_requires=['requests>=0.12.1',
                       'simplejson',
                       'six',
                       'typing>=3.5.2.2',
-                      'flask>=0.12.2',
-                      'mock>=2.0.0',
-                      # for pep8 linter
-                      'pycodestyle==2.3.1',
```
aero31aero: This line needs to be kept.

aero31aero: This dependency is for the pep8 linter and I figure we should move this code outside of both the packages soon, but perhaps not in this PR.

eeshangarg: No I think we should move it to `.travis`, it isn't a dependency for the package, only for the repo

eeshangarg: I'll take care of that. :)

bots/john/assets/var/jokes.json
timabbott: Yeah, this is deleted in 240ee77bfc07149e18023e3afcaa168300f1996e.  It's really easy to see if you use `gitk` or another graphical viewer to inspect these commits.

timabbott:
Do we need it to be `zulip_bots.api.lib`, etc?  I feel like if we put the bots in a subdirectory, we probably don't need the library code to be in one as well.

timabbott:
I have a few concerns with this branch:

```
    bots: Stop provisioning bot dependencies.
    
    So far, there are only two bots with dependencies (define and
    thesaurus). Since there are not many bots that require extra
    bot-specific dependencies, we now specify these in setup.py so
    that they are installed as part of the package. This move
    obsoltes bots_api/provision.py and the --provision option for
    bots_api/run.py.
```
This change seems wrong to me; we have other bots with dependencies (e.g. the GitHub bot, etc.); they just don't have tests yet and thus didn't have declared dependencies yet.  I think it'd probably be good to keep the provisioning code for situations where we find we have a third-party dependency that we don't want to add to the main package.  I think it's fine to get rid of the existing `html2text` user case, though.

```
    bots_api: Remove test-bots script.
    
    Since there is no "provisioning" of dependencies and no manual
    modifications of sys.path in order to test bots, we can now
    use 'python -m unittest discover' to run all tests.
```

I would prefer to keep our own test script, since we're inevitably going to want to extend our tooling to do more things (test coverage, etc.), and I don't `python -m unittest discover` is hard to remember, whereas `tools/test-bots` is easy to remember.  (Also, `python -m unittest discover` doesn't print the test names, and thus has much harder to read output than it should).  I think we can put the code under `tools/`, since we probably wouldn't want/need to distribute the test runner as part of one of the packages.  

A few more small issues from reading the diff:
* The root `README.md` is inaccurate.
* `jokes.json` seems to have been deleted accidentally?

I think once those issues are resolved, we should do a bit of `git` work to make the history sensible (quite possibly just squashing things, but at the various least we shouldn't move all the files around and then back again...) and merge this.  

As a more general note on this, the way that you've combined rearranging code with functional changes make it very difficult to review and merge incrementally.

timabbott:
This looks great.  Merged, thanks @eeshangarg!!  Thanks to everyone who helped review this!

I think there's a few things that I'm not sure about (e.g. does that `MANIFEST.in` configuration mean that the test fixtures are part of the package that gets shipped?  That seems wrong).  Let's do a few rounds of smaller incremental fixups this week and plan to do a `pip` package release early next week.  

@aero31aero @abhijeetkaur @vabs22 @neiljp FYI.  If any of you has time, I think it'd be great to work on adding test fixtures for the remaining bots that we have tests for, so that the test suite isn't so slow.  With fixtures, it should take less than a second to run all of our tests.

abhijeetkaur:
@timabbott I'd like to take up the work of adding fixtures for bots.

PR #16, opened 2017-07-16
Zulip bot server fix for handling bots without handler class.

vabs22:
There can be cases where flaskbotrc contains config details of a bot, but the user hasn't added it's handler class code. This fixes server to handle such cases, by reporting it to user.

### DIFF
--- a/zulip_botserver/zulip_botserver/server.py
+++ b/zulip_botserver/zulip_botserver/server.py
@@ -34,20 +34,29 @@ def load_lib_modules():
     # type: () -> None
     for bot in available_bots:
         try:
-            module_name = 'zulip_bots.{bot}.{bot}'.format(bot=bot)
-            bots_lib_module[bot] = import_module(module_name)
+            module_name = 'zulip_bots.bots.{bot}.{bot}'.format(bot=bot)
+            lib_module = import_module(module_name)
+            bots_lib_module[bot] = lib_module
         except ImportError:
             print("\n Import Error: Bot \"{}\" doesn't exists. Please make sure you have set up the flaskbotrc "
                   "file correctly.\n".format(bot))
-            sys.exit(1)
+
+def get_bot_lib_module(bot):
+    # type: (str) -> Any
+    if bot in bots_lib_module.keys():
+        return bots_lib_module[bot]
+    return None
 
 app = Flask(__name__)
 
 @app.route('/bots/<bot>', methods=['POST'])
 def handle_bot(bot):
     # type: (str) -> Union[str, BadRequest]
-    if bot not in available_bots:
-        return BadRequest("requested bot service {} not supported".format(bot))
+    lib_module = get_bot_lib_module(bot)
+    if lib_module is None:
+        return BadRequest("Can't find the configuration or Bot Handler code for bot {}. "
+                          "Make sure that the `zulip_bots` package is installed!".format(bot))
+
     client = Client(email=bots_config[bot]["email"],
                     api_key=bots_config[bot]["key"],
                     site=bots_config[bot]["site"])
@@ -58,7 +67,8 @@ def handle_bot(bot):
     except SystemExit:
         return BadRequest("Cannot fetch user profile for bot {}, make sure you have set up the flaskbotrc "
                           "file correctly.".format(bot))
-    message_handler = bots_lib_module[bot].handler_class()
+
+    message_handler = lib_module.handler_class()
 
     # TODO: Handle stateful bots properly.
     state_handler = StateHandler()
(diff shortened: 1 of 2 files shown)


vabs22:
@timabbott @robot-dreams please take a look.

timabbott:
I think we should do this after #15 is merged, since that reorganizes a lot of code.

vabs22:
@timabbott I have updated this PR, please take a look.

zulip_botserver/zulip_botserver/server.py
```diff
@@ -56,7 +62,11 @@ def handle_bot(bot):
     except SystemExit:
         return BadRequest("Cannot fetch user profile for bot {}, make sure you have set up the flaskbotrc "
                           "file correctly.".format(bot))
-    message_handler = bots_lib_module[bot].handler_class()
+
+    lib_module = get_bot_lib_module(bot)
+    if lib_module is None:
+        return BadRequest("Cannot fetch Bot Handler Class for bot {}.".format(bot))
```
timabbott: How is this check different from the `available bots` check?

robot-dreams: It looks like bots that weren't in `zuliprc` will fail the `available_bots` check above, but bots that had a misconfigured `zuliprc` will fail this check.  However, I don't think this distinction is necessary; for the sake of simplicity, would it make sense to remove the `available_bots` check and move this `lib_module` stuff to the top of the function?

vabs22: Yeah this can be done. I am on it.

timabbott:
@eeshangarg @derAnfaenger can you review this one?

zulip_botserver/zulip_botserver/server.py
```diff
@@ -33,18 +33,29 @@ def load_lib_modules():
...
 app = Flask(__name__)
 
 @app.route('/bots/<bot>', methods=['POST'])
 def handle_bot(bot):
     # type: (str) -> Union[str, BadRequest]
+    lib_module = get_bot_lib_module(bot)
+    if lib_module is None:
+        return BadRequest("Either the configuration for bot {} is missing or the Bot Handler Class cannot be "
+                          "fetched.".format(bot))
+
```
eeshangarg: On another note, I think the error message passed to `BadRequest` is misleading. Not being able to import the bots could mean a lot of things (the `zulip_bots` package isn't installed, they misspelled the name of the bot on the command-line, etc), but so far, this error doesn't have anything to do with finding the handler class. It has more to do with the fact that it can't find the source code for the bot to begin with. So we should mention that in the error message plus recommend making sure that the `zulip_bots` package is installed. Maybe, like so:

```
BadRequest("Can't find the configuration for bot {}. Make sure that the `zulip_bots` package is installed!".format(bot)))
```
or something! I mean if we can't find the whole module to begin with, there's no point in mentioning a handler class. So, I have a feeling it could confuse people not familiar with how the bots are written. Let me know what you think! :)

vabs22: @eeshangarg Thanks for reviewing this. There exists a similar test case for this, I will refactor it. :)

The main motivation for this was to fix the case where there exists a bot which works purely as an outgoing webhook bot i.e. posts data to a third party URL but doesn't have the bot code. Hence the bot handler couldn't be fetched and zulip-bot-server would have failed. Hence I added the check and I tried to convey the possible info through the message. But I like your idea, and I think error message such as above would convey the most info about the error. 
`BadRequest("Can't find the configuration or bot code for bot {}. Make sure that the `zulip_bots` package is installed!".format(bot)))`  
What do you think?

zulip_botserver/zulip_botserver/server.py
```diff
@@ -65,7 +74,8 @@ def handle_bot(bot):
     message_handler.handle_message(message=event["message"],
                                    bot_handler=restricted_client,
                                    state_handler=state_handler)
-    return "Success!"
+    response_json = json.dumps(dict(response_not_required=True))
+    return response_json
```
timabbott: @vabs22 is this still the right way to send responses to the outgoing webhooks system?

timabbott:
I posted one quick comment; otherwise, this looks good!

zulip_botserver/zulip_botserver/server.py
vabs22: @timabbott this is would work, although it is a bit redundant. Returning `json.dumps("")` should also work. However we could stick with this as it is more readable.

timabbott:
Merged, thanks @vabs22!

vabs22:
thank you @timabbott

PR #17, opened 2017-07-18
bots: Remove common send-reply function.

abhijeetkaur:
Previously, the annotation for the (now removed)send-reply
function was also wrong, as "send-message" function of
EmbeddedBotHandler class does not return any value, contrary
to "Dict[str, Any]" as specified by the funtion.

This is done as a follow up for #5842.


timabbott:
Merged, thanks @abhijeetkaur!

Issue #3, opened 2017-07-06
Use request sessions in Python API.


showell:
See https://github.com/zulip/zulip/issues/4509

dehnert:
@zulipbot claim

zulipbot:
Congratulations, @dehnert, you've successfully claimed your first issue! Contributing to Zulip will (hopefully) be a fun and rewarding experience, and I'm here to help get you off to a good start.

First things first:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](http://zulip.readthedocs.io/en/latest/index.html#code-docs), as well as the rest of the developer documentation on that site.


See you on the other side (that is, the pull request side)!

zulipbot:
Hello @, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 7 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 3 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

PR #18, opened 2017-07-19
api: Use requests.Session

dehnert:
Using requests.Session allows the requests library to reuse HTTP connections,
which is potentially helpful for performance. This fixes python-zulip-api #3.

Some notes:
* I'm currently lazily initializing the Session object. This has the advantage if somebody constructs a client and then modifies some field before making any requests, it should work as they probably expect. It has the disadvantage of probably not being threadsafe (although the client as a whole might not be either), not to mention whatever style views people have.
* It appears to pass lint. AFAIK, that's it for "testing" the API -- I think the bot tests mock it out? (But those tests do pass.)
* I ran `mypy` (though not the version that Zulip has pinned). Beyond some warnings about stubs that I think are preexisting, I get some errors about request.Session's field typing. I think typeshed is just wrong; https://github.com/python/typeshed/issues/1093 certainly does not leave me with much faith in the stubs. Comparing [Session's stubs](https://github.com/python/typeshed/blob/master/third_party/2and3/requests/sessions.pyi) to [.request's stubs](https://github.com/python/typeshed/blob/master/third_party/2and3/requests/api.pyi) the former is certainly more restrictive, and I don't believe the stubs' typing for [verify](http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification) (which really does seem to take a string) or auth (which does seem to take [an object](http://docs.python-requests.org/en/master/user/advanced/#custom-authentication) -- though in fairness it looks like passing a (user,pass) tuple should work too, and the stubs support it). I'm not sure what to do about this, especially given that the API seems annotated but not currently being tested.

I think this is sane to merge, although you might reasonably want me to change how I handled these notes (or have other objections).

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -285,6 +285,30 @@ class Client(object):
         self.client_cert = client_cert
         self.client_cert_key = client_cert_key
 
+        self.session = None  # type: Union[None, requests.Session]
+
+    def ensure_session(self):
+        # type: () -> None
+
+        # Check if the session has been created already, and return
+        # immediately if so.
+        if self.session:
+            return
+
+        # Build a client cert object for requests
+        if self.client_cert_key is not None:
+            client_cert = (self.client_cert, self.client_cert_key)  # type: Union[str, Tuple[str, str]]
+        else:
+            client_cert = self.client_cert
+
+        # Actually construct the session
+        session = requests.Session()
+        session.auth = requests.auth.HTTPBasicAuth(self.email, self.api_key)  # type: ignore # https://github.com/python/typeshed/pull/1504
+        session.verify = self.tls_verification  # type: ignore # https://github.com/python/typeshed/pull/1504
+        session.cert = client_cert
+        session.headers = {"User-agent": self.get_user_agent()}
+        self.session = session
+
     def get_user_agent(self):
         # type: () -> str
         vendor = ''
@@ -326,6 +350,8 @@ class Client(object):
 
         for f in files:
             req_files.append((f.name, f))
+
+        self.ensure_session()
 
         query_state = {
             'had_error_retry': False,
@@ -370,21 +396,10 @@ class Client(object):
                 if files:
                     kwargs['files'] = req_files
 
-                # Build a client cert object for requests
-                if self.client_cert_key is not None:
-                    client_cert = (self.client_cert, self.client_cert_key)  # type: Union[str, Tuple[str, str]]
-                else:
-                    client_cert = self.client_cert
-
-                res = requests.request(
+                res = self.session.request(
                     method,
                     urllib.parse.urljoin(self.base_url, url),
-                    auth=requests.auth.HTTPBasicAuth(self.email,
-                                                     self.api_key),
-                    verify=self.tls_verification,
-                    cert=client_cert,
                     timeout=90,
-                    headers={"User-agent": self.get_user_agent()},
                     **kwargs)
 
                 # On 50x errors, try again after a short sleep


zulip/zulip/__init__.py
```diff
@@ -327,6 +351,8 @@ def do_api_query(self, orig_request, url, method="POST", longpolling=False, file
         for f in files:
             req_files.append((f.name, f))
 
+        self._init_session()
```
timabbott: maybe `self.ensure_session()` would be a better name here?

timabbott:
The code looks good; posted one quick comment.  Some thoughts on your notes:
* I think threadsafety is something I'd rather handle via locking/whatever if we find a use for it.  
* For typeshed, I'd open an issue in Typeshed and use `# type: ignore # link to issue` to silence the warnings in our code and reference it so it's clear why and when we can clean it up.  We'll be type-checking the API again eventually.
* Correct, this has no real test suite in this project yet (there's a bit more of one in the server project in the form of `tools/test-api` there).  Might be worth running that against this project, because I think there's a good chance that its mocking strategy won't work with this.

dehnert:
Thanks for the feedback. I'm traveling this weekend, but I'll try to deal early next week. 

On July 20, 2017 4:07:54 PM PDT, Tim Abbott <notifications@github.com> wrote:
>The code looks good; posted one quick comment.  Some thoughts on your
>notes:
>* I think threadsafety is something I'd rather handle via
>locking/whatever if we find a use for it.  
>* For typeshed, I'd open an issue in Typeshed and use `# type: ignore #
>link to issue` to silence the warnings in our code and reference it so
>it's clear why and when we can clean it up.  We'll be type-checking the
>API again eventually.
>* Correct, this has no real test suite in this project yet (there's a
>bit more of one in the server project in the form of `tools/test-api`
>there).  Might be worth running that against this project, because I
>think there's a good chance that its mocking strategy won't work with
>this.  
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/zulip/python-zulip-api/pull/18#issuecomment-316855358

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

timabbott:
Cool

timabbott:
@dehnert quick ping on this :)

dehnert:
I fixed these comments last night, except that I haven't been able to run `test-api` yet -- my Zulip proper environment (without these changes) is failing to run the tests. On the typeshed front, I submitted https://github.com/python/typeshed/pull/1504 upstream and referenced it from the code.

timabbott:
@eeshangarg do you have time to quickly run `test-api` against an updated version `python-zulip-api` that contains this code?

rht:
[Passed in Travis](https://travis-ci.org/rht/zulip/builds/262537254) ! (I
suppose a fresh dev build doesn't have the hysteresis of accumulated errs from
past experimentation.)

Tried porting a subset of the backend test in zulip/zulip
https://github.com/rht/python-zulip-api/commit/2f967458d6171f1abb0ceb205669937aa558fbeb
just enough to run `./tools/test-api` but to no avail since the
`./tools/provision` from the zulip/zulip repo would still fetch the old version
instead of the hash being tested. So I treaded the path of least loc-patch by
patching the requirements in zulip/zulip.

timabbott:
@rht for the zulip/zulip issue, you just need to update `requirements.txt` to point to the Git URL of this branch in the relevant fork, and then you can provision against this version.

timabbott:
Merged, thanks @dehnert!  I think it'll be easier to test by hand once it's merged and we can just include it in the server repo.  It seems pretty unlikely that it'll fail in a nontrivial way.

dehnert:
Thanks folks!

PR #19, opened 2017-07-19
bots: Add test fixture files for bots.

abhijeetkaur:


abhijeetkaur:
Added test-fixtures for all the bots with already existing tests. 

Adding a separate json for request/response behavior of the bot encourages people to add more test cases as one does not need to look into the code to see where to add the test samples. Folks can directly open up the test-fixture files and add tests.

zulip_bots/zulip_bots/bots/followup/fixtures/test_send_reply.json
```diff
@@ -0,0 +1,8 @@
+{
+    "expectations":[
+        {
+            "request":"",
+            "response":"Please specify the message you want to send to followup stream after @mention-bot"
+        }
+    ]
+}
```
aero31aero: Add a new line at the end. Ideally the linter should have caught this.

timabbott: I think the linter is probably not running against the JSON files?  We do that in the server repository, but maybe that bit of config wasn't ported over.

timabbott:
@abhijeetkaur the file names for the tests here could be better.  E.g. for `followup_bot`, you have `test_send_message` and `test_send_reply`, but I think those should be `test_successful_request` and `test_empty_request`, since the fact that it's using `send_reply` is unimportant; what's important is that the input is invalid.

timabbott:
I'm also confused by the test fixture strategy here.  What I had been imagining is that we would be adding test fixtures that mock the return values of the network requests (using `mock_http_conversation`, while keeping the actual testing code mostly unchanged.  This appears to have changed the testing code to be split between code and fixtures files, while not actually solving the problem that the tests won't work without Internet.

abhijeetkaur:
@timabbott Oh, I wasn't aware of the primary objective.Well, I have mostly worked on the files not using internet. 
So, we intend to add test-fixtures for only those bots that use the internet, right?

timabbott:
@abhijeetkaur I'm going to go ahead and close this; thanks again for your work on this and also your more recent work on adding fixtures for all the network connections in our bot tests.  The tests are a lot faster now, and my hope is that with a bit more work we can have a test suite we can be proud of.

PR #22, opened 2017-07-22
zulip_bots/run: Support importing bot modules by path.

eeshangarg:
The zulip-run-bot script now supports passing in a --path-to-bot
option. This allows users to specify the path to the source file
for their own custom bots, the first step towards being able to
support out-of-tree bots.

To run an existing bot in the zulip_bots package, just passing in
the name of the bot should suffice.

### DIFF
--- a/tools/test-bots
+++ b/tools/test-bots
@@ -1,72 +1,80 @@
 #!/usr/bin/env python
 
-from __future__ import absolute_import
-from __future__ import print_function
-
+from os.path import dirname, basename
+from importlib import import_module
 import argparse
 import os
-import sys
+import glob
 import unittest
 import logging
-from unittest import TestCase
 
-def dir_join(dir1, dir2):
-    # type: (str, str) -> str
-    return os.path.abspath(os.path.join(dir1, dir2))
+
+def load_tests_from_modules(names, template):
+    loader = unittest.defaultTestLoader
+    test_suites = []
+
+    for name in names:
+        module = import_module(template.format(name=name))
+        test_suites.append(loader.loadTestsFromModule(module))
+
+    return test_suites
+
+
+def parse_args(available_bots):
+    description = """
+Script to run test_<bot_name>.py files in the
+zulip_bot/zulip_bots/bots/<bot_name> directories.
+
+Running tests for all bots:
+
+./test-bots
+
+Running tests for specific bots:
+
+./test-bots define thesaurus
+
+Running tests for all bots excluding certain bots (the
+following command would run tests for all bots except
+the tests for xkcd and wikipedia bots):
+
+./test-bots --exclude xkcd wikipedia
+"""
+    parser = argparse.ArgumentParser(description=description)
+
+    parser.add_argument('bots_to_test',
+                        metavar='bot',
+                        nargs='*',
+                        default=available_bots,
+                        help='specific bots to test (default is all)')
+
+    parser.add_argument('--exclude',
+                        metavar='bot',
+                        nargs='*',
+                        default=[],
+                        help='bot(s) to exclude')
+    return parser.parse_args()
+
+
+def main():
+    glob_pattern = os.path.abspath('../zulip_bots/zulip_bots/bots/*/test_*.py')
+    test_modules = glob.glob(glob_pattern)
+
+    # get only the names of bots that have tests
+    available_bots = map(lambda path: basename(dirname(path)), test_modules)
+
+    options = parse_args(available_bots)
+
+    bots_to_test = filter(lambda bot: bot not in options.exclude,
+                          options.bots_to_test)
+    test_suites = load_tests_from_modules(
+        bots_to_test,
+        template='zulip_bots.bots.{name}.test_{name}'
+    )
+
+    suite = unittest.TestSuite(test_suites)
+    runner = unittest.TextTestRunner(verbosity=2)
+    runner.run(suite)
 
 
 if __name__ == '__main__':
-
-    tools_dir = os.path.dirname(os.path.abspath(__file__))
-    bots_dir = os.path.join(tools_dir, '..', 'zulip_bots', 'zulip_bots')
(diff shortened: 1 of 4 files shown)


eeshangarg:
I'm working on more improvements, this PR is still a work in progress. I would appreciate everyone's feedback as I push more commits! Thanks! :)

timabbott:
I'm not sure moving from optparse to argparse here is actually a good idea; we probably actually want to be using the `zulip.init_from_options()` mechanism that the `zulip` module exports for its dependencies, which is based on `optparse`.

zulip_bots/zulip_bots/provision.py
```diff
@@ -1,67 +1,85 @@
...
-def run():
-    # type: () -> None
-    usage = '''
-        Installs dependencies of bots in api/bots directory. Add a
-        reuirements.txt file in a bot's folder before provisioning.
-
-        To provision all bots, use:
-        ./provision.py
+        logging.info('Installed dependencies successfully.')
+        sys.path.insert(0, install_path)
```
timabbott: I feel like the `if not force` else clause here is probably not correct.

zulip_bots/zulip_bots/provision.py
```diff
@@ -1,67 +1,85 @@
...
+                        action='store_true',
+                        default=False,
+                        help='Turn off logging output.')
+
+    return parser.parse_args()
+
+
+def main():
+    # type: () -> None
+    bots_dir = os.path.abspath("bots")
```
timabbott: I'm guessing this tool probably doesn't work correctly unless the current working directly is "correct".  Probably a task for a follow-up commit, but we probably want it to find the paths of the files it references based on relative paths.  Scripts in `zulip.git` have lots of examples of doing this correctly.

tools/test-bots
```diff
@@ -1,72 +1,80 @@
...
-        except ImportError:
-            logging.error("Bot %s requires __init__.py to be added" % (bot_to_test))
-            failed = True
-
-    suite = unittest.TestSuite(suites)
+
+    parser.add_argument('--exclude',
+                        metavar='bot',
+                        nargs='*',
+                        default=[],
```
showell: Any reason not to use a set() here?

zulip_bots/zulip_bots/run.py
```diff
@@ -3,12 +3,33 @@
...
 
 from zulip_bots.lib import run_message_handler_for_bot
+from zulip_bots.provision import provision_bot
+
+
+def import_module_from_source(path, name=None):
+    if name is None:
+        name = splitext(basename(path))[0]
+
+    if six.PY2:
```
showell: Did we make a firm decision yet to go to python3 only for bots?  If so, we can eliminate the PY2 stuff.

zulip_bots/zulip_bots/run.py
```diff
@@ -24,31 +45,61 @@ def parse_args():
...
-    (options, args) = parse_args()
-    bot_name = args[0]
+    options = parse_args()
+
+    if options.path_to_bot:
+        if options.provision:
+            bot_dir = os.path.dirname(os.path.abspath(options.path_to_bot))
+            provision_bot(bot_dir, options.force)
+        lib_module = import_module_from_source(options.path_to_bot, name=options.name)
+    else:
```
showell: You can probably use `elif` for now here.  A short comment explaining why these options are mutually exclusive would be helpful, too.

tools/test-bots
eeshangarg: Well, any arguments presented after `--exclude` will eventually be gathered into a list, according to `argparse's` docs, so I didn't see any point in using `set()`. Should I still?

zulip_bots/zulip_bots/run.py
eeshangarg: I'm not sure actually, I started a discussion in `#integrations` once but it didn't lead to any firm conclusions on this matter! I feel like I remember Tim mentioning that we definitely want the `zulip` API package to support Python 2. And in a past discussion, someone mentioned that we want to move the bots API to the `zulip` API package as well. If we end up doing that, then this code will have to support Python 2. I think this definitely warrants a longer discussion though.

zulip_bots/zulip_bots/provision.py
eeshangarg: Correct me if I am wrong, but since the default value of `force` is `False`, that means that if there are errors, then the `provision.py` will `sys.exit` *before* it adds the `intall_dir` to `sys.path`.

If `force` is `True` and if there are pip errors, then it will move on and add `install_dir` to `sys.path` anyway. Isn't that how it should be? Or am I missing something here and maybe there's a better way?

zulip_bots/zulip_bots/run.py
eeshangarg: Done! :)

zulip_bots/zulip_bots/run.py
timabbott: I think we should plan to support both.  Very few organizations do Python 3 yet, and this is classically the type of code you want to be able to run anywhere.

zulip_bots/zulip_bots/provision.py
timabbott: We shouldn't print that we installed the dependencies successfully if we didn't but ignored errors due to `--force`.

tools/test-bots
showell: You can't say `default={}`?

timabbott: I could be wrong on this, but I think structurally, optparse produces a list; the default is only used if the argument isn't passed.  So if we want a set, I think we need to do that in post-processing.

timabbott:
Nice cleanup.  Merged, thanks @eeshangarg!

tools/test-bots
showell: Oh, I understand now.

PR #24, opened 2017-07-23
Add custom regular expressions to the linter.

derAnfaenger:
`custom_check.py` is taken from the `zulip` repository. It has been pruned to not include rules superfluous for this repo, like js linting, etc.


derAnfaenger:
@sampritipanda FYI

tools/server_lib/lister.py
```diff
@@ -66,7 +66,7 @@ def list_files(targets=[], ftypes=[], use_shebang=True, modified_only=False,
     # sys.argv as str, so that battle is already lost.  Settle for hoping
     # everything is UTF-8.
     repository_root = subprocess.check_output(['git', 'rev-parse', '--show-toplevel']).strip().decode('utf-8')
-    exclude_abspaths = [os.path.join(repository_root, fpath).rstrip('/') for fpath in exclude]
+    exclude_abspaths = [os.path.normpath(os.path.join(repository_root, fpath).rstrip('/')) for fpath in exclude]
```
timabbott: Is this a change we should also submit for the server repository?

tools/custom_check.py
```diff
@@ -0,0 +1,279 @@
...
+    '''},
+        # Directly fetching Message objects in e.g. views code is often a security bug.
+        {'pattern': '[^r][M]essage.objects.get',
+         'description': 'Please use access_message() to fetch Message objects',
+         },
+        {'pattern': 'render_to_response\(',
+         'description': "Use render() instead of render_to_response().",
+         },
+        # This rule might give false positives in virtualenv setup files which should be excluded,
+        # and comments which should be rewritten to avoid use of "python2", "python3", etc.
```
timabbott: I think the ~10 rules from here up (first one is `json_success`) are all unnecessary in the bots repo and we should perhaps remove.

tools/custom_check.py
```diff
@@ -0,0 +1,279 @@
...
+        {'pattern': '((?<!\s)\s$)|(\s\s\s+$)|(^\s+$)',
+         'strip': '\n',
+         'description': 'Fix trailing whitespace'},
+        {'pattern': '^#+[A-Za-z0-9]',
+         'strip': '\n',
+         'description': 'Missing space after # in heading'},
+    ]  # type: RuleList
+    python_rules = cast(RuleList, [
+        {'pattern': '^(?!#)@login_required',
+         'description': '@login_required is unsupported; use @zulip_login_required'},
```
timabbott: This decorator doesn't exist in this repo.

timabbott:
I posted a few quick comments on potential improvements.  I think we could make life a bit easier for maintaining this in the future by splitting apart the code from the configuration in the `zerver` project, so that we can just copy in the code file verbatim.

tools/server_lib/lister.py
derAnfaenger: Yup, definitely! I think the only reason I didn't notice any abnormal behavior in the main repo is that malformatted paths don't produce an error. Will make a PR for it.

derAnfaenger: We also don't need the `rstrip` anymore.

timabbott:
Thanks for working on this @derAnfaenger!  I merged the first two commits (After making a small change to fix a lint issue merged in the meantime)

Can you submit a quick PR to the server to fix that `lister.py` Windows issue there?  I figure we might as well fix bugs "upstream".  Medium-term, I think we may want to split the linter framework to be its own project included by both as a third-party tool, but that can wait.

I didn't merge the third commit because I want some opinions on whether we want to do that.  My temptation would be not to, since technically that's part of the shared tools in the server project, not the `zerver` sub-project.  But I'd like to hear thoughts from @showell and @eeshangarg on it.

showell:
For the third commit, I don't like the name `zerver_lib`, because, as @timabbott pointed out, `zerver` has a more narrow meaning than "server."

timabbott:
OK closing this under the assumption that we're not merging the last commit, though happy to keep discussion :)

derAnfaenger:
Haha, the last commit wasn't too serious anyways :P

PR #25, opened 2017-07-23
Bot testing: Add state testing & use in tictactoe bot

neiljp:
I'm aware some changes have happened or are ongoing in the test infrastructure, but here's a straightforward port (with minor adjustments) of zulip/zulip#5624 from the main repo.

I've ported this as best as I can, based on the previously-working code, but couldn't get the tests to actually run (for any bot) so can't test directly - will test-bots be replaced?

**update**: I've got the tests running now, but I definitely miss the simplicity of the test-bots script.

PR #26, opened 2017-07-23
Bots: Add optional contextmanager for StateHandler

neiljp:
**[This is a port of zulip/zulip#5671 from the main zulip repo]**

This PR simplifies the use of state-handling (following a suggestion in zulip/zulip#5653) and looks pretty clean to me, after having adjusted the existing stateful bots.

The default (initial value) parameter is relevant to all bots, so seemed fairly useful to include. Not all the stateful bot commits need be included, but at least two (virtual_fs and tictactoe) simply get state at the start and set it at the end, which makes it equivalent by my understanding.

As per the commit message for the lib commit, this is an additional feature, so the existing functions can still be used; equally the contextmanager can be used to identify prior state or 'flush' state manually within the 'with' block.

Does this approach seem reasonable @timabbott? (since it was your comment motivating this)

### DIFF
--- a/zulip_bots/zulip_bots/bots/incrementor/incrementor.py
+++ b/zulip_bots/zulip_bots/bots/incrementor/incrementor.py
@@ -12,18 +12,16 @@ class IncrementorHandler(object):
         '''
 
     def handle_message(self, message, bot_handler, state_handler):
-        state = state_handler.get_state() or {'number': 0, 'message_id': None}
-        state['number'] += 1
-        state_handler.set_state(state)
-        if state['message_id'] is None:
-            result = bot_handler.send_reply(message, str(state['number']))
-            state['message_id'] = result['id']
-            state_handler.set_state(state)
-        else:
-            bot_handler.update_message(dict(
-                message_id = state['message_id'],
-                content = str(state['number'])
-            ))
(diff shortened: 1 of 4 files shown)


timabbott:
This looks great to me as an improved API.  We should maybe do some further thinking about the API's name as a group; can you bring that up at our next meeting?

I'm merging now, but I hope we add tests for a few of these bots soon!

neiljp:
@timabbott Glad it looks ok; if I'm here I'll bring it up at the meeting.

The tests are pending something like #25; is that good, or were you waiting for more tests in that PR before committing it?

timabbott:
Either way works.  If you get a chance to add more tests, go for it, but
I'll probably merge when I get a chance

-Tim Abbott (mobile)

On Jul 24, 2017 7:16 PM, "Neil Pilgrim" <notifications@github.com> wrote:

@timabbott <https://github.com/timabbott> Glad it looks ok; if I'm here
I'll bring it up at the meeting.

The tests are pending something like #25
<https://github.com/zulip/python-zulip-api/pull/25>; is that good, or were
you waiting for more tests in that PR before committing it?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/zulip/python-zulip-api/pull/26#issuecomment-317608755>,
or mute the thread

PR #27, opened 2017-07-24
bot testing: Improve failure output & add testing of multiple responses

neiljp:
This is a (further-reduced) port of zulip/zulip#5587 in the main zulip repo, but focused on only testing multiple responses from single messages. For example, a message might cause the bot to respond to the sender with a confirmation message, a group message and even one to one (or more) admins/maintainers/moderators. It originated from wanting to test the 'pollbot' I wrote (not yet merged, zulip/zulip#5426 in main zulip tracker).

As it stands, the PR should enable all bots to continue to pass tests, though the testing does pass through the new/adjusted code. As such, I've only extended the interface for testing bots to allow testing multiple responses, but if this code is considered suitable then it seems reasonable to switch to switching from `assert_bot_response` to `assert_bot_responses`, and extending `check_expected_responses` in a similar way.

The error reporting is hopefully also much improved and was motivated by wanting to determine which message triggered the failure and in what way - and there are now multiple ways in which it can go wrong.

Once this receives attention or is merged then I'll look at working on testing of return values from send_* functions, which was the other feature I included in zulip/zulip#5587.

### DIFF
--- a/zulip_bots/zulip_bots/test_lib.py
+++ b/zulip_bots/zulip_bots/test_lib.py
@@ -11,7 +11,7 @@ import mock
 import requests
 import unittest
 
-from mock import MagicMock, patch
+from mock import MagicMock, patch, call
 
 from zulip_bots.lib import StateHandler
 import zulip_bots.lib
@@ -95,20 +95,42 @@ class BotTestCaseBase(TestCase):
                 # For calls with send_message, r is already a Dict.
                 message = dict(message_template, content = m)
                 response = {'content': r} if expected_method == 'send_reply' else r
-                self.assert_bot_response(message=message, response=response,
-                                         expected_method=expected_method)
+                self.assert_bot_responses(message, (response, expected_method))
 
-    def call_request(self, message, expected_method, response):
-        # type: (Dict[str, Any], str, Dict[str, Any]) -> None
-        # Send message to the concerned bot
-        self.message_handler.handle_message(message, self.mock_bot_handler)
+    def call_request(self, message, *responses):
+        # type: (Dict[str, Any], *Tuple[Dict[str, Any], str]) -> None
 
-        # Check if the bot is sending a message via `send_message` function.
-        # Where response is a dictionary here.
-        if expected_method == "send_message":
-            self.mock_bot_handler.send_message.assert_called_with(response)
-        else:
-            self.mock_bot_handler.send_reply.assert_called_with(message, response['content'])
+        # Mock BotHandler; get instance
+        instance = self.MockClass.return_value
+
+        # Send message to the bot
+        try:
+            self.message_handler.handle_message(message, self.mock_bot_handler)
+        except KeyError as key_err:
+            raise Exception("Message tested likely required key {}.".format(key_err))
+
+        # Determine which messaging functions are expected
+        send_messages = [call(r[0]) for r in responses if r[1] == 'send_message']
+        send_replies = [call(message, r[0]['content']) for r in responses if r[1] == 'send_reply']
+
+        # Test that call were matching in quantity, and then in details
+        fail_template = "\nMESSAGE:\n{}\nACTUAL CALLS:\n{}\nEXPECTED:\n{}\n"
+        functions_to_test = (('send_message', instance.send_message, send_messages),
+                             ('send_reply', instance.send_reply, send_replies))
+        for version, actual, expected in functions_to_test:
+            assert len(expected) == actual.call_count, (
+                "Numbers of '{}' called do not match those expected ({} calls, {} expected)" +
+                fail_template).format(version, actual.call_count, len(expected),
+                                      message, actual.call_args_list, expected)
+            if len(expected) > 0:
+                try:
+                    actual.assert_has_calls(expected)
+                except AssertionError:
+                    raise AssertionError(
+                        ("Calls to '{}' do not match those expected" +
+                         fail_template).format(version,
+                                               message, actual.call_args_list, expected))
+                actual.reset_mock()  # Ensure the call details are reset
 
     @contextmanager
     def mock_config_info(self, config_info):
@@ -154,7 +176,11 @@ class BotTestCaseBase(TestCase):
         # type: (Dict[str, Any], Dict[str, Any], str) -> None
         # Strictly speaking, this function is not needed anymore,
         # kept for now for legacy reasons.
-        self.call_request(message, expected_method, response)
+        self.call_request(message, (response, expected_method))
+
+    def assert_bot_responses(self, message, *response_list):
+        # type: (Dict[str, Any], *Tuple[Dict[str, Any], str]) -> None
+        self.call_request(message, *response_list)
 
 class BotTestCase(BotTestCaseBase):
     """Test class extending BotTestCaseBase to add common default tests


neiljp:
PR updated to be compatible with the merged changes to allow state testing in #25, including an additional minor adjustment to the tictactoe bot to use a new keyword form for state_handler. That avoids the need for one argument to the testing function (eg. `assert_bot_response*`) to be an explicit list of responses, and instead uses a `*response_list` form, while I believe retaining python2 compatibility.

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -70,29 +71,58 @@ def check_expected_responses(self, expectations, expected_method='send_reply',
...
+            self.message_handler.handle_message(message, bot_handler, state_handler)
+        except KeyError as key_err:
+            raise Exception("Message tested likely required key {}.".format(key_err))
+
+        # Determine which messaging functions are expected
+        send_messages = [call(r[0]) for r in responses if r[1] == 'send_message']
+        send_replies = [call(message, r[0]['content']) for r in responses if r[1] == 'send_reply']
+
+        # Test that call were matching in quantity, and then in details
+        fail_template = "\nMESSAGE:\n{}\nACTUAL CALLS:\n{}\nEXPECTED:\n{}\n"
```
derAnfaenger: Maybe capitalize FAIL_TEMPLATE ?

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -70,29 +71,58 @@ def check_expected_responses(self, expectations, expected_method='send_reply',
...
+        functions_to_test = (('send_message', instance.send_message, send_messages),
+                             ('send_reply', instance.send_reply, send_replies))
+        for ftt in functions_to_test:
```
derAnfaenger: Maybe we want to make this a bit easier to understand by unpacking `ftt` here.

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -70,29 +71,58 @@ def check_expected_responses(self, expectations, expected_method='send_reply',
...
+            raise Exception("Message tested likely required key {}.".format(key_err))
+
+        # Determine which messaging functions are expected
+        send_messages = [call(r[0]) for r in responses if r[1] == 'send_message']
+        send_replies = [call(message, r[0]['content']) for r in responses if r[1] == 'send_reply']
+
+        # Test that call were matching in quantity, and then in details
+        fail_template = "\nMESSAGE:\n{}\nACTUAL CALLS:\n{}\nEXPECTED:\n{}\n"
+        functions_to_test = (('send_message', instance.send_message, send_messages),
+                             ('send_reply', instance.send_reply, send_replies))
```
derAnfaenger: Is there any benefit of using a tuple of tuples vs a list of tuples?

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -70,29 +71,58 @@ def check_expected_responses(self, expectations, expected_method='send_reply',
...
+            else:
+                try:
+                    if len(ftt[2]) > 0:
+                        ftt[1].assert_has_calls(ftt[2])  # any_order = True FIXME
+                except AssertionError:
+                    err = ("Calls to {} do not match those expected" +
+                           fail_template).format(ftt[0],
+                                                 message, ftt[1].call_args_list, ftt[2])
+            if err is not None:
+                raise AssertionError(err)
```
derAnfaenger: I'm not sure about the control flow here - maybe we want to print the error messages directly in the respective assertion and then just call `raise`? This way, one gets the line number where the exception is happening.

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -70,29 +71,58 @@ def check_expected_responses(self, expectations, expected_method='send_reply',
...
-        if expected_method == "send_message":
-            instance.send_message.assert_called_with(response)
-        else:
-            instance.send_reply.assert_called_with(message, response['content'])
+
+        # Send message to the bot
+        try:
+            self.message_handler.handle_message(message, bot_handler, state_handler)
+        except KeyError as key_err:
+            raise Exception("Message tested likely required key {}.".format(key_err))
```
neiljp: I added this test, since some of my bots (maybe pollbot, possibly others) depended on parameters of message, which were not supplied in the version provided in the test. So, the KeyError is a hint to a test-writer that an additional key should be added to the test message. I'm fully open to having better descriptive text here, but I believe that's the case it is attempting to catch.

Of course, KeyError could arise due to other reasons, so this could be at least 'perhaps' rather than 'likely'?

zulip_bots/zulip_bots/test_lib.py
neiljp: For what reason? I've not seen a guide towards this format, as compared to simply lower-case?

zulip_bots/zulip_bots/test_lib.py
neiljp: I only made this change recently (it was either a list of tuples or list of lists), simply on the premise that it indicates an intent that the structure is immutable.

zulip_bots/zulip_bots/test_lib.py
neiljp: Yes, that's a great point.

neiljp: Amended in additional commit (to be squashed into first?)

zulip_bots/zulip_bots/test_lib.py
neiljp: Do you mean, to simply let the `AssertionError`s be uncaught?
I had considered re-raising after outputting the error, but a quick test of that fail for some reason. I've found it very useful to have the full information in each `err`, so I hope you mean to keep that? I'm not sure what is lost by the approach I've used (at least in this case).

zulip_bots/zulip_bots/test_lib.py
derAnfaenger: I just thought it to be more suitable for a string template; also, I know quite a few bots signal their constant strings. On the other hand, others don't, so there doesn't seem to be too much consistency. I'm fine with leaving this the way it is :)

zulip_bots/zulip_bots/test_lib.py
derAnfaenger: Ah, now I understand, thanks. I'm just a little wary of raising a new exception in an except block, since you then do not get the full traceback to there `KeyError` happened, but only to where your own exception was raised. How about
```.py
try:
             instance.send_reply.assert_called_with(message, response['content'])		 +            self.message_handler.handle_message(message, bot_handler, state_handler)
         except KeyError as key_err:
             print("Message tested likely required key {}.".format(key_err))
             raise
``` ?

File: zulip_bots/zulip_bots/test_lib.py [hunk 6]
```diff
@@ -70,29 +71,58 @@ def check_expected_responses(self, expectations, expected_method='send_reply',
...
+            else:
+                try:
+                    if len(expected) > 0:
+                        actual.assert_has_calls(expected)  # any_order = True FIXME
+                except AssertionError:
+                    err = ("Calls to {} do not match those expected" +
+                           fail_template).format(version,
+                                                 message, actual.call_args_list, expected)
+            if err is not None:
+                raise AssertionError(err)
```
derAnfaenger: I meant this in a way to do smt like:
```.py
try:
                assert(len(expected) == actual.call_count)
            except AssertionError:
                print("Numbers of {} called do not match those expected ({} calls, {} expected)" +
                       fail_template).format(version, actual.call_count, len(expected),
                                             message, actual.call_args_list, expected)
                raise
            else:
                try:
                    if len(expected) > 0:
                        actual.assert_has_calls(expected)  # any_order = True FIXME
                except AssertionError:
                    print ("Calls to {} do not match those expected" +
                           fail_template).format(version,
                                                 message, actual.call_args_list, expected)
                    raise
```

neiljp: I tried something like this, with simply `print(err); raise` after each `err=...` assignment, but this shifts the output to the start of the test output. For example, the (current) error message in master in the weather-bot without it:
```
AssertionError: Calls to send_reply do not match those expected
MESSAGE:
{'content': 'United Kingdom'}
ACTUAL CALLS:
[call({'content': 'United Kingdom'}, 'Weather in London, GB:\n58.73 F / 14.85 C\nShower Rain')]
EXPECTED:
[call({'content': 'United Kingdom'}, 'Weather in London, GB:\n58.33 F / 14.85 C\nShower Rain')]
```
is then printed right near the start (away from the traceback where you'd see it), with the 'final' text then only:
```
AssertionError: Calls not found.
Expected: [call({'content': 'United Kingdom'}, 'Weather in London, GB:\n58.33 F / 14.85 C\nShower Rain')]
Actual: [call({'content': 'United Kingdom'}, 'Weather in London, GB:\n58.73 F / 14.85 C\nShower Rain')]
```
(I actually thought the text was just not being printed for some reason)

With python3 the blocks can be of the form of the following, and works more like you want:
```
try:
    [...]
except AssertionError as e:
    raise AssertionError("...") from e
```
...but apparently this is new in python3, so isn't an option here?

The rest of the traceback is the same if I re-raise; is it the case of accidentally catching other assert() failures in code that you're wanting to avoid? If so, would it be simpler to just switch to using `if`, at least for the first test?

derAnfaenger: I see, this makes sense. Ye, that python 3 option would perfectly fit my thoughts, but you are right - I think your way is the one to go then.

zulip_bots/zulip_bots/test_lib.py
neiljp: Did you mean to include the `.assert_called_with` in there?

This is essentially the same issue as later in the code, with a `print` outputting not in the same context.

I've been exploring this a bit, and I believe that, at least in python3 (even without a `from key_err`), raising within an except still gives a nice 'context', such as when I remove a line from testing tictactoe:
```
ERROR: test_bot (zulip_bots.bots.tictactoe.test_tictactoe.TestTictactoeBot)
  File "/home/neil/repos/python-zulip-api/zulip_bots/zulip_bots/test_lib.py", line 95, in call_request
    self.message_handler.handle_message(message, bot_handler, state_handler)
  File "/home/neil/repos/python-zulip-api/zulip_bots/zulip_bots/bots/tictactoe/tictactoe.py", line 282, in handle_message
    original_sender = message['sender_email']
KeyError: 'sender_email'

During handling of the above exception, another exception occurred:

  File "/home/neil/repos/python-zulip-api/zulip_bots/zulip_bots/bots/tictactoe/test_tictactoe.py", line 48, in test_bot
    'send_message', state_handler=state)
  File "/home/neil/repos/python-zulip-api/zulip_bots/zulip_bots/test_lib.py", line 174, in assert_bot_response
    self.call_request(message, (response, expected_method), **kwargs)
  File "/home/neil/repos/python-zulip-api/zulip_bots/zulip_bots/test_lib.py", line 97, in call_request
    raise Exception("Message tested likely required key {}.".format(key_err))
Exception: Message tested likely required key 'sender_email'.
```
though in python2 this becomes only:
```
ERROR: test_bot (zulip_bots.bots.tictactoe.test_tictactoe.TestTictactoeBot)
  File "zulip_bots/bots/tictactoe/test_tictactoe.py", line 48, in test_bot
    'send_message', state_handler=state)
  File "zulip_bots/test_lib.py", line 174, in assert_bot_response
    self.call_request(message, (response, expected_method), **kwargs)
  File "zulip_bots/test_lib.py", line 97, in call_request
    raise Exception("Message tested likely required key {}.".format(key_err))
Exception: Message tested likely required key 'sender_email'.
```
So, I agree with you that context is lost in this case at least. Can we just suggest using python3? :)

neiljp:
Thanks for the reviewing @derAnfaenger; not sure about the outstanding exception issue, but the code reads much more nicely after the tuple unpacking!

zulip_bots/zulip_bots/test_lib.py
derAnfaenger: I think @timabbott 's position on this is that if we decide to migrate `python-zulip-api` to use python3 one day, it'll be one of the last things to migrate, since this repo is something that is likely to get integrated in 3rd party code, possibly still relying on python2.

derAnfaenger: And unfortunately, I don't think `six` will be of any help here.

zulip_bots/zulip_bots/test_lib.py
neiljp: For this and for the query you had above re testing KeyError?

derAnfaenger: Yes, for both, unfortunately.

neiljp:
For now possible outstanding changes I might consider are:
* `version` to `function` or `msg_function` in a minor final naming adjustment;
* directly calling `raise AssertionError(...)` and avoiding the err variables?

timabbott:
@neiljp can you rebase this?  Also, am I correct that this is an independent set of changes from the state_handler things we just merged?

neiljp:
Now rebased; the changes are independent of the state-handler changes, though it does touch the testing coding so I did have to resolve some conflicts now that you've merged the test_lib.py changes from that.

neiljp:
I added assert_bot_responses, but ideally we'd just do a sweep through the bots to use a tuple? Some other interface is feasible too, of course, but the pairs keep it clean.

neiljp:
@timabbott @derAnfaenger 
I just rebased and fixed a few issues after a previous commit which were causing failing tests.

I'd consider the plural form to be the generic form, hence the keyword form required for state_handler. If considered appropriate, perhaps even switching to the pluralised form as that preferred in future.

Is there anything limiting the merge of this? Or just a low priority since no bot currently-merged does this?

The last commit can certainly be squashed into the first in the final merge.

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -95,24 +96,52 @@ def check_expected_responses(self, expectations, expected_method='send_reply',
...
-            instance.send_reply.assert_called_with(message, response['content'])
+
+        # Send message to the bot
+        try:
+            self.message_handler.handle_message(message, bot_handler, state_handler)
+        except KeyError as key_err:
+            raise Exception("Message tested likely required key {}.".format(key_err))
+
+        # Determine which messaging functions are expected
+        send_messages = [call(r[0]) for r in responses if r[1] == 'send_message']
```
derAnfaenger: I think `unittest.call` would make this statement clearer.

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -95,24 +96,52 @@ def check_expected_responses(self, expectations, expected_method='send_reply',
...
-        else:
-            instance.send_reply.assert_called_with(message, response['content'])
+
+        # Send message to the bot
+        try:
+            self.message_handler.handle_message(message, bot_handler, state_handler)
+        except KeyError as key_err:
+            raise Exception("Message tested likely required key {}.".format(key_err))
+
+        # Determine which messaging functions are expected
```
derAnfaenger: "are called"

zulip_bots/zulip_bots/test_lib.py
derAnfaenger: Having a second look at this - would the whole chunk here be made easier if we just called `assert` with the error string as an argument in both cases? We could omit the except clauses and error variable.

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -97,7 +97,7 @@ def check_expected_responses(self, expectations, expected_method='send_reply',
                 message = dict(trigger_message, content = m)
                 response = {'content': r} if expected_method == 'send_reply' else r
                 self.assert_bot_responses(message, (response, expected_method),
-                                          state_handler=state_handler)
+                                          state_handler=current_state_handler)
```
derAnfaenger: Was this a bug?

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -95,24 +96,52 @@ def check_expected_responses(self, expectations, expected_method='send_reply',
...
+            else:
+                try:
+                    if len(ftt[2]) > 0:
+                        ftt[1].assert_has_calls(ftt[2])  # any_order = True FIXME
+                except AssertionError:
+                    err = ("Calls to {} do not match those expected" +
+                           fail_template).format(ftt[0],
+                                                 message, ftt[1].call_args_list, ftt[2])
+            if err is not None:
+                raise AssertionError(err)
```
derAnfaenger: Having a second look at this - would the whole chunk here be made easier if we just called assert with the error string as an argument in both cases? We could omit the except clauses and error variable.

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -95,24 +96,52 @@ def check_expected_responses(self, expectations, expected_method='send_reply',
...
+            else:
+                try:
+                    if len(expected) > 0:
+                        actual.assert_has_calls(expected)  # any_order = True FIXME
+                except AssertionError:
+                    err = ("Calls to {} do not match those expected" +
+                           fail_template).format(version,
+                                                 message, actual.call_args_list, expected)
+            if err is not None:
+                raise AssertionError(err)
```
derAnfaenger: Having a second look at this - would the whole chunk here be made easier if we just called assert with the error string as an argument in both cases? We could omit the except clauses and error variable.

derAnfaenger: Reposted this comment because the previous one was made for an outdated commit and hence collapsed.

zulip_bots/zulip_bots/test_lib.py
neiljp: Sorry, I don't understand what you mean.

derAnfaenger: Explicitly writing `unittest.call` instead of `call` in this case. But I'm surprised this comment is still here? I thought I deleted it - it's not really important, after all.

zulip_bots/zulip_bots/test_lib.py
neiljp: Good catch; I've added another commit to show this separately, but this can be again squashed into the other if it looks good.

neiljp:
@derAnfaenger Just checking on this PR now; have rebased and see some test errors but haven't looked into them as yet.

timabbott:
@neiljp can you squash the fixup commit (it's definitely better) and fix the test failures?  I'll add a couple other small comments.

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -90,21 +91,43 @@ def check_expected_responses(self, expectations, expected_method='send_reply',
...
+        functions_to_test = (('send_message', instance.send_message, send_messages),
+                             ('send_reply', instance.send_reply, send_replies))
+        for version, actual, expected in functions_to_test:
+            assert len(expected) == actual.call_count, (
+                "Numbers of '{}' called do not match those expected ({} calls, {} expected)" +
+                fail_template).format(version, actual.call_count, len(expected),
+                                      message, actual.call_args_list, expected)
+            try:
+                if len(expected) > 0:
+                    actual.assert_has_calls(expected)  # any_order = True FIXME
```
timabbott: We should probably just pass `any_order=True`, right?

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -90,21 +91,43 @@ def check_expected_responses(self, expectations, expected_method='send_reply',
...
+            except AssertionError:
+                raise AssertionError(
+                    ("Calls to '{}' do not match those expected" +
+                     fail_template).format(version,
+                                           message, actual.call_args_list, expected))
+            actual.reset_mock()  # Ensure the call details are reset
```
timabbott: This indentation layering seems funny to me; maybe move the `if len(expected)` check to outside the try?

neiljp: I understand the reasoning and have made the change; I think I didn't do so before as then all the code inside the `if` becomes yet another step indented, which is the compromise. I'm happy with the `if` outside though.

zulip_bots/zulip_bots/test_lib.py
neiljp: I *think* that should be right, but I wasn't sure when working on it, though I believe so now.

I'm wondering if this would be cleaner and more concise as:
```python
if len(expected) > 0:
    if set(expected) != set(actual.call_args_list):
        raise AssertionError("...")
    actual.reset_mock()
```
That avoids the slightly strange re-raise and is more explicit?

neiljp:
Pending:
- look at why tests are failing
- (possibly) adjust non-matching code to use sets, as in above comment.

neiljp:
This should now be ready for review.

neiljp:
BTW, I assume for backwards-compatibility we do want the singular and plural forms of `assert_bot_response`?

showell:
Merged, thanks @neiljp!

PR #29, opened 2017-07-25
Clean up googlesearch bot.

abhijeetkaur:
I think that this one is a very high utility bot and needs to be worked upon.
I have made the changes in separate commit as I am not sure if we want to go ahead with the second one right now.
The first commit is just reverting the googlesearch.py file back to its original state (mistakenly changed by someone)
The second commit makes it python 2.x compatible.

### DIFF
--- a/zulip_bots/zulip_bots/bots/googlesearch/googlesearch.py
+++ b/zulip_bots/zulip_bots/bots/googlesearch/googlesearch.py
@@ -1,7 +1,7 @@
 # See readme.md for instructions on running this code.
 from __future__ import print_function
 import logging
-import http.client
+from six.moves.urllib import error
 from six.moves.urllib.request import urlopen
 
 # Uses the Google search engine bindings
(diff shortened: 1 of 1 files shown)


showell:
@timabbott For the second commit here, do we want to use `httplib2` to solve the 2/3 issue?  Are we absolutely sure we want to support python2 for bots?

timabbott:
I merged the first commit.  For the second, we should use six's replacement for  httplib2; see https://pythonhosted.org/six/.

And yes, we definitely want to support Python 2 for bots.  Unlike the Zulip server, our bot software is a library, and may end up being integrated with other third-party code that doesn't support Python 3.  Library get to drop Python 2 support last.

showell:
From the six docs:  "For the most part, six.moves aliases are the names of the modules in Python 3."  So I think you can say something like `from fix.moves.http.client import RemoteDisconnected`.

timabbott:
The name for `http.client` is a bit different (I think `six.moves.http_client`) but regardless the mapping is in a table in the link I posted.

timabbott:
Nice, merged, thanks @abhijeetkaur!

Issue #9, opened 2017-07-06
bots: Audit bots for filesystem access.


showell:
see https://github.com/zulip/zulip/issues/5395 for context

We want to double check that:
* bots use StateHandler for saving state
* bots don't directly access the file system for config info

This project may already be done, and the best person to address this issue is somebody fairly involved with the project already, so they know what to check for.

derAnfaenger:
Sorry for this late reply!
@tabbott this is partially done, except for the part where we want to handle general file accesses by bots. I've started to write up a PR for this, but got heavily delayed due to my Asia trip. Assuming this is nothing critical, I'll be happy to say that I'm still working on this.

tabbott:
thanks for the update :q

derAnfaenger:
Oh no, now I'm already tagging the wrong people :/ sorry, meant @timabbott

tabbott:
No worries! It's actually not the first time this has happened. Perhaps not the last time either :)

timabbott:
OK that plan makes sense @derAnfaenger.

derAnfaenger:
@timabbott this issue can be closed.

timabbott:
Sounds good!

PR #30, opened 2017-07-25
bots: Add a restricted open function.

derAnfaenger:
This finalizes #9 .

@sampritipanda

### DIFF
--- a/zulip_bots/zulip_bots/bots/john/john.py
+++ b/zulip_bots/zulip_bots/bots/john/john.py
@@ -101,11 +101,11 @@ class JohnHandler(object):
         self.bot.train(
             "chatterbot.corpus.english"
         )
-        self.bota = create_chat_bot(True)
+        self.chatterbot = create_chat_bot(True)
 
     def handle_message(self, message, bot_handler, state_handler):
         original_content = message['content']
-        bot_response = str(self.bota.get_response(original_content))
+        bot_response = str(self.chatterbot.get_response(original_content))
         bot_handler.send_reply(message, bot_response)
 
 handler_class = JohnHandler
--- a/zulip_bots/zulip_bots/bots/john/john.py
+++ b/zulip_bots/zulip_bots/bots/john/john.py
@@ -11,10 +11,10 @@ except ImportError:
                       Please: pip install chatterbot""")
 
 BOTS_DIR = os.path.dirname(os.path.abspath(__file__))
-JOKES_PATH = os.path.join(BOTS_DIR, 'assets/var/jokes.json')
 DATABASE_PATH = os.path.join(BOTS_DIR, 'assets/var/database.db')
 DIRECTORY_PATH = os.path.join(BOTS_DIR, 'assets')
 VAR_PATH = os.path.join(BOTS_DIR, 'assets/var')
+JOKES_PATH = 'assets/var/jokes.json'
 
 if not os.path.exists(DIRECTORY_PATH):
     os.makedirs(DIRECTORY_PATH)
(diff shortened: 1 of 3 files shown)


zulip_bots/zulip_bots/lib.py
```diff
@@ -106,6 +109,17 @@ def get_config_info(self, bot_name, section=None, optional=False):
             raise
         return dict(config.items(section))
 
+    @contextmanager
+    def open(self, filepath):
+        # type: (str) -> None
+        filepath = os.path.normpath(filepath)
+        abs_filepath = os.path.join(self._bot_dir, filepath)
+        if abs_filepath.startswith(self._bot_dir):
+            yield open(abs_filepath)
```
showell: You don't really need to make this a contextmanager, since you're not really wrapping any behavior within this function.  Instead, you can simply return the other context manager, so change `yield` to `return`.

zulip_bots/zulip_bots/lib.py
```diff
@@ -47,12 +49,13 @@ def show_error_and_exit(self):
...
 class ExternalBotHandler(object):
-    def __init__(self, client):
-        # type: (Client) -> None
+    def __init__(self, client, bot_name):
+        # type: (Client, string) -> None
         # Only expose a subset of our Client's functionality
         user_profile = client.get_profile()
         self._rate_limit = RateLimit(20, 5)
         self._client = client
+        self._bot_dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'bots', bot_name)
```
showell: I'd try to avoid have `os.path.join` in ExternalBotHandler, since that limits how easy it to deploy it.  Can we change the argument to be something like `root_dir` and have `server.py` figure out the path?

derAnfaenger: Yup, that's actually what I originally had. My thinking was actually that it would limit how `server.py` is to deploy :P. What do you think?

zulip_bots/zulip_bots/lib.py
derAnfaenger: Good point!

zulip_bots/zulip_bots/lib.py
showell: I think we want the library to be as flexible as possible.  Nothing prevents us from writing more flexible code in `server.py` down the road, or, if the server is liberal about what it runs, we can have an option where the `server` doesn't pass in the `root_dir`.  In the library, I think if `root_dir` is `None`, we assume it doesn't matter where the files come from.

derAnfaenger: Alright, makes sense.

zulip_bots/zulip_bots/bots/john/john.py
```diff
@@ -111,9 +57,55 @@ def usage(self):
...
+            for joke in json.load(data_file):
+                self.bot.train([
+                    "Please can you tell me a joke?",
+                    joke['joke'],
+                ])
+        self.bot.set_trainer(ChatterBotCorpusTrainer)
+        self.bot.train(
+            "chatterbot.corpus.english"
+        )
+        self.bota = create_chat_bot(True)
```
timabbott: Maybe add another commit giving this a better name?  E.g. `self.chatterbot` ?

timabbott:
I posted a couple other minor comments; this looks good to me modulo the comments above.

derAnfaenger:
@timabbott updated.

timabbott:
Merged, thanks @derAnfaenger!  I think we may want to add a lint rule to prevent using the normal `open` in our bots, so that people don't add references accidentally.

derAnfaenger:
@timabbott we added that linter rule a while ago, but that reminds me that we still need to remove the exception for the john bot.

Issue #28, opened 2017-07-25
Set User-Agent correctly for Zulip bots


timabbott:
We should set User-Agent so it is easy for the Zulip server's logging to track usage of the various bots.

The pattern we've established for this sort of thing is e.g. `ZulipTrelloWebhook` for our webhooks.  So I think we probably want e.g. `ZulipConverterBot` as the naming scheme for these.  Ideally, the framework would generate the client name, so there's no need to make any individual per-bot changes.  

We would pass this as an extra option to `zulip.Client` or `zulip.init_from_options` (depending what we're using).

neiljp:
Working on this just now.

neiljp:
How does #33 look as a resolution to this?

PR #33, opened 2017-07-26
bots: Set user-agent for individual bots automatically.

neiljp:
I'm unsure how to test if this provides a resolution for #28, since I don't run a server right now, but it seems like it should.

### DIFF
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -117,7 +117,7 @@ class StateHandler(object):
         # type: () -> Any
         return self.state
 
-def run_message_handler_for_bot(lib_module, quiet, config_file):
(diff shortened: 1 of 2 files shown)


timabbott:
The code looks correct.  I would probably test manually by just hacking the Zulip API bindings code to print out the User-Agent before sending just to make sure it looks right, but eventually this seems like the kind of thing we'd check in a unit test suite for the framework, which would probably be worth starting to build soon :)

neiljp:
Various locations indicate this looks ok for wikipedia, giving a final user-agent string of (here): `ZulipWikipediaBot (debian; 9.1)`

timabbott:
Great, merging!

timabbott:
Merged, thanks @neiljp!

PR #35, opened 2017-07-27
python-zulip-api: Migrate from optparse to argparse.

eeshangarg:


timabbott:
So, if there's any third-party code using the Zulip Python API that was using `zulip.generate_option_group` with optparse, it's going to break with these changes, right?  I don't think we want to break third-party projects without reason.

So I think we want to actually just keep the old library code as `zulip.generate_option_group` (maybe with an added `logging.warning()` encouraging the user to migrate to argparse), and make the new feature have some other (perhaps clearer) name like `zulip.add_default_options`.

eeshangarg:
@timabbott I was wrong, there is no way to migrate to `argparse` without breaking third-party code because all function names and everything has been changed. And the way argument groups are implemented is significantly different from `optparse`. I misunderstood what you had meant by "breaking our API", I thought you meant just breaking the code in the repo (it didn't cross my mind that people might be using `generate_group_options` too, I just assumed they only use `zulip.init_from_options` (now that I think about it, its a stupid assumption to make). So I am sorry, I'll get to working on what you suggested! Thanks so much for the review! :)

eeshangarg:
@timabbott Updated. I decided to call the new function `add_default_arguments` because the use of the word `arguments` instead of `options` is more consistent with the function names already present in `argparse`. I ran all of the examples with my changes and they all worked! :)

Next steps:
* Migrate `integrations/` to `argparse`
* Also, make sure that all `integrations` use `zulip.init_from_options` (a few of them don't, as reported by someone on `chat.zulip.org`; I can't recall who exactly reported this bug)
* Move on to working on documentation (I think I'll start on that alongside the two points above)

Let me know what you think!

Regards,
Eeshan

timabbott:
That plan sounds good.  I noticed two things:
* What's the story with the prefix argument?
* Also, why did you remove all the sys.path code in the API examples?  That code, if I'm not mistaken, meant that you could run the examples directly from the zulip checkout without having to actually install the zulip package, if you wanted to.

(As a sidenote, even if that `sys.path` editing code was useless, I would still not remove it in the same commit as an unrelated refactoring, since it's actually easier to review the 2 pieces separately).

eeshangarg:
@timabbott: Updated with the following changes.

* The new `add_default_arguments` function now doesn't accept a `prefix` kwarg.
* The changes removing `sys.path` modifications and associated redundant imports have been moved a commit of their own.

Once we get this and the `requests.Sessions` PR merged, we should be ready for an official release of the three packages! :)

Thanks,
\- Eeshan

timabbott:
Looks good, merged, thanks @eeshangarg!

PR #40, opened 2017-07-28
[WIP] Stateful bots: improve test coverage.

neiljp:
**Current state**:
* Awaiting confirmation that changes to bot testing lib are good, and whether to switch to list[tuple]
* TicTacToe tests should be good
* VirtualFS tests are a preliminary change pending ordered changes, unless just using a loop like in tictactoe; further other work pending discussion on whether 'sample conversation' should be re-used between bot and tests, or just copied.
* Incrementor test is highly minimal pending adding update_message testing being possible, likely in another branch, perhaps after multi-response #40?

### DIFF
--- a/zulip_bots/zulip_bots/bots/virtual_fs/test_virtual_fs.py
+++ b/zulip_bots/zulip_bots/bots/virtual_fs/test_virtual_fs.py
@@ -4,45 +4,46 @@ from __future__ import absolute_import
 from __future__ import print_function
 
 from zulip_bots.test_lib import BotTestCase
+from zulip_bots.lib import StateHandler
 
 class TestVirtualFsBot(BotTestCase):
     bot_name = "virtual_fs"
 
     def test_bot(self):
+        help_txt = ('foo_sender@zulip.com:\n\nThis bot implements a virtual file system for a stream.\n'
+                    'The locations of text are persisted for the lifetime of the bot\n'
+                    'running, and if you rename a stream, you will lose the info.\n'
+                    'Example commands:\n\n```\n'
+                    '@mention-bot sample_conversation: sample conversation with the bot\n'
+                    '@mention-bot mkdir: create a directory\n'
+                    '@mention-bot ls: list a directory\n'
+                    '@mention-bot cd: change directory\n'
+                    '@mention-bot pwd: show current path\n'
+                    '@mention-bot write: write text\n'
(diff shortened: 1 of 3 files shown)


timabbott:
This approach looks pretty good!  I hope we manage to have a gain finish with victory :)

neiljp:
Thanks @timabbott. TicTacToe uses assert_bot_response, but since VirtualFS uses check_expected_responses I've added two extra commits to ensure that state isn't reused between sources (stream/private) and that the "expectations" are ordered, whereas the regular dict input isn't.

If we're open to tests being refactored, the List[Tuple[str,Any]] is more straightforward?

Also, particularly given the complexity of virtualfs and the 'sample conversation', I'm wondering if this can be split into a separate file used by both the bot and for testing?

timabbott:
@neiljp can you rebase this and fix all the lint errors you'll get once the linter is running?

zulip_bots/zulip_bots/bots/virtual_fs/test_virtual_fs.py
```diff
@@ -27,22 +24,25 @@ def test_bot(self):
...
         }
         self.check_expected_responses(expected)
+
+        expected = [
+            ("help", help_txt),
+            ("help ls", "foo_sender@zulip.com:\nsyntax: ls <optional_path>"),
+            ("", help_txt),
+            ("pwd", "foo_sender@zulip.com:\n/"),
+            ("cd /home", "foo_sender@zulip.com:\nERROR: invalid path"),
+            ("mkdir home", "foo_sender@zulip.com:\ndirectory created"),
```
timabbott: can you add a final `cd /home` here?

timabbott:
I merged the framework changes, after making a few tweaks to improve the variable names and make it all more readable.  

@neiljp can you rebase the actual tests and fix all the lint errors?  Apparently the linter wasn't running at whatever commit this is based on, and there's a ton of whitespace issues.

timabbott:
Merged, thanks @neiljp!  I agree these are a lot harder to read than with the nicely whitespace-aligned code.  I'm just not sure what an easy way to quiet the PEP8 linter for just those blocks is.  May be worth googling a bit for suggestions, since it's a popular tool.

neiljp:
It actually doesn't seem as bad now the comments are moved :)

PR #42, opened 2017-07-31
python-zulip-api: Upgrade package versions (0.3.1 -> 0.3.2).

eeshangarg:
I just released all three of our pip packages:
* [zulip API bindings package](https://pypi.python.org/pypi/zulip)
* [zulip bots/bots API pakcage](https://pypi.python.org/pypi/zulip-bots)
* [zulip Flask bot server package](https://pypi.python.org/pypi/zulip-botserver)

I tested all three on my system and everything seems fine for now. The `zulip_bots` package doesn't ship fixtures or tests (like it shouldn't). I would really appreciate it if someone could test all three out in a virtualenv and give me some feedback! Thanks! :simple_smile:

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -41,7 +41,7 @@ import logging
 import six
 from typing import Any, Callable, Dict, Iterable, IO, List, Mapping, Optional, Text, Tuple, Union
 
-__version__ = "0.3.1"
+__version__ = "0.3.2"
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


timabbott:
We should probably think about our release management story.  I think we'll definitely want to at least tag the `0.3.2` release in the repository with the exact commit the release was generated from (@eeshangarg can you do this).  And another good practice we have in http://zulip.readthedocs.io/en/latest/release-checklist.html is to immediately add `+git` to the version numbers after a release, so versions installed off of Git are properly named.

PR #46, opened 2017-08-01
log2zulip: Fix to work on Python 3.

gnprice:
I'm not thrilled with the `replace` error handler losing information
if the logfile contains invalid UTF-8 for some reason; but that sure
beats a UnicodeDecodeError, and for this script I can't quite be
bothered to run the rather tricky riddle trail that Python 3 makes it
to pass arbitrary byte data through layers of ordinary text processing.

### DIFF
--- a/zulip/integrations/log2zulip/log2zulip
+++ b/zulip/integrations/log2zulip/log2zulip
@@ -88,7 +88,7 @@ def process_logs():
             # a log file ends up at the same line length as before
             # immediately after rotation, this tool won't notice.
             file_data["last"] = 1
-        new_lines = subprocess.check_output(["tail", "-n+%s" % (file_data["last"],), log_file]).split('\n')[:-1]
+        new_lines = subprocess.check_output(["tail", "-n+%s" % (file_data["last"],), log_file]).decode('utf-8', errors='replace').split('\n')[:-1]
         if len(new_lines) > 0:
             process_lines(new_lines, filename)
             file_data["last"] += len(new_lines)


timabbott:
Have you tested whether this break it for Python 2?  We might need to use a `force_bytes` type method here.

timabbott:
@derAnfaenger can you help test this?  I think it might be easy enough to just test whether it still runs on both Python 2 and 3 with this change.  The tool just works with a log file, so you should be able to just point it at any old file and then use `cat` to add lines to it to test the flow.

derAnfaenger:
image

Worked for me without errors concerning the Python2/3 compatibility, all unicode and invalid unicode characters passed just fine for both versions.

On the other hand, it was somewhat difficult to get this running on Windows, since all the paths are Unix-Hardcoded. While WIndows-support is certainly not a priority, I think it'd turn out helpful if we gave this script the standard argparse abilities other scripts have, so that we could e.g. specify the zuliprc location, etc. 

Also, I had to install the `usjon` package, it was *not* installed as a dependency when installing the `zulip` package. Is this the case because it was once part of the big repo, where `ujson` was a dependency? If so, there might be a couple other dependencies we shold add for the zulip package. What do you think?

timabbott:
Yeah, doing that sort of cleanup would be a good idea.  That history is I think why it used `ujson`, but I think we should just have this use the built-in `json` module; there's no benefit to `ujson` for this application anyway.

I'll merge this, and while you have the test procedure down, I think it'd be worth refactoring this to use our normal argument parsing stuff and work on Windows.

timabbott:
Merged, thanks @gnprice!

PR #47, opened 2017-08-01
Bots: support run-time bot identity (improve eg. help commands) 

neiljp:
This improves the ability of a bot to specify how to mention it, which varies at run-time depending upon the identity used to run it; this is commonly used in many bot help commands.

The wikipedia and xkcd bots are ported as examples.

A similar approach to `mock_config_info` is used to mock the bot identity in a tests, which are also ported for these two bots.

I've wanted to have bots be able to give more contextual examples and this goes another step towards that. I'm also particularly open to thoughts on:
* how bot/user names are quoted (`user` vs `**user with space**`?)
* would a `quoted_name` method be more useful? (with or without the stars)
* if name/email is reasonable (it was available internally already); only name is currently used/tested

### DIFF
--- a/zulip_bots/zulip_bots/bots/wikipedia/test_wikipedia.py
+++ b/zulip_bots/zulip_bots/bots/wikipedia/test_wikipedia.py
@@ -56,7 +56,7 @@ class TestWikipediaBot(BotTestCase):
             )
 
         # Empty query, no request made to the Internet.
-        bot_response = "Please enter your message after @mention-bot"
+        bot_response = "Please enter your message after @**test bot**"
         self.assert_bot_response(
             message = {'content': ''},
             response = {'content': bot_response},
--- a/zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
+++ b/zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
@@ -37,10 +37,11 @@ class WikipediaHandler(object):
         bot_handler.send_reply(message, bot_response)
 
     def get_bot_wiki_response(self, message, bot_handler):
-        help_text = 'Please enter your message after @mention-bot'
+        help_text = 'Please enter your message after @{}'
         query = message['content']
         if query == '':
-            return help_text
+            quoted_name = bot_handler.identity().quoted_name
+            return help_text.format(quoted_name)
         query_wiki_link = ('https://en.wikipedia.org/w/api.php?action=query&'
                            'list=search&srsearch=%s&format=json'
                            % (urllib.parse.quote(query),))
--- a/zulip_bots/zulip_bots/bots/xkcd/test_xkcd.py
+++ b/zulip_bots/zulip_bots/bots/xkcd/test_xkcd.py
@@ -6,19 +6,22 @@ from __future__ import print_function
 import mock
 from mock import MagicMock, patch
 from zulip_bots.test_lib import BotTestCase
+from zulip_bots.lib import ExternalBotHandler
 
 class TestXkcdBot(BotTestCase):
     bot_name = "xkcd"
+    bot_identity = ExternalBotHandler.BotIdentity(name="xkcd bot", quoted_name="**xkcd bot**", email=None)
 
     @mock.patch('logging.exception')
     def test_bot(self, mock_logging_exception):
         help_txt = "xkcd bot supports these commands:"
         err_txt  = "xkcd bot only supports these commands, not `{}`:"
         commands = '''
-* `@xkcd help` to show this help message.
-* `@xkcd latest` to fetch the latest comic strip from xkcd.
-* `@xkcd random` to fetch a random comic strip from xkcd.
-* `@xkcd <comic id>` to fetch a comic strip based on `<comic id>` e.g `@xkcd 1234`.'''
+* `{0} help` to show this help message.
+* `{0} latest` to fetch the latest comic strip from xkcd.
+* `{0} random` to fetch a random comic strip from xkcd.
+* `{0} <comic id>` to fetch a comic strip based on `<comic id>` e.g `{0} 1234`.'''.format(
+            "@" + self.bot_identity.quoted_name)
         invalid_id_txt = "Sorry, there is likely no xkcd comic strip with id: #"
 
         # 'latest' query
--- a/zulip_bots/zulip_bots/test_lib.py
+++ b/zulip_bots/zulip_bots/test_lib.py
@@ -25,6 +25,7 @@ from typing import List, Dict, Any, Optional, Callable, Tuple
 from types import ModuleType
 
 from copy import deepcopy
+from zulip_bots.lib import ExternalBotHandler
 
 class BotTestCaseBase(TestCase):
     """Test class for common Bot test helper methods"""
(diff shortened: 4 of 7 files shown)


zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
```diff
@@ -31,10 +31,12 @@ def handle_message(self, message, bot_handler, state_handler):
         bot_handler.send_reply(message, bot_response)
 
     def get_bot_wiki_response(self, message, bot_handler):
-        help_text = 'Please enter your message after @mention-bot'
+        help_text = 'Please enter your message after @{1}{0}{1}'
         query = message['content']
         if query == '':
-            return help_text
+            name = bot_handler.bot_identity().name
+            pad = '' if name.find(' ') < 0 else '**'
```
rht: (The number loc changes required would have been fewer if there is a method that returns the "{1}{0}{1}" padded string in the lib.)

neiljp: Right; this comes in to my first and second bullet points in the description. There are various options. Bot names are known in the bot_handler, so this could be handled at that level too, with the aforementioned `quoted_name` providing star-quotes if necessary.

I'm as much interested in the general concept of this PR and whether something like it might be considered, as much as the smaller details.

timabbott: This is super hard to read.  Can't we just create a `padding_name` variable and then have the code access that?

zulip_bots/zulip_bots/bots/wikipedia/test_wikipedia.py
```diff
@@ -47,9 +47,10 @@ def test_bot(self):
...
 
         # Empty query, no request made to the Internet.
-        bot_response = "Please enter your message after @mention-bot"
-        self.assert_bot_response(
-            message = {'content': ''},
-            response = {'content': bot_response},
-            expected_method='send_reply'
-        )
+        bot_response = "Please enter your message after @foobot"
+        with self.mock_bot_identity("foobot"):
```
timabbott: I think for the tests, we should just have some fixed value, `@test-bot` be the mocked name of the bot in the test framework, and have that be used everywhere in the tests.  It feels really ugly to need every test to need to do this `with` statement.

timabbott:
This is a nice concept, though the current implementation makes the tests a lot worse.  I think it could be not bad with those issues fixed.

neiljp:
I've made a few changes:
* `BotIdentity` now also has an extra `quoted_name` member, which is set to the form that was in the bots;
* The testing approach is now to just set the bot name to "test bot" in test_lib.py, so I've adjusted the tests for the trial bots that use this approach.

I'm open to changing the `bot_identity` function to perhaps just `identity`, if that's sufficiently obvious?

Also, perhaps including the `@` in the equivalent of `quoted_name` and naming it `mentioned_name` or `mentionable_name` instead?

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -103,12 +104,20 @@ def call_request(self, message, expected_method, response, state_handler):
         # type: (Dict[str, Any], str, Dict[str, Any], Optional[StateHandler]) -> None
         if state_handler is None:
             state_handler = StateHandler()
+
+        instance = self.MockClass.return_value
+
+        # Set mock identity of the bot
+        name = "test bot"
+        pad = '**' if ' ' in name else ''
```
derAnfaenger: We might just always want to wrap a bot's name in `**`, since it's more consistent and also gives a name a nice @-mention  (or bold) formatting.

zulip_bots/zulip_bots/bots/xkcd/test_xkcd.py
```diff
@@ -15,10 +15,10 @@ def test_bot(self, mock_logging_exception):
...
         err_txt  = "xkcd bot only supports these commands:"
         commands = '''
-* `@xkcd help` to show this help message.
-* `@xkcd latest` to fetch the latest comic strip from xkcd.
-* `@xkcd random` to fetch a random comic strip from xkcd.
-* `@xkcd <comic id>` to fetch a comic strip based on `<comic id>` e.g `@xkcd 1234`.'''
+* `@**test bot** help` to show this help message.
+* `@**test bot** latest` to fetch the latest comic strip from xkcd.
+* `@**test bot** random` to fetch a random comic strip from xkcd.
+* `@**test bot** <comic id>` to fetch a comic strip based on `<comic id>` e.g `@**test bot** 1234`.'''
```
derAnfaenger: Maybe .format('@**test bot**'), like in the xkcd bot itself? Or even dynamically fetch the bot's name in the test? But nowt a must.

neiljp: This could be factored-out, but I'm not sure for what purpose, unless we expect the test name to change?

Rather than just take a fixed name in the test lib, we could take a similar approach as in the default methods PR, with an optional class variable in each bot-test-class (ie. like `bot_name`) to specify the `BotIdentity`? It could default to something, but also specified there, so there would be consistency and the `.format` approach could reference the BotIdentity variable too, in the bot-specific test?

zulip_bots/zulip_bots/test_lib.py
neiljp: I'm increasingly feeling that the 'padded' form is really just for mentioning, so I'm inclined to adjust this to providing a `mentioned_name` or similar which even includes the `@` prefix.

Re the unpadded/mentioned name, or even email, we could leave it/them out, but if so then keeping an interface like BotIdentity could be useful?

Having an unquoted name could be useful, in terms of "introducing" themselves without enforcing styling, eg.
```
Hello, you can call me *bot_name*; I'm a bot_type,
so try some of the following to interact with me:
    quoted_name help - show my help text
...
```

How does that sound?

neiljp:
Thanks for the review @derAnfaenger :)

Hopefully I've addressed the two comments and the last issue; rebased with a few more commits for an example generalised identity for specific bot tests, plus fixing the zulip-bot-output.py script.

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -108,11 +108,14 @@ def call_request(self, message, expected_method, response, state_handler):
         instance = self.MockClass.return_value
 
         # Set mock identity of the bot
-        name = "test bot"
-        pad = '**' if ' ' in name else ''
-        instance.bot_identity.return_value = (
-            ExternalBotHandler.BotIdentity(name, "{}{}{}".format(pad, name, pad), None))
+        if hasattr(self.__class__, 'bot_identity'):
+            bot_identity = self.bot_identity
```
derAnfaenger: I find this line a bit confsuing; I think we can drop it and instead do
```
if hasattr(self.__class__, 'bot_identity'):
    self.bot_identity = ExternalBotHandler.BotIdentity(name, "{}{}{}".format(pad, name, pad), None)
instance.bot_identity.return_value = self.bot_identity
```
?

derAnfaenger:
Thanks again for working on this @neiljp! @timabbott the first three commits lgtm. Left a minor comment on the third one, but I'd also like someone else to review the logic.

timabbott:
I have a few requests:
* `self.identity` seems clearer than `self.bot_identity`.
* I think the inconsistent padding doesn't make any sense.  We should show the name surrounded by `**` unconditionally when displaying it; that's much less confusing for it to be consistent.
* For the last commit, I agree with Robert that it's using more magic than needed.

zulip_bots/zulip_bots/test_lib.py
neiljp: I didn't follow this immediately, but upon further inspection I understand what you mean :+1:

neiljp:
@timabbott 
* Agreed wrt verbosity of bot_identity; the original PR was somewhat of a RFC on the idea, and this is totally reasonable;
* Do you disagree with the split name/quoted_name/email approach? If only one of the first two is included it would seem cleaner to me to have just the name, and have an api/utility function just do the quoting (or just have it in each bot that wants to do it, as previously);
* As above, I see and agree with the simpler approach now :)

Due to the changes in name and general acceptance of the testing approach via `bot_identity` in the `BotTestCase`, I've squashed the testing commits together in the lastest push.

neiljp:
@timabbott @derAnfaenger 

I believe this should be mergeable, pending discussion of the [quoted_]name interface. My preference is either roughly as it is, or perhaps providing (name, email) and a quote() or mention() utility function somewhere. Having `name` be quoted/bolded seems too much of a 'surprise'?

neiljp:
Rebased & updated in #369.

PR #48, opened 2017-08-01
integrations/zephyr: Upgrade to argparse.

eeshangarg:


timabbott:
These look good; testing them end-to-end is probably annoying, but you can probably at least check that the arguments get processed correctly by just running it.

eeshangarg:
@timabbott: I tested them by running the commands and just observing whether all the arguments got parsed or not. I'll test them all end-to-end when I'm done with all of them. :)

eeshangarg:
Two things we need to consider:

* Thinking of not upgrading `integrations/jabber` to `argparse`. Mainly because the integration isn't documented anywhere and we need to make sure that the implementation is reliable enough to be documented before we make any further changes. If it *isn't* reliable enough, then the upgrade to `argparse` can be a part of a more involved PR revamping the integration's implementation.
* Also, not upgrading `integrations/perforce` to `argparse` just yet. Argument parsing in said integration relies on the `optparse.make_option` API, which has no equivalent in `argparse`. Here, upgrading to `argparse` might end up being more trouble then its worth. (It would involve changing around the structure of a 1000+ line integration). So, this should be a part of a dedicated PR as well, *if* we decide its worth it to upgrade to `argparse` to begin with.

So, for now, I am just going to test the commits I've pushed as best as I can and submit this PR for review. After this gets merged, I can take a further look at making sure that the integrations I touched in this PR use `zulip.init_from_options`. And then I think its high time that we move on to writing some documentation for these packages!

@timabbott: Thoughts?

eeshangarg:
@timabbott: This is ready for review for now! I tested the argument parsing, it works. :)

timabbott:
I merged all but the Zephyr changes.  The Zephyr ones are wrong, because the argparse options code uses `options.zulip_site`, but the old code stored the value and accessed it as `options.site`.

timabbott:
@eeshangarg can you fix the Zephyr commits to use the right variable names so we can close this out?

rht:
ping

eeshangarg:
@timabbott: Updated and ready for review! :)

eeshangarg:
Updated again! :)

zulip/integrations/zephyr/check-mirroring
```diff
@@ -43,11 +37,11 @@ sys.path[:0] = [os.path.join(options.root_path, "api/"),
 mit_user = 'tabbott/extra@ATHENA.MIT.EDU'
 
 sys.path.append(".")
-import zulip
```
timabbott: If you read the comment above, you'd have known that this was here for a reason.  We want to adjust `sys.path`  before trying to import `zulip`.  What I think you could do is parse the args twice; first, above, just parse `root_path`, and then below here, reparse with the full set of arguments using `zulip_default_args`.

timabbott:
Merged the sync_public_streams part as a new commit.  The other two both have the problem described above, which is that we need to access the `root_path` option before we can import `zulip`.

I don't like rereading something this big repeatedly.  Perhaps we should split this into a few sub-projects: First, move them to argparse (without changes), then, rename the variables in the argparse configuration to be e.g. `zulip_site`, not `site`, etc., and change the scattered code.  And finally, we can do a small commit that does the gymnastics around the problem I just described to be able to use the fancy zulip module argparse stuff.

eeshangarg:
@timabbott: My apologies, I did not realize the purpose behind that `sys.path` modification was anything beyond being just a remnant from the days that the `zulip` package was in the `api/` directory of the main repo. So I overlooked the structure of the arguments, since all of them were simply grouped together. I'll make sure I fix that before I ask for another review! Thanks! :)

zulip/integrations/zephyr/check-mirroring
eeshangarg: @timabbott: In all honesty, I am a bit confused. If you look at the code following the comment that you just mentioned, I think that the adjustment made to `sys.path` is redundant seeing that the "zulip" is a package now and not a directory under `<options.root_path>/api/` that needs to be added to `sys.path` before making the import. Even if you do add it to `sys.path` the, fact is that you would still be importing the `zulip` package (and not some local directory) because the `api/zulip/` doesn't even exist anymore. So isn't importing `zulip` independent of the `sys.path` manipulation in this case now that zulip is an installed package and not necessarily a module just lying in the root of one's `zulip` directory?

I think we talked about how changing the whole redundant `sys.path` modifications would be a follow-up to this PR, but that was few weeks ago, I think. Again, my apologies for not catching this and explaining it earlier but I'd love to know your thoughts about this! :)

timabbott: This script is deployed independently from the rest of the package in one of its 2 main places it's used, for historical reasons.  I suppose I could migrate it to not do that so we can eliminate this feature, but that's what's going on.  Let me see if I find time to investigate doing so today.

timabbott:
The problem is that one of the main places this is deployed is installed in a funny way, and that option is needed for it.  But perhaps I should take this opportunity to try to remove that.

timabbott:
@eeshangarg I just did the work of cleaning up the `--root-path` mess in the `zephyr_mirror` system, so you should be able to finish this now.

zulip/integrations/zephyr/zephyr_mirror_backend.py
```diff
@@ -1108,14 +1100,13 @@ def die_gracefully(signal, frame):
         logger.error("\n" + "nagios_path is required with nagios_class\n")
         sys.exit(1)
 
-    zulip_account_email = options.user + "@mit.edu"
-    import zulip
+    zulip_account_email = options.zulip_email + "@mit.edu"
```
timabbott: This is wrong; the old `--user` option was a username, not an email address, which this line makes clear.  

I think we could probably just do a split to construct the username.

eeshangarg: @timabbott: Hey, if `zulip_account_email` is already an email now, should we add `@mit.edu` at the end? Can't we just stop doing that? Or will this integration only be used by people at MIT?

zulipbot-test:
Hello @eeshangarg, a Zulip maintainer reviewed this pull request over 10 days ago, but you haven't updated your pull request since. Please take a look at the requested changes and update your pull request accordingly.

Thank you for your valuable contributions to Zulip!

showell:
I'm closing this.  I think the risk of breaking the zephyr code outweighs the benefits here.

PR #51, opened 2017-08-05
API reqs: Pin the versions.

rht:


timabbott:
Good call, thanks!  Maybe also worth adding one of those Github shields for whether dependencies are out of date?  This seems like the sort of package where the reminder would be helpful.

timabbott:
Hmm, though thinking about this more, was this change actually a good idea?  I feel like as a library, you want to be really permissive, since other packages might depend on you + other libraries...

timabbott:
Reverted for now.

timabbott:
Reverted for now.

PR #52, opened 2017-08-07
[WIP] Add instructions for setting up a dev environment.

eeshangarg:


README.md
```diff
@@ -4,5 +4,41 @@ This repository contains the source code for Zulip's PyPI packages:
...
+2. Make sure you have [pip](https://pip.pypa.io/en/stable/installing/)
+   and [virtualenv](https://virtualenv.pypa.io/en/stable/installation/)
+   installed.
+
+3. `cd` into the repository cloned earlier:
+   `cd python-zulip-api`
+
+4. Run:
+   ```
+   ./tools/provision
```
aero31aero: Running `source tools/provision` should suffice.

eeshangarg: Huh, I didn't know that! Seems like a weird workflow though! Are you saying that that would also activate the virtualenv?

aero31aero:
Another bug that was discovered because of which I was hesitant to call my PR final was the python2/python3 issue. Virtualenv sets the default python as the one we install venv with. If we did a `pip install virtualenv`, we'll get python2 as the only python executable in the virtual env. Similarly for `pip3 install virtualenv`.

One way around is to manually specify the path to the python executable to use `virtualenv <other options> --python /usr/bin/python3` to hardcode which Python binary you want. The issue is to reliably say which path the binary will be in, or if it will be installed in the first place.

Perhaps mentioning these quirks in the docs would be helpful.

README.md
```diff
@@ -4,5 +4,41 @@ This repository contains the source code for Zulip's PyPI packages:
...
+  for Zulip's Flask bot server.
+
+## Development
+
+1. Fork and clone the Git repo:
+   `git clone https://github.com/<your_username>/python-zulip-api.git`
+
+2. Make sure you have [pip](https://pip.pypa.io/en/stable/installing/)
+   and [virtualenv](https://virtualenv.pypa.io/en/stable/installation/)
+   installed.
```
timabbott: Why not have `tools/provision` install these if they aren't already installed?

README.md
```diff
@@ -4,5 +4,41 @@ This repository contains the source code for Zulip's PyPI packages:
...
+
+### Running tests
+
+You can run the tests for the `zulip_bots` package by typing:
+
+`./tools/test-bots`
+
+You can run the tests for the `zulip_botserver` by typing:
+
+`python -m unittest discover -v ./zulip_botserver`
```
timabbott: Maybe putting this inside `tools/test-botserver` ?

tools/provision
```diff
@@ -0,0 +1,16 @@
...
+    echo "Virtualenv created."
+fi
+
+source $BASEDIR/.zulip-api-env/bin/activate
+
+if [ ! -f "$BASEDIR/.zulip-api-env/updated" -o $BASEDIR/requirements.txt -nt $BASEDIR/.zulip-api-env/updated ]; then
+    pip install -r $BASEDIR/requirements.txt
+    touch $BASEDIR/.zulip-api-env/updated
+    echo "Requirements installed."
+fi
```
timabbott: Let's diff with `requirements.txt` rather than using a timestamp; that's a bit safer if you're playing around with things.

README.md
timabbott: I think we probably want to teach users how this works, but maybe we should have `provision` print out `source` command when it finishes so you only need to remember that part?

timabbott:
This is now #56, I think?

PR #54, opened 2017-08-08
bots: Add complete test coverage for yoda bot.

abhijeetkaur:
Add error handling support for empty messages and invalid
input to yoda bot, which was previously making the bot crash.
Add comments to describe tests.

This is as a followup for #53. I wasn't sure where to put the check for invalid input as the bot code was not responding to invalid input of special characters. Initially, I thought of putting a check by regular expressions (re.match)  but on further exploration, I found out that it is indeed sending an empty reply which is not handled properly by the code.

### DIFF
--- a/zulip_bots/zulip_bots/bots/yoda/fixtures/test_invalid_input.json
+++ b/zulip_bots/zulip_bots/bots/yoda/fixtures/test_invalid_input.json
@@ -1,0 +1,14 @@
+{
+  "request": {
+    "api_url": "https://yoda.p.mashape.com/yoda?sentence=@#$%^&*",
+    "headers": {
+      "X-Mashape-Key": "12345678",
+      "Accept": "text/plain"
+    }
+  },
+  "response": {"text": ""},
+  "response-headers": {
+    "status": 200,
+    "content-type": "text/html; charset=utf-8"
+  }
+}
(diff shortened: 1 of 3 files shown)


timabbott:
Merged, thanks @abhijeetkaur!  @derAnfaenger FYI; I wonder if we want to move the handling of empty messages into some piece of common framework code, so that individual bots don't need to write something unless they want to do something special.

derAnfaenger:
You mean some optional function `handle_empty_message`/`handle_mention` ? I can see this being useful, since crashing for empty messages has been a bug for several bots now. I feel that a new method for this rare case would be a little unproportionate. What do you think of adding a generic test case for this scenario to the `test_lib.py` `BotTestCase` class, given that this is possible?

timabbott:
I think a good approach might be to have default behavior for empty messages provided by the library of sending the user an error back that is controlled by whether or not the `ACCEPT_EMTPY_MESSAGES = 1` is set on the class object.  If it is, then the default behavior doesn't run, but if it is, then the bot will automatically block ignore empty messages.  And then we just need a couple tests at the framework level for whether that feature works correctly, and the 95% of bots that don't want to process empty messages don't need to think about this.

derAnfaenger:
Will create an issue for it.

PR #56, opened 2017-08-10
[WIP] Add tools/provision, virtual-env and corresponding docs

aero31aero:


README.md
```diff
@@ -8,3 +8,37 @@ This repository contains the source code for Zulip's PyPI packages:
...
+   and [virtualenv](https://virtualenv.pypa.io/en/stable/installation/)
+   installed.
+
+3. `cd` into the repository cloned earlier:
+   `cd python-zulip-api`
+
+4. Run:
+   ```
+   ./tools/provision
+   source .zulip-api-env/bin/activate
```
timabbott: These docs need updating, right?

aero31aero: Yes. I'll push them soon. I'm still waiting for confirmation on whether the cygwin fix works as expected or not.

aero31aero: BTW, I think this approach is more in line with how most python devs might be used to. As in, first you set up the environment (`tools/provision`) and then you activate it and start working. So, should we change it to `source tools/provision` or not?

tools/provision
```diff
@@ -0,0 +1,32 @@
...
+
+BASEDIR=`dirname $0`/..
+
+if [ ! -d "$BASEDIR/zulip-api-py2-venv" ]; then
+    virtualenv $BASEDIR/zulip-api-py2-venv
+    echo "Virtualenv created."
+fi
+
+VENVBINDIR="bin"
+if [[ "$OSTYPE" == "cygwin" ]]; then
```
derAnfaenger: `echo $OSTYPE` in CygWin gives me `msys`, so that's what should stand here.

derAnfaenger: @timabbott since this works for all Linux users anyway (and for me with `msys`, just tested it), I think this is ready to merge. Checking the ostype variable here is not optimal, since it doesn't even exist it cmd :). I'll refactor this to rather check for the existence of the directory.

timabbott:
Merged, with a few tweaks:
* Fixed the path to the venv in the README.
* Added proper shell quoting to most variables.
* Made `provision` print out the command for activating the virtualenv.
* Fixed some tab-based whitespace; @derAnfaenger does our linter not check for that, or is this PR just too old?

A follow-up task we should open issues for and/or just fix:
* We should properly support Python 3 here; right now this kinda hardcodes Python 2 in the filenames, even if you're using Python 3.

derAnfaenger:
@timabbott it's seven days, so probably too old. I just added some tabs to a file, and the linter went off.

derAnfaenger:
#78 for the followup.

Issue #55, opened 2017-08-09
Handle empty bot messages through the bots framework.


roberthoenig:
To this point, several bots have shown difficulties to cope with empty messages, leading to crashes.

As discussed in #54 , this issue could be addressed by letting a bot set a state in its class whether they want to receive empty messages.

Steps involved:
* Modify `lib.py` to check a bot's `ACCEPT_EMTPY_MESSAGES` state and conditionally send messages.
* Inform the user in case of missing empty message support.
* (add tests for `lib.py` to verify the feature's functionality? Probably worth another PR)

neiljp:
We definitely noticed this at pycon! I would be interested in pursuing this, since it fits quite well with the ideas I was looking at in zulip/zulip#5047.

While this issue is (currently) focused on empty messages, as per my previous PR I'm interested if this could be extended to a broader range of (typically default) messages. As in #54, I see the benefit of keeping the main handler interface minimal. However, perhaps the `ACCEPT_EMPTY_MESSAGES` or similar variable could contain that data-structure, ie. extend through composition? So in bot classes, something like `empty_message_handler = ... # type: Optional[Callable]`, or a more general form incorporating a list or dict.

Saying all that, getting empty messages handled is a great first step :)

neiljp:
@zulipbot claim

zulipbot:
Hello @neiljp, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 7 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 3 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

neiljp:
A PR for this issue was initially developed in #58, but then extended into #66, both of which are still open pending feedback (primarily focusing on #66).

zulipbot:
Hello @neiljp, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 7 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 3 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

neiljp:
[...]

PR #58, opened 2017-08-12
Handle empty messages in bot framework by default

neiljp:
This aims to implement #55.

These commits:
* Modify `lib.py` to intercept empty messages, if the bot class does not explicitly define `ACCEPT_EMPTY_MESSAGES` (and evaluates to `True`)
* Amend a local doc to mention this optional element of bot classes
* Amend the wikipedia bot, as a trial bot.

Outstanding points regarding this level of implementation:
* Other than the wikipedia bot, all other bots respond to empty messages via `lib.py`, since they've not all been updated.
* All the tests pass! This is since the tests just test the bot code, and so aren't aware of this interception.
* Is this document the accurate bot document? What about `docs/bot-guide.md` in the main repo?
* Is `ACCEPT_EMPTY_MESSAGES` a typical name/format to use in this case?

### DIFF
--- a/zulip_bots/README.md
+++ b/zulip_bots/README.md
@@ -107,9 +107,11 @@ Each bot library simply needs to do the following:
 
 - Define a class that supports the methods `usage`
 and `handle_message`.
+- Optionally add a class member `ACCEPT_EMPTY_MESSAGES`, to
+specify that empty messages will be passed to the bot.
 - Set `handler_class` to be the name of that class.
 
-(We make this a two-step process to reduce code repetition
+(We make this a three-step process to reduce code repetition
 and to add abstraction.)
 
 ## Portability
--- a/zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
+++ b/zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
@@ -18,6 +18,8 @@ class WikipediaHandler(object):
     kind of external issue tracker as well.
     '''
 
+    ACCEPT_EMPTY_MESSAGES = 1
+
     def usage(self):
         return '''
             This plugin will allow users to directly search
--- a/zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
+++ b/zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
@@ -18,7 +18,10 @@ class WikipediaHandler(object):
     kind of external issue tracker as well.
     '''
 
-    ACCEPT_EMPTY_MESSAGES = 1
+    META = {'name': 'Wikipedia',
+            'description': 'Searches Wikipedia for a term and returns the top article.',
+            'no_defaults': True,  # Let bot handle all messages
+    }
 
     def usage(self):
         return '''
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -156,6 +156,10 @@ def run_message_handler_for_bot(lib_module, quiet, config_file, bot_name):
 
     state_handler = StateHandler()
 
+    # Look in bot class to see if empty messages will be handled in this function
+    # (use inverted getattr since if handler_class accepts messages, we don't handle here)
+    handle_empty_here = not getattr(lib_module.handler_class, 'ACCEPT_EMPTY_MESSAGES', False)
+
     if not quiet:
         print(message_handler.usage())
 
@@ -196,6 +200,12 @@ def run_message_handler_for_bot(lib_module, quiet, config_file, bot_name):
             message['content'] = extract_query_without_mention(message=message, client=restricted_client)
             if message['content'] is None:
                 return
+
+        # Handle empty message directly if bot (handler_class) does not
+        # explicitly request to do so
+        if not message['content'] and handle_empty_here:
+            restricted_client.send_reply(message, "Oops. Your message was empty.")
+            return
 
         if is_private_message or is_mentioned:
             message_handler.handle_message(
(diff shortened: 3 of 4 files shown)


zulip_bots/README.md
```diff
@@ -107,9 +107,11 @@ Each bot library simply needs to do the following:
 
 - Define a class that supports the methods `usage`
 and `handle_message`.
+- Optionally add a class member `ACCEPT_EMPTY_MESSAGES`, to
+specify that empty messages will be passed to the bot.
```
derAnfaenger: It would be awesome if you could also add this to `bots_guide.md`. The guide is quite outdated, but it will be the foundation for a major refactoring.

zulip_bots/zulip_bots/lib.py
```diff
@@ -18,6 +18,8 @@
 
 from zulip import Client
 
+default_empty_message_response = "Oops. Your message was empty."
```
derAnfaenger: Maybe we can plug that string directly in the function, since it only appears one time and your comment down below explains that it is used as a default empty message response.

zulip_bots/zulip_bots/lib.py
```diff
@@ -156,6 +158,12 @@ def run_message_handler_for_bot(lib_module, quiet, config_file, bot_name):
 
     state_handler = StateHandler()
 
+    # Look in bot class to see if empty messages will be handled in this function
+    try:
+        handle_empty_here = not (lib_module.handler_class.ACCEPT_EMPTY_MESSAGES)
+    except AttributeError:
+        handle_empty_here = True
```
derAnfaenger: `handle_empty = getattr(lib_module.handler_class, 'ACCEPT_EMPTY_MESSAGES', True)`

zulip_bots/zulip_bots/lib.py
```diff
@@ -197,6 +205,12 @@ def handle_message(message):
             if message['content'] is None:
                 return
 
+        # Handle empty message directly if bot (handler_class) does not
+        # explicitly request to do so
+        if len(message['content']) == 0 and handle_empty_here:
```
derAnfaenger: `if not message['content'] and handle_empty`

zulip_bots/zulip_bots/lib.py
```diff
@@ -197,6 +205,12 @@ def handle_message(message):
...
+            restricted_client.send_reply(message, default_empty_message_response)
+            return
```
derAnfaenger: I'm not sure, but wouldn't the `return` here cause the whole script to stop?

zulip_bots/README.md
neiljp: If these commits go through, rather than the other, then I'll look into that; I wasn't sure if one was outdated, or both should be updated.

zulip_bots/zulip_bots/lib.py
neiljp: Sure; I wasn't sure if I'd end up with more strings further down the line.

zulip_bots/zulip_bots/lib.py
neiljp: Nice catch of a pattern, but if this is changed, I think it probably should be
```
handle_empty = not getattr(lib_module.handler_class, 'ACCEPT_EMPTY_MESSAGES', False)
```
That is, if it gets the attr, then True there (bot handles it) implies False to `handle_empty` (lib handles it); if not present then the result is the same as with yours.

RIght?

derAnfaenger: Aah, you're right. Complicated boolean logic :) I'd add a comment explaining what's going on here.

zulip_bots/zulip_bots/lib.py
neiljp: a) It doesn't seem to ;)
b) This return is from `handle_message()`, which is called by the lib for each message;
c) A few lines above, it returns, and not for exceptional circumstances, which indicates it should be fine.

neiljp:
Pushed some changes. Last commit is also in separate PR #66, which might be easier place to discuss.

Thanks for the review @derAnfaenger 👍

zulip_bots/zulip_bots/lib.py
neiljp: Maybe the exception approach is clearer here?
- Take inverse of the found variable
- ...or default to True

zulip_bots/zulip_bots/lib.py
derAnfaenger: All good points, somehow missed that it was in `handle_message`.

zulip_bots/zulip_bots/lib.py
derAnfaenger: I found both confusing, but I prefer `gettr` just for the sake that it is more conventional.

neiljp: Fine by me 👍

derAnfaenger:
@neiljp I like your default commands PR, will give that one a look first, so this one might actually become superfluous :)

neiljp:
As per the comment by @derAnfaenger, this is a backup PR which was the origin of the work now in #66, in case #66 is considered too extensive. I'll update this when a decision is made wrt that.

zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
```diff
@@ -18,6 +18,8 @@ class WikipediaHandler(object):
     kind of external issue tracker as well.
     '''
 
+    ACCEPT_EMPTY_MESSAGES = 1
```
showell: This should be `True`, correct?

neiljp:
Superseded by #225, so closing this previous alternative implementation.

PR #59, opened 2017-08-12
Dig a tunnel to matrix.

rht:


timabbott:
@rht can you give a bit of info on how well this works?  I've sorta been holding off on review pending more detail.

rht:
It is O(zephyr_mirror)/O(irc-mirror) in terms of availability and error handling. Any further improvements should depend on how often it is being used (demand).

rht:
Brief documentation added.

zulip/integrations/matrix/matrix_bridge.py
```diff
@@ -0,0 +1,103 @@
...
+        else:
+            content = event['type']
+        msg_data = {
+            "sender": zulip_client.email,
+            "type": "stream",
+            "to": cfg["stream"],
+            "subject": cfg["subject"],
+            "content": content
+        }
+        not_from_bot = 'body' not in event['content'] or not event['content']['body'].startswith("@**")
```
timabbott: Can you explain this?

rht: `event['content']` doesn't always have the 'body' key.
1. When it doesn't contain the key, it **certainly** does not originate from the
   bot. Additionally its use is to prevent `KeyError`.
2. When it does, if the 'body' starts with the at-mention prefix that is
   specific to Zulip user's full name, it certainly does originate from the bot.

rheaparekh:
@rht This starts throwing an error after sometime 
```
ERROR:matrix_client.client:Exception thrown during sync
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/matrix_client/client.py", line 434, in listen_forever
    self._sync(timeout_ms)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/matrix_client/client.py", line 578, in _sync
    room._put_event(event)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/matrix_client/room.py", line 311, in _put_event
    listener['callback'](self, event)
  File "matrix_bridge.py", line 46, in _matrix_to_zulip
    "content": content
UnboundLocalError: local variable 'content' referenced before assignment
``` 
and the server stops running. I am going to investigate this more. 

Also, this didn't work as bi-directional. I can see the IRC message in my Zulip channel, but it doesn't get the Zulip message in that channel to IRC.

rht:
Huh I haven't encountered the error so far. Does it persist if you use `python3`?
As with the bi-directional thing, sometimes the Matrix bot got banned. You could test at the `#python-mypy` (already defunct, replaced by a gitter channel) freenode channel.

rheaparekh:
@rht The above was mainly a problem with "leave" event not being detected. I'll test the bi-directional thing more.

timabbott:
@rheaparekh should we merge this as an experimental integration so that it's easy to submit PRs fixing specific things?

rheaparekh:
@rht Aaah you are right, It would have been the issue with the Matrix bot being banned. I tested it in #python-mypy and it is working perfectly. I think that the one thing we can focus on is how to render the messages properly (the format of messages and the changing of topics) 

Screenshots attached:

screen shot 2018-05-05 at 10 33 40 am

screen shot 2018-05-05 at 10 33 20 am

@timabbott yes, it will be better if we merge this once this PR has been rebased and I can directly add PRs to this. I think the banning of the matrix bot should be noted in the documentation, as it affects Zulip->IRC messages ( I couldn't get bi-directional messages running in #git channel)

PR #61, opened 2017-08-14
zulip_bots: Add a function to get the path to <bot_name>/README.md.

eeshangarg:
This is part of our efforts to have the documentation for a
particular bot live in this repo but still be able to render it
in the main repo (in a way similar to how we render the webhooks
docs).


eeshangarg:
This PR will need to be merged and the changes in this PR will need to be released before #6103 can be merged!

timabbott:
One thing we should think about is whether we want "readme.md" to be the filename for the bot docs.  Maybe "doc.md", similar to how the webhook docs are named, would be a better choice.  The role of the new bot docs is probably going to be less "README" in style anyway, since I imagine we plan to support the use of macros and things...

timabbott:
Merged, thanks @eeshangarg!  Could we easily change things to make `MANIFEST.in`  be generated by `setup.py` just before it's needed (And then make it a generated file and add it to `.gitignore`)?  It'd be nice if there were no need to do manual work to maintain it...

eeshangarg:
@timabbott: Ah, we probably could! I already tried but ran into some issues that needed a bit more investigation, but I'll probably pick this up fairly soon after zulip/zulip#6103 is working on my end!

timabbott:
Cool, can you open an issue for auto-generating MANIFEST.in with some notes on the plan?

PR #63, opened 2017-08-15
bots: Remove redundant newlines for giphy bot.

derAnfaenger:


timabbott:
I think the PEP8 linter should catch this; are we running it on this project?

derAnfaenger:
We are, I made a git blame on these lines and they come from a rather old commit, possibly at a time when we didn't run linter for this repo.

PR #65, opened 2017-08-15
bot tests: Make test-bots fail with proper exit code.

### LINKED ISSUE #45: Failures in test-bots don't fail the Travis build
In my [quick test of Python 3 support](https://travis-ci.org/gnprice/python-zulip-api/jobs/259568217), `tools/test-bots` printed failures but then exited with status 0, meaning success -- which means that if the `python -m unittest` run hadn't also failed, the Travis build would have cheerfully passed even though there were errors.

We should fix `tools/test-bots` to make sure it returns an exit status that reflects the success or failure of the tests. The [stdlib `unittest` docs](https://docs.python.org/3/library/unittest.html#unittest.TextTestRunner) aren't super clear on this subject, but I suspect `TextTestRunner.run()`, which we call at the end of that script, returns some kind of result object, which we can inspect that to find out the result. Then we can either `sys.exit(0)` for success or `sys.exit(1)` for failure.

derAnfaenger:
Fixes #45.

### DIFF
--- a/tools/test-bots
+++ b/tools/test-bots
@@ -4,6 +4,7 @@ from os.path import dirname, basename
 from importlib import import_module
 import argparse
 import os
+import sys
 import glob
 import unittest
 import logging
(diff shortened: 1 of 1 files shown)


timabbott:
Thanks for catching and fixing!  I suppose this made Travis CI useless :)

PR #66, opened 2017-08-15
Bots: Add metadata scheme for bots.

### LINKED ISSUE #55: Handle empty bot messages through the bots framework.

neiljp:
* Name & (short) description
* Whether to use any default commands (default to enabled):
(about, '' (empty), usage [remove?], help/commands)
* Provide a list of user commands, which help/commands uses

This is an alternative/supplement/extension of #58, which would also resolve #55.

I also raised the idea of bot metadata in #integrations.

Currently this just provides default-commands which use the simple metadata and (optionally) a list of bot commands provided by the developer. It could allow the user to provide callbacks directly, with the handle_message in the bot only being used for more complex input (like zulip/zulip#5047)


zulip_bots/zulip_bots/lib.py
```diff
@@ -156,7 +158,58 @@ def run_message_handler_for_bot(lib_module, quiet, config_file, bot_name):
...
+        ('', lambda: ("Oops. Your message was empty.", )),
+        ('about', (def_about, "The type and use of this bot")),
+        ('usage', ((lambda: message_handler.usage(), "Bot-provided usage text"))),
+        ('help', (lambda: "{}\n{}\n{}".format(def_about(), message_handler.usage(), def_help()),
+                 "This help text")),
+        ('commands', (def_commands, "A short list of supported commands"))
+    ])
+    # Update bot_details from those in class, if present
+    try:
+        bot_details.update(lib_module.handler_class.META)
```
derAnfaenger: `bot_details.update(getattr(lib_module.handler_class, 'META'. {})` ?

zulip_bots/zulip_bots/lib.py
```diff
@@ -198,6 +251,18 @@ def handle_message(message):
...
         if is_private_message or is_mentioned:
+            # Handle any default_commands first
+            if len(default_commands) > 0:
+                if '' in default_commands and len(message['content']) == 0:
+                    restricted_client.send_reply(message, default_commands[''][0]())
+                    return
+                for command in default_commands:
+                    if command == '':
+                        continue
+                    if message['content'].startswith(command):
```
derAnfaenger: Maybe we want a simple equality check here, so we can simplify the whole `# Handle any default_commands first` if clause?

zulip_bots/zulip_bots/lib.py
```diff
@@ -156,7 +158,58 @@ def run_message_handler_for_bot(lib_module, quiet, config_file, bot_name):
...
+        return ("\n".join("**{}** - {}".format(k, v[1])
+                          for k, v in default_commands.items() if k != '') +
+                "\n" +
+                "\n".join("**{}** - {}".format(k, v)
+                          for k, v in bot_details['commands'].items() if k != ''))
+    def def_commands():
+        return "**Commands**: {} {}".format(
+            " ".join(k for k in default_commands if k != ''),
+            " ".join(k for k in bot_details['commands'] if k != ''))
+    default_commands = OrderedDict([
```
derAnfaenger: Why does this need to be an ordered dict?

zulip_bots/zulip_bots/lib.py
```diff
@@ -156,7 +158,58 @@ def run_message_handler_for_bot(lib_module, quiet, config_file, bot_name):
...
+        if len(bot_details['commands']) == 0:  # No commands specified, so don't use this feature
+            del default_commands['commands']
+            del default_commands['help']
+        else:
+            for command in bot_details['commands']:  # Bot commands override defaults
+                if command in default_commands:
+                    del default_commands[command]
+    # Sync default_commands changes with bot_details
+    if len(default_commands) == 0:
+        bot_details['no_defaults'] = True
```
derAnfaenger: I've not entirely looked through this `bot_detail` `default_commands` code logic. Maybe this would be easier to understand if we outsource it to a separate function?

zulip_bots/zulip_bots/lib.py
neiljp: If these are not ordered, then when iterated in the commands and help text, the order is not guaranteed - perhaps not *vital* for regular use, but tests won't be consistent either!

zulip_bots/zulip_bots/lib.py
```diff
@@ -156,7 +158,58 @@ def run_message_handler_for_bot(lib_module, quiet, config_file, bot_name):
...
+        return "**Commands**: {} {}".format(
+            " ".join(k for k in default_commands if k != ''),
+            " ".join(k for k in bot_details['commands'] if k != ''))
+    default_commands = OrderedDict([
+        ('', lambda: ("Oops. Your message was empty.", )),
+        ('about', (def_about, "The type and use of this bot")),
+        ('usage', ((lambda: message_handler.usage(), "Bot-provided usage text"))),
+        ('help', (lambda: "{}\n{}\n{}".format(def_about(), message_handler.usage(), def_help()),
+                 "This help text")),
+        ('commands', (def_commands, "A short list of supported commands"))
```
neiljp: The outer construction is the form of the OrderedDict constructor, the value is simply lambda+Text (so tuple seems ok?); are you referring to the formatting, the data-structure, or some other aspect?

zulip_bots/zulip_bots/bots/xkcd/xkcd.py
```diff
@@ -6,6 +6,8 @@
 XKCD_TEMPLATE_URL = 'https://xkcd.com/%s/info.0.json'
 LATEST_XKCD_URL = 'https://xkcd.com/info.0.json'
 
+from collections import OrderedDict
```
neiljp: Amended, pending push.

zulip_bots/zulip_bots/lib.py
```diff
@@ -156,7 +158,58 @@ def run_message_handler_for_bot(lib_module, quiet, config_file, bot_name):
...
+    # Bot details and default commands from defaults, then override if provided
+    bot_details = { 'name': bot_name.capitalize(),
+                    'description': "",
+                    'commands': {},
+                    'no_defaults': False,
+    }
+    def def_about():
+        desc = bot_details['description']
+        return "**{}**{}".format(bot_details['name'],
+                                 "" if desc == "" else ": {}".format(desc))
```
neiljp: Interesting construction; I just played with this but it doesn't seem to work in practice as "not desc" gives True :)

Alternatively, just:
```
        if bot_details['description'] == "":
            return "**{}**".format(bot_details['name'])
        return "**{}**: {}".format(bot_details['name'], bot_details['description'])
```
?

zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
```diff
@@ -18,6 +18,11 @@ class WikipediaHandler(object):
     kind of external issue tracker as well.
     '''
 
+    META = {'name': 'Wikipedia',
+            'description': 'Searches Wikipedia for a term and returns the top article.',
+            'no_defaults': True,  # Let bot handle all messages
```
neiljp: Amended, pending push.

zulip_bots/zulip_bots/lib.py
neiljp: To clarify, do you mean something like `if default_commands == {}:`, or even `if default_commands:` ?

neiljp:
Updated to reflect most review comments; pending are final if statement and linter (since code has changed).

zulip_bots/zulip_bots/bots/xkcd/xkcd.py
```diff
@@ -14,6 +16,17 @@ class XkcdHandler(object):
...
 
+    META = {
+        'name': 'XKCD',
+        'description': 'Fetches comic strips from https://xkcd.com.',
+        'defaults': True,
+        'commands': OrderedDict([
+                ('latest', "Show the latest comic strip"),
+                ('random', "Show a random comic strip"),
+                ('<comic id>', "Show a comic strip with a specific 'comic id'"),
+        ])  # NOTE: help not listed here, so default command used
```
timabbott: I don't understand this comment.

zulip_bots/zulip_bots/lib.py
```diff
@@ -159,6 +161,48 @@ def state(self, default):
...
+        ('about', (def_about, "The type and use of this bot")),
+        ('usage', ((lambda: message_handler.usage(), "Bot-provided usage text"))),
+        ('help', (lambda: "{}\n{}\n{}".format(def_about(), message_handler.usage(), def_help()),
+                  "This help text")),
+        ('commands', (def_commands, "A short list of supported commands"))
+    ])
+    return defaults
+
+def sync_botdetails_defaultcommands(bot_details, default_commands):
+    # Update default_commands from any changes in bot_details
```
timabbott: I don't understand this function; is it just "updating" in the sense that it's overriding the global defaults with this particular bot's settings?

zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
```diff
@@ -18,6 +18,12 @@ class WikipediaHandler(object):
     kind of external issue tracker as well.
     '''
 
+    META = {
+        'name': 'Wikipedia',
+        'description': 'Searches Wikipedia for a term and returns the top article.',
+        'defaults': False,  # Let bot handle all messages
```
timabbott: I think the fact that you feel the need for a comment here suggests we might want a more verbose name than "defaults".  I think "default_commands_enabled" would be a good name.

zulip_bots/zulip_bots/lib.py
```diff
@@ -159,6 +161,48 @@ def state(self, default):
         yield new_state
         self.set_state(new_state)
 
+def setup_default_commands(bot_details, message_handler):
+    def def_about():
+        if bot_details['description'] == "":
+            return "**{}**".format(bot_details['name'])
+        return "**{}**: {}".format(bot_details['name'], bot_details['description'])
```
timabbott: This doesn't feel particularly readable.  A better approach might be to write these as as e.g.

`"**{name}**: {description}".format(**bot_details)`

and probably even better would be e.g.

`"**%(name)s**: %(description)s" % bot_details

zulip_bots/zulip_bots/lib.py
```diff
@@ -159,6 +161,48 @@ def state(self, default):
...
+
+    def def_help():
+        return ("\n".join("**{}** - {}".format(k, v[1])
+                          for k, v in defaults.items() if k) + "\n" +
+                "\n".join("**{}** - {}".format(k, v)
+                          for k, v in bot_details['commands'].items() if k))
```
timabbott: Let's use more verbose variable names than k/v here.  I think we want "command_name" and "command_description".  

I think some of the other new functions in this file could use the same sort of work.

As a general rule, one should basically always be verbose in variable names.  Brevity doesn't provide much value, and the extra clarity is huge.

Also, put the newline on its own line so it is more visible.

zulip_bots/zulip_bots/lib.py
```diff
@@ -159,6 +161,48 @@ def state(self, default):
...
+    def def_commands():
+        return "**Commands**: {} {}".format(
+            " ".join(k for k in defaults if k),
+            " ".join(k for k in bot_details['commands'] if k))
+    defaults = OrderedDict([  # Variable definition required for callbacks above
+        ('', (lambda: "Oops. Your message was empty.", "[BLANK MESSAGE NOT SHOWN]")),
+        ('about', (def_about, "The type and use of this bot")),
+        ('usage', ((lambda: message_handler.usage(), "Bot-provided usage text"))),
+        ('help', (lambda: "{}\n{}\n{}".format(def_about(), message_handler.usage(), def_help()),
+                  "This help text")),
```
timabbott: I think the lambdas here aren't super readable; I'd rather those have an actual `def` method like the others.

zulip_bots/zulip_bots/lib.py
```diff
@@ -159,6 +161,48 @@ def state(self, default):
...
+        if len(bot_details['commands']) == 0:  # No commands specified, so don't use this feature
+            del default_commands['commands']
+            del default_commands['help']
+        else:
+            for command in bot_details['commands']:  # Bot commands override defaults
+                if command in default_commands:
+                    del default_commands[command]
+    # Sync default_commands changes with bot_details
+    if len(default_commands) == 0:
+        bot_details['defaults'] = False
```
timabbott: This function would be much better as something that returned a new dictionary based on the inputs, rather than mutating the default_commands dict.  In addition to being a cleaner flow, I'm pretty sure the mutation means this code ends up being broken if multiple bots with different command sets run in the same process.

zulip_bots/zulip_bots/lib.py
timabbott: Yeah, I think we want an equality check, not startswith.  It also means `''` doesn't need to be a special case.

zulip_bots/zulip_bots/bots/xkcd/xkcd.py
neiljp: In this bot, with this PR, help is now handled internally by the lib, so the default command is used instead as it stands. If 'help' was listed in `commands` then the bot wouldn't handle it and the help text would be as specified in the tuple.

zulip_bots/zulip_bots/lib.py
neiljp: I broke out the code which adjusted `default_commands` based on `bot_details` (in some cases), but also have the last conditional which set `default_commands_enabled`. That's only (currently) used in that function, so could be cut, but that's why I prefixed it with 'sync'.

zulip_bots/zulip_bots/lib.py
neiljp: * I've generally been wary of verbosity here (not my code), perhaps because I'm aware how verbose I've been in my own code! (and I've generally gone very verbose, and wanted to fit in shortish lines here).
* I've gone for variables without the `command_` prefix, which I think provides a balanced approach.
* The newline was on it's own line in a previous version; moved it back :)

neiljp:
@timabbott Changes made in a new commit to show the changes separately, though will squash or split as necessary.

neiljp:
Current status is: (copying from zulip, to keep the context)
* I've mostly updated the metadata PR, and put the changes into a separate commit to be squashed in the final version, if that works.
* The updated (simpler) updated_default_commands approach @timabbott hinted at leaves the functions (were lambdas) pointing at the old data (notably for help & commands), so listing them in `commands` actually shows them twice (old default_commands and bot-provided commands list). This could be solved by going back to one data structure, if that could work, or some other approach involving passing in default_commands (new version) to the functions? (the alternative solution would be to just make default commands be all-or-nothing, ie. `default_commands_enabled = False` and `commands` is ignored, or `True` and they're used?)

derAnfaenger:
@neiljp I find this rather hard to review - it'd be awesome if you could split this in multiple, easily digestible commits; e.g. one for updating a bot with the feature, one for adding default commands, one for the core update command logic, etc.

Additionally, I think this would really profit from some unittests, which could for example be added in the updated bots themselves.

neiljp:
@derAnfaenger I've done a basic split into 3 parts. The last real commit may benefit from discussion/readjustment still, but the first two should be clear separate commits to review/apply as they stand.

zulip_bots/zulip_bots/lib.py
```diff
@@ -218,6 +246,16 @@ def handle_message(message):
                 return
 
         if is_private_message or is_mentioned:
+            # Handle any default commands first
+            if len(updated_defaults) > 0:
+                for command in updated_defaults:
+                    if command == '' and len(message['content']) == 0:
```
derAnfaenger: Why would we need an extra if-clause for this case; shouldn't it be caught by the following if-clause?

zulip_bots/zulip_bots/lib.py
```diff
@@ -218,6 +246,16 @@ def handle_message(message):
...
+                        restricted_client.send_reply(message, updated_defaults[''][0]())
+                        return
+                    if command == message['content']:
+                        restricted_client.send_reply(message, updated_defaults[command][0]())
```
derAnfaenger: To avoid accessing our data structures with `[0]` (which makes code harder to understand), I suggest to implement command_defaults as a dict of dicts, which could look like this:

```python
command_defaults = {
         '': {'function': default_empty_response, 'explanation': "[BLANK MESSAGE NOT SHOWN]"},
         'about': {'function': default_about_response, 'explanation': "The type and use of this bot"},
         'usage': 'function': {lambda: message_handler.usage(), 'explanation': "Bot-provided usage text"},
     }
```

With nicer line-breaks than here oc :)

This would also allow us to not loop over all commands, but to directly access a command with a key.

zulip_bots/zulip_bots/lib.py
```diff
@@ -187,7 +187,17 @@ def run_message_handler_for_bot(lib_module, quiet, config_file, bot_name):
...
+    bot_details = {
+        'name': bot_name.capitalize(),
+        'description': "",
+    }
+    bot_details.update(getattr(lib_module.handler_class, 'META', {}))
+
     if not quiet:
+        print("Running {} Bot:".format(bot_details['name']))
+        if bot_details['description'] != "":
+            print("\n\t{}".format(bot_details['description']))
```
derAnfaenger: Nit: I find the partial indentation odd-looking:

```
Running Wikipedia Bot:

        Searches Wikipedia for a term and returns the top article.

            This plugin will allow users to directly search
            Wikipedia for a specific key term and get the top
            article that is returned from the search. Users
            should preface searches with "@mention-bot".
```
What do you think?

zulip_bots/zulip_bots/lib.py
neiljp: You mean that we could remove the first `if` block and the functionality would be covered by the second? That would seem to be the case, yes; possibly it is a remnant of when I was just covering the empty case.

zulip_bots/zulip_bots/lib.py
neiljp: You mean the name vs description vs usage indentation?

If so, I previously had the (short) description indented the same as the title, but chose this layout to illustrate the 3 components. I'm happy to adjust it back, though it's just shown on the console when running the bot.

zulip_bots/zulip_bots/lib.py
neiljp: Implemented this way for now, though I would also consider a named-tuple approach.

neiljp:
@derAnfaenger Pushed updated commits based on your suggestions.

Thoughts:
- 'help' parts of the dict could be left out of the second commit (and the dict form), if wanted
- bot testing & `zulip_bot_output.py` both miss out on the changes from using this code, as it stands, so all tests will pass for these changes, but primarily since the tests don't "see" the change in code.
- the default vs extend vs overwriting of command (help) remains a little challenging, and could even be "simplified" by (optionally) allowing the lib to have callback functions for listed commands?

timabbott:
I merged the first commit.

zulip_bots/zulip_bots/lib.py
derAnfaenger: I think the components are already separated by the empty lines between them, so we should be fine using only one or two indentation levels.

neiljp:
I've refactored some of these commits to be smaller and clearer, and added some which amend the testing framework and `zulip_bot_output.py` to use the same logic as `lib.py` (before those commits, they provide different results from tests and running them, which indicates they should be in the same commit?).

I still would like to think/discuss about the way commands might work, but this is getting closer, I feel.

neiljp:
Currently the CI is failing, since I'm unsure whether to encode the complex default command responses in the tests, when they're so automated from the other data? That is, should I update the "help" text for the xkcd bot to ensure it passes, or remove that test?

zulip_bots/zulip_bots/lib.py
```diff
@@ -168,6 +168,14 @@ def is_private_message_from_another_user(message_dict, current_user_id):
         return current_user_id != message_dict['sender_id']
     return False
 
+def get_bot_details(bot_class, bot_name):
```
rht: LGTm

zulip_bots/zulip_bots/lib.py
```diff
@@ -168,10 +170,30 @@ def is_private_message_from_another_user(message_dict, current_user_id):
         return current_user_id != message_dict['sender_id']
     return False
 
+def setup_default_commands(bot_details, message_handler):
```
rht: I suppose this is the nucleus of the PR.

showell:
Hi @neiljp Can you resolve the conflicts here?

PIG208:
The concepts here are pretty valuable, but the PR has become too stale. I think we can close it but it's definitely worth doing a follow-up on this.

neiljp:
@PIG208 This PR is very old now and out of date, but my recollection was of various driving elements:
* Using the `META` tags to be able to not just know the user-name for a bot, but what bot code is running (if there is) behind it - by whatever method; some bots already have these added
* Having a layer on top of the bots to make a standard set of commands, or at least simple ways of interacting with bots, easier; so eg. `help`, `about`, ... and these could use the `META` tags (or not)

neiljp:
Note that #169 takes forward one of these points forward.

PR #68, opened 2017-08-15
bots: Fix "help" message async in yoda.py and test_yoda.py.

abhijeetkaur:
The help message string was modified in yoda.py file to correctly
instruct the user. But the help message string was not modified
accordingly in the test file.


timabbott:
Nice, merged, thanks @abhijeetkaur!

PR #71, opened 2017-08-16
testing: Make linter fail with exit code if any check fails.

derAnfaenger:
This might lead to a couple PRs failing that are currently not.


timabbott:
Sigh, I guess that's what we get for writing a semi-duplicate forked version of the original linter code: new bugs :)

Thanks for fixing!

PR #75, opened 2017-08-16
zulip_bots: Remove thesaurus bot.

eeshangarg:
This bot depends on PyDictionary, which isn't very well-implemented
or well-maintained. PyDictionary's dependency on goslate and
goslate's dependency on concurrent.futures has been known to cause
problems in Python 3 virtualenvs. This bot has also been the
source of disruptive BeautifulSoup warnings. Since this bot is only
meant to be an example bot, and for all the above reasons,
it makes sense to remove this bot. The cons of debugging the above
issues outweight the pros of having the bot at all.


timabbott:
Merged, thanks @eeshangarg!  I can imagine us resurrecting this at some point in the future to just talk to the relevant website's API directly; it's not clear we need a library for this anyway.

PR #77, opened 2017-08-17
travis: Move linter to a separate job.

derAnfaenger:
This moves the `tools/lint` command to a new Travis job. 
I tested this on my repo, both with a [build where the linter should fail](https://travis-ci.org/derAnfaenger/python-zulip-api/builds/265557352), and a [build where it should pass](https://travis-ci.org/derAnfaenger/python-zulip-api/builds/265557955): 

A second commit fixes a small issue the litner complains about.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,14 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
+matrix:
+  include:
+  - python: "3.6"
+    env: TEST_SUITE=linter
 install:
   - pip install pycodestyle==2.3.1
   - pip install ./zulip
   - pip install ./zulip_bots
   - pip install ./zulip_botserver
 script:
-  - tools/lint
-  - python -m unittest discover -v zulip_botserver
-  - tools/test-bots
+  - tools/travis_tests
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -4,12 +4,14 @@ python:
...
 install:
   - pip install pycodestyle==2.3.1
   - pip install ./zulip
   - pip install ./zulip_bots
   - pip install ./zulip_botserver
 script:
-  - tools/lint
-  - python -m unittest discover -v zulip_botserver
-  - tools/test-bots
+  - tools/travis_tests
```
timabbott: I think a cleaner model would be to have a separate script for each, sorta like we have in the zulip/zulip project, but this is probably fine when things are this simple.

timabbott:
Looking at https://travis-ci.org/zulip/python-zulip-api/builds/265559583?utm_source=github_status&utm_medium=notification, we might also want to set TEST_SUITE=main (or something) so the environment variables feel a bit more readable, which could go along with moving these into separate scripts in the style of zulip/zulip.

derAnfaenger:
Will do so! I felt that a separate file for one, respectively two commands is a little overkill, but I guess comprehensability counts more :)

timabbott:
Files are cheap; what matters is readability.  And I imagine these will probably end up growing over time anyway.

PR #78, opened 2017-08-18
Refactor provision script.

derAnfaenger:
This PR
* refactors script variables and checks for Windows/Linux differences in a more robust way.
* dynamically names the venv with the pattern <zulip-api-py<PYVERSION>-venv> and allows users to specify custom python versions.

### DIFF
--- a/tools/provision
+++ b/tools/provision
@@ -1,15 +1,28 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 BASEDIR=$(dirname `dirname $0`)
-VENVDIR="zulip-api-py2-venv"
 VENVEXECDIR="bin"
+PYVERSION=$(python -V | grep -Po '(?<=Python ).+')
+PYCUSTOMPATH_ARG=""
+while getopts "hp:" arg; do
+    case "$arg" in
+        h)
+            echo "USAGE: $0 [-p <path_to_python_version>]"
+            exit
+            ;;
+        p)
+            PYCUSTOMPATH_ARG="-p $OPTARG"
+            PYVERSION=$("$OPTARG" "-V" 2>&1 | grep -Po '(?<=Python ).+')
+            ;;
+    esac
+done
+VENVDIR="zulip-api-py$PYVERSION-venv"
 
 if [ ! -d "$BASEDIR/$VENVDIR" ]; then
-    virtualenv "$BASEDIR/$VENVDIR"
+    virtualenv $PYCUSTOMPATH_ARG "$BASEDIR/$VENVDIR"
     echo "Virtualenv created."
 fi
 
-VENVBINDIR="bin"
 if [[ ! -d "$BASEDIR/$VENVDIR/$VENVEXECDIR" ]]; then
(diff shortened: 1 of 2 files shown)


tools/provision
```diff
@@ -1,33 +1,48 @@
...
-if [ ! -d "$BASEDIR/zulip-api-py2-venv" ]; then
-    virtualenv $BASEDIR/zulip-api-py2-venv
+if [ ! -d "$BASEDIR/$VENVDIR" ]; then
+    virtualenv $PYCUSTOMPATH_ARG "$BASEDIR/$VENVDIR"
     echo "Virtualenv created."
 fi
 
-VENVBINDIR="bin"
-if [[ "$OSTYPE" == "msys" ]]; then
+if [[ ! -d "$BASEDIR/$VENVDIR/$VENVEXECDIR" ]]; then
```
timabbott: This test seems less precise than the old one; why did you change this?

timabbott: (The risk with this kind of test is that some other bug results in it being detected as Windows, causing its own problems, and then folks spend time debugging the wrong thing).

timabbott:
This looks pretty good!  I posted one comment; happy to merge after that's fixed.  

Though now I'm thinking we probably will want to rewrite this into Python, so that the process for picking Python version can be as simple as running e.g. `python3 tools/provision`, since that feels a lot cleaner.

tools/provision
derAnfaenger: "$OSTYPE" is a variable that doesn't even exist on Windows; `msys` is only set by CygWin (a terminal environment for Windows that emulates some linux behavior).
If we want to be comrehensive with  the variable check approach, we would need to check these cases:
```
        # Windows
if [[ "$OSTYPE" exists ]]; then
        # Mac OSX
elif [[ "$OSTYPE" == "cygwin" ]]; then
        # POSIX compatibility layer and Linux environment emulation for Windows
elif [[ "$OSTYPE" == "msys" ]]; then
        # Lightweight shell and GNU utilities compiled for Windows (part of MinGW)
elif [[ "$OSTYPE" == "win32" ]]; then
```
I thought it to be much easier to check on the existence of the `Scripts` directory, since that is the only difference for `venv` between Windows and Linux.

timabbott: Hmm, can you invert the check then, to see if the `Scripts` directory exists?  Basically I don't want a bug where we have `VENVDIR` wrong due to some little path issue to result in the Windows-related code path running.  So a positive check for the Windows path is better than a negative check for the Linux/Mac one.

derAnfaenger: Aah, I see, yep, that makes sense!

derAnfaenger: Shall I do that or just rewrite it in Python?

timabbott: I guess we can merge this as-is and then just plan to that model when we rewrite.

timabbott: (I think it'd be good to start working on a rewrite while this is fresh).

derAnfaenger: Cool. For #85, we will now need to set `source "zulip-api-<pythonversion>-venv/bin/activate"` dynamically for the different python version.

PR #80, opened 2017-08-18
travis: Split test jobs into separate files and add test coverage.

derAnfaenger:


derAnfaenger:
Yay, seems to work!

timabbott:
Generally looks good; can you add a comment documenting that issue?

timabbott:
Another follow up that probably makes sense is to add a coverage badge for this GitHub project in the README.md.

PR #81, opened 2017-08-18
Clean is_private function to work based on user IDs and modify interface to work for Embedded bots.

abhijeetkaur:


timabbott:
I merged the first bit of this, after some changes to variable names etc.  Want to spend some more time thinking about the rest.

showell:
@abhijeetkaur This looks great, but you need to fix a merge conflict!  Thanks!

PIG208:
We need to merge this PR after tweaking it a little bit. As embedded bots will not work in private groups otherwise.

PR #84, opened 2017-08-22
testing: Add build and coverage badges to README.md

derAnfaenger:
Note the difference between this README.md and the zulip repo's one: Here, the coverage status badge is directly obtained from codecov, whereas in zulip, it's fetched from img.shields.io. This is why this badge says "codecov", whereas zulip's badge says "coverage".

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 # Zulip API
 
+![Build status](https://travis-ci.org/zulip/python-zulip-api.svg?branch=master)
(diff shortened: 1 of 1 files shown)


timabbott:
Cool, thanks for doing this!  I'd rather it say "coverage", since advertising codecov is less important in my view than being clear about what the number means :)

Feel free to merge after changing that.

Looking at our actual coverage data, basically coverage is low because of the N bots that don't have tests yet.

derAnfaenger:
@timabbott updated.

PR #85, opened 2017-08-22
provision: Refactor .travis.yml and provision.py

derAnfaenger:
This allows us to just call tools/provision in .travis.yml,
instead of installing all packages separately.
This conflicts with #78 - merging either PR will require the other to be updated.

### DIFF
--- a/tools/provision
+++ b/tools/provision
@@ -1,48 +1,59 @@
-#!/usr/bin/env bash
+#!/usr/bin/env python
 
-BASEDIR=$(dirname `dirname $0`)
-VENVEXECDIR="bin"
-PYVERSION=$(python -V | grep -Po '(?<=Python ).+')
-PYCUSTOMPATH_ARG=""
-while getopts "hp:" arg; do
-    case "$arg" in
-        h)
-            echo "USAGE: $0 [-p <path_to_python_version>]"
-            exit
-            ;;
-        p)
-            PYCUSTOMPATH_ARG="-p $OPTARG"
-            PYVERSION=$("$OPTARG" "-V" 2>&1 | grep -Po '(?<=Python ).+')
-            ;;
-    esac
-done
-VENVDIR="zulip-api-py$PYVERSION-venv"
+import os
+import sys
+import argparse
+import pip
+import six
+import subprocess
 
-if [ ! -d "$BASEDIR/$VENVDIR" ]; then
-    virtualenv $PYCUSTOMPATH_ARG "$BASEDIR/$VENVDIR"
-    echo "Virtualenv created."
-fi
+from importlib import import_module
 
-if [[ ! -d "$BASEDIR/$VENVDIR/$VENVEXECDIR" ]]; then
-    # POSIX compatibility layer and Linux environment emulation for Windows
-    # Virtual uses /Scripts instead of /bin on Windows.
-    # Read https://virtualenv.pypa.io/en/stable/userguide/
-    VENVEXECDIR="Scripts"
-fi
+def main():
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -10,16 +10,8 @@ matrix:
...
-
-    # Install pycodestyle, the library for checking pep8 code conformity.
-  - pip install pycodestyle==2.3.1
-
-    # Install all API packages
-  - pip install ./zulip
-  - pip install ./zulip_bots
-  - pip install ./zulip_botserver
+  - tools/provision
+  - source "zulip-api-py2-venv/bin/activate"
```
timabbott: I like the idea of doing `provision` in the tests here (though be careful not to end up hardcoding the Python version to 2 by accident!), but since we don't actually need codecov in development, we should probably leave `pip install codecov` here.

derAnfaenger: My reasoning was that developer might want to check coverage increase/decrease when developing tests. Or is that use case too rare?

timabbott: Ahh, so `python-coverage` is the library for that, which would be reasonable to add to a development-only version of `requirements.txt`.  But `codecov`, the library for talking to the CodeCov server (Which probably declares a dependency on `python-coverage`), we don't need except in Travis CI.

derAnfaenger:
@timabbott updated and added a commit for rewriting the provision script. I faced some difficulties regarding the activation of the virtual environment in a python script, the comments should explain it.

timabbott:
Merged the first 2 commits.  I'd love it if you can get one of the other developers to try out the third.

tools/provision
```diff
@@ -1,48 +1,57 @@
...
+        venv_exec_dir = 'Scripts'
+    else:
+        venv_exec_dir = 'bin'
+
+    # In order to install all required packages for the venv, we need to activate it. Since
+    # the activation script sets environmental variables, it needs to be executed inline with
+    # `exec`. The arguments `globals(), locals()` are required for Python 2.7 compatibility.
+    # `exec` cannot execute shell scripts, so we need to call `activate_this.py` instead of `activate`.
+    activate_path = os.path.abspath(os.path.join(venv_dir, venv_exec_dir, 'activate_this.py'))
+    exec(compile(open(activate_path).read(), activate_path, 'exec'), globals(), locals())
```
timabbott: I think we do this in `scripts/lib/setup_venv`; does that not work?:
```
    activate_this = os.path.join(cached_venv_path, "bin", "activate_this.py")                        
    exec(open(activate_this).read(), {}, dict(__file__=activate_this))                               
```

derAnfaenger: Oh, I haven't seen that code piece. Can't tell which one is better off of my head, but I got the `exec(compile` version from a Python2/3 compatibility site, and tested with both python versions.

tools/provision
```diff
@@ -1,48 +1,60 @@
...
+import subprocess
+
+from importlib import import_module
+
+def main():
+    if len(sys.argv) > 1:
+        print("USAGE: {exec_path}\n"
+              "Creates a Python virtualenv. Its Python version is equal to "
+              "the Python version this command is executed with.".format(exec_path=__file__))
+        sys.exit(2)
```
timabbott: Use argparse with a `--help` option for this.  It's basically never cleaner to do this by hand.

derAnfaenger:
@timabbott updated :)

timabbott:
This looks great.  I merged this, with a few commits on top to fix the usage and directories.

I think a short-term follow-up is that we need to fix the fact that the virtualenv directories aren't in gitignore.

PR #87, opened 2017-08-22
set event_types to empty list if not supplied in register

### LINKED ISSUE #86: default argument to Client.call_on_each_event's event_type param should be an empty list, or set to one
Example:
``` python3
    c = zulip.Client(
        email="user@example.com",
        site="https://zulip.com/api",
        api_key="key")
    c.call_on_each_event(
        lambda msg: print(str(msg) + "\n"))
```

returns:
```
Server returned error:
event_types is not a list
```

The current argument for [call_on_each_event](https://github.com/zulip/python-zulip-api/blob/095cea7ab9518dfd9658b03df69ab876bb6e1268/zulip/zulip/__init__.py#L499)'s event_types is defined as None, but it's never set to a list within the function body. Passing an empty list for event_types works as expected.

alphor:
Resolves #86 

Be free, my tiny patch!

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -616,6 +616,9 @@ class Client(object):
         if narrow is None:
             narrow = []
 
+        if event_types is None:
+            event_types = []
+
         request = dict(
             event_types=event_types,
             narrow=narrow,


timabbott:
Merged, thanks @alphor!  It should show up in a release in pypi soon.

PR #89, opened 2017-08-23
twitter bots: Write internal config to separate file and merge twitter-bot and twitter-search-bot.

### LINKED ISSUE #88: Fix up twitter-search-bot integration for Python 3 and better state management
We've been using `twitter-search-bot` on chat.zulip.org for a long time, but with the Python 3 upgrade, it seems to be broken.

The last time I ran it, it seems to have lost its `since_id` value (And the rest of its config), because of a Python 3 compatibility exception trying to update its config file.  

We should fix this (and its `twitter-bot` friend) to use a separate file for storing `since_id`.

derAnfaenger:
Fixes #88.

While writing this I've discoverd a few discrepancies in our twitter bots:

* both `twitter-bot` and `twitter-search-bot` use the same config files `zulip_twitterrc` (and now, `zulip_twitterrc_internal` **in the same directory**. Should we name them differently?

* the [integration doc](https://chat.zulip.org/integrations/doc/twitter) for our twitter bot doesn't mention that a bot user needs to be created and a `.zuliprc` file exported to `~/.zuliprc` for the bots to function.

* the [integration doc](https://chat.zulip.org/integrations/doc/twitter) talks only about `twitter-search-bot`. Should we add a sectino for `twitter-bot`?

* the docstrings for both bots seem to be quite outdated, with unsupported arguments, and misleading names (`--user-id` is used for specifying the user name, even though `user-id` is a unique number assigned to a user by twitter).

Which of these qualify as follow-ups?


timabbott:
To be honest, it's not clear to me why we need 2 scripts, rather than an option within a single script.  So I think if we want to fix this properly, we should just merge the two and have a `--search` option.

What do you think?

zulip/integrations/twitter/twitter-bot
```diff
@@ -164,4 +165,10 @@ for status in statuses[::-1][:options.limit_tweets]:
...
         since_id = status.id
 
-write_config(config, since_id, user_id)
+if 'twitter' not in config_internal.sections():
+    config_internal.add_section('twitter')
+config_internal.set('twitter', 'since_id', str(since_id))
+config_internal.set('twitter', 'user_id', str(user))
+
+write_config(config_internal, CONFIGFILE_INTERNAL)
+write_config(config, CONFIGFILE)
```
derAnfaenger: Ha, you're right, we don't. I was too fixed on just implementing the second file :P

derAnfaenger:
@timabbott updated this with both of your suggestions (don't write the normal config file and merge both bots.) Sorry for the messed up diff view.

timabbott:
I merged the first commit (after fixing the `write_state` issue in it).  I'd love it if you can get someone else to look through this somewhat carefully since I'm kinda swamped (ideally, including testing it out in their dev environment!)

timabbott:
Any luck getting someone to test or review?  @eeshangarg do you have time?

eeshangarg:
I can review this one tonight! @timabbott :)

derAnfaenger:
@eeshangarg thanks for the review! I created [#6259](https://github.com/zulip/zulip/pull/6259) a while ago for updating the docs, can be looked at once this is merged :)

timabbott:
Merged, thanks folks!

PR #90, opened 2017-08-23
bots: Add manual command to get bot output.

abhijeetkaur:
This eliminates the need to setup dev environment and to create a bot, setup zuliprc file, subscribe the bot to the stream in order to try out a bot.
Manual command get-bot-output gives the bots response content directly.

### DIFF
--- a/zulip_bots/zulip_bots/run.py
+++ b/zulip_bots/zulip_bots/run.py
@@ -17,19 +17,19 @@ from zulip_bots.provision import provision_bot
 
 
 def import_module_from_source(path, name=None):
-    if name is None:
+    if not name:
         name = splitext(basename(path))[0]
 
     if six.PY2:
         import imp
         module = imp.load_source(name, path)
-        return module
     else:
         import importlib.util
         spec = importlib.util.spec_from_file_location(name, path)
         module = importlib.util.module_from_spec(spec)
         spec.loader.exec_module(module)
-        return module
+
+    return module
 
 
 def parse_args():
--- a/zulip_bots/zulip_bots/run.py
+++ b/zulip_bots/zulip_bots/run.py
@@ -15,7 +15,6 @@ import six
 from zulip_bots.lib import run_message_handler_for_bot
 from zulip_bots.provision import provision_bot
 
-
 def import_module_from_source(path, name=None):
     if not name:
         name = splitext(basename(path))[0]
@@ -31,6 +30,12 @@ def import_module_from_source(path, name=None):
 
     return module
 
+def is_different(given_name, path_to_bot):
+    if given_name and path_to_bot:
+        name_by_path = os.path.splitext(os.path.basename(path_to_bot))[0]
+        if (given_name != name_by_path):
+            return True
+    return False
 
 def parse_args():
     usage = '''
@@ -78,6 +83,16 @@ def parse_args():
 You must either specify the name of an existing bot or
 specify a path to the file (--path-to-bot) that contains
 the bot handler class.
+"""
+        parser.error(error_message)
+    # Checks if both name and path to bots are provided:
+    # checks if both of these are in sync, otherwise we'll
+    # have to be bias towards one and the user may get incorrect
+    # result.
+    elif is_different(options.name, options.path_to_bot):
+        error_message = """
+Please make sure that the given name of the bot and the
+given path to the bot are same and valid.
 """
         parser.error(error_message)
 
--- a/zulip_bots/zulip_bots/zulip_bot_output.py
+++ b/zulip_bots/zulip_bots/zulip_bot_output.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env python
+from __future__ import print_function
+from __future__ import absolute_import
+
+import argparse
+import sys
+import os
+from types import ModuleType
+from importlib import import_module
+from os.path import basename, splitext
+
+import six
+import mock
+from mock import MagicMock, patch
+
+from zulip_bots.lib import run_message_handler_for_bot, StateHandler
+from zulip_bots.provision import provision_bot
+from zulip_bots.run import import_module_from_source
+
+
+def parse_args():
+    usage = '''
+        zulip-bot-output <bot_name> --message "Send this message to the bot"
+        Example: zulip-bot-output xkcd --message "1"
+        This tool can be used for testing bots by sending simple messages
+        and capturing the response.
+        (Internally, this program loads bot-related code from the
+        library code and then feeds the message provided
+        in the command to the library code to handle.)
+        '''
+
+    parser = argparse.ArgumentParser(usage=usage)
+    parser.add_argument('name',
+                        action='store',
+                        nargs='?',
+                        default=None,
+                        help='the name of an existing bot to run')
+    parser.add_argument('--message',
+                        action='store',
+                        help='the message content to send to the bot')
+
+    parser.add_argument('--path-to-bot',
+                        action='store',
+                        help='path to the file with the bot handler class')
+
+    parser.add_argument('--force',
+                        action='store_true',
+                        help='Try running the bot even if dependencies install fails.')
+
+    parser.add_argument('--provision',
+                        action='store_true',
+                        help='Install dependencies for the bot.')
+
+    options = parser.parse_args()
+    if not options.name and not options.path_to_bot:
+        error_message = """
+You must either specify the name of an existing bot or
+specify a path to the file (--path-to-bot) that contains
+the bot handler class.
+"""
+        parser.error(error_message)
+
+    return options
+
+def main():
+    # type: () -> None
+    options = parse_args()
+    bot_name = options.name
+    if options.path_to_bot:
+        if options.provision:
+            bot_dir = os.path.dirname(os.path.abspath(options.path_to_bot))
+            provision_bot(bot_dir, options.force)
+        lib_module = import_module_from_source(options.path_to_bot, name=bot_name)
+    elif options.name:
+        if options.provision:
+            current_dir = os.path.dirname(os.path.abspath(__file__))
+            bots_parent_dir = os.path.join(current_dir, "bots")
+            bot_dir = os.path.join(bots_parent_dir, options.name)
+            provision_bot(bot_dir, options.force)
+        lib_module = import_module('zulip_bots.bots.{bot}.{bot}'.format(bot=bot_name))
+
+    message = {'content': options.message}
+    message_handler = lib_module.handler_class()
+
+    with patch('zulip_bots.lib.ExternalBotHandler') as mock_bot_handler:
+        mock_bot_handler.send_reply = MagicMock() 
+        message_handler.handle_message(
+            message=message,
+            bot_handler=mock_bot_handler,
+            state_handler=StateHandler()
+        )
+        print("On sending ", options.name, " bot the following message:\n\"", options.message, "\"")
+        print("\nThe bot gives the following output message:\n\"", list(mock_bot_handler.send_reply.call_args)[0][1], "\"")
+
+if __name__ == '__main__':
+    main()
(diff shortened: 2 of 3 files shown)


zulip_bots/zulip_bots/get_bot_output.py
```diff
@@ -0,0 +1,130 @@
...
+    parser.add_argument('--force',
+                        action='store_true',
+                        help='Try running the bot even if dependencies install fails.')
+
+    parser.add_argument('--provision',
+                        action='store_true',
+                        help='Install dependencies for the bot.')
+
+    options = parser.parse_args()
+    if not options.name and not options.path_to_bot:
```
derAnfaenger: Maybe another check if both are defined and don't match each other?

zulip_bots/zulip_bots/get_bot_output.py
```diff
@@ -0,0 +1,130 @@
...
+        lib_module = import_module('zulip_bots.bots.{bot}.{bot}'.format(bot=bot_name))
+
+    message = {'content': options.message}
+    message_handler = lib_module.handler_class()
+
+    with patch('zulip_bots.lib.ExternalBotHandler') as mock_bot_handler:
+        def get_config_info(bot_name, section=None, optional=False):
+            # type: (str, Optional[str], Optional[bool]) -> Dict[str, Any]
+            conf_file_path = os.path.realpath(os.path.join(
+                'zulip_bots', 'bots', bot_name, bot_name + '.conf'))
```
derAnfaenger: I think some bots use `bot_name.config`

derAnfaenger: Only the converter bot. We should update that one.

zulip_bots/zulip_bots/get_bot_output.py
```diff
@@ -0,0 +1,130 @@
...
+            section = section or bot_name
```
derAnfaenger: Maybe have section default to bot_name?

zulip_bots/zulip_bots/get_bot_output.py
```diff
@@ -0,0 +1,130 @@
...
+            config = configparser.ConfigParser()
+            try:
+                with open(conf_file_path) as conf:
+                    config.readfp(conf)  # type: ignore
+            except IOError:
+                if optional:
+                    return dict()
```
derAnfaenger: What is the purpose of this line?

zulip_bots/zulip_bots/get_bot_output.py
```diff
@@ -0,0 +1,130 @@
...
+                raise
+            return dict(config.items(section))
+
+        mock_bot_handler.get_config_info = get_config_info
+        if (hasattr(message_handler, 'initialize') and callable(message_handler.initialize)):
+            message_handler.initialize(mock_bot_handler)
+
+        mock_bot_handler.send_reply = MagicMock()
```
derAnfaenger: What about the bots that directly call `send_message`?

zulip_bots/zulip_bots/get_bot_output.py
abhijeetkaur: As a matter of fact, `converter` bot does not require a config file at all.
Will remove it.

zulip_bots/zulip_bots/run.py
```diff
@@ -73,11 +80,13 @@ def parse_args():
...
     options = parser.parse_args()
 
-    if not options.name and not options.path_to_bot:
+    if (not options.name and not options.path_to_bot) or is_not_equal(options.name, options.path_to_bot):
         error_message = """
 You must either specify the name of an existing bot or
 specify a path to the file (--path-to-bot) that contains
 the bot handler class.
+If you provide both: the name and the path to the bot,
+make sure that both point to the same bot.
```
abhijeetkaur: I could have separated out the above 4-5 lines of code in the if block to create a new function, as these lines are being used again in `zulip_bot_output.py` file.
But it doesn't make much sense to me as this error message won't be used anywhere else except for these 2 places.

I guess the modifications to create a new function for a code being used twice isn't worth it. What are your thoughts?

abhijeetkaur:
@derAnfaenger Thanks for the in-detail review. :)
I have tried to address all of your comments. It would be really great if you review once again.

zulip_bots/zulip_bots/zulip_bot_output.py
```diff
@@ -0,0 +1,124 @@
...
+
+import six
+from six.moves import configparser
+import mock
+from mock import MagicMock, patch
+
+from zulip_bots.lib import run_message_handler_for_bot, StateHandler
+from zulip_bots.provision import provision_bot
+from zulip_bots.run import import_module_from_source, is_not_equal
+
```
derAnfaenger: Redundant newline.

zulip_bots/zulip_bots/run.py
derAnfaenger: I agree, modifiying the function in run.py certainly doesn't hurt.

zulip_bots/zulip_bots/run.py
```diff
@@ -17,20 +17,27 @@
...
 
+    return module
+
+def is_not_equal(name, given_path_to_bot):
+    if name and given_path_to_bot:
+        path_to_bot_by_name = 'zulip_bots/bots/{bot}/{bot}.py'.format(bot=name)
+
+        if path_to_bot_by_name == given_path_to_bot:
+            return False
+    return True
```
derAnfaenger: How about comparing name to ` os.path.splitext(os.path.basename(path_to_bot))[0]`?

zulip_bots/zulip_bots/zulip_bot_output.py
```diff
@@ -0,0 +1,124 @@
...
+    message = {'content': options.message, 'sender_email': 'foo_sender@zulip.com'}
+    message_handler = lib_module.handler_class()
+
+    with patch('zulip_bots.lib.ExternalBotHandler') as mock_bot_handler:
+        def get_config_info(bot_name, section=None, optional=False):
+            # type: (str, Optional[str], Optional[bool]) -> Dict[str, Any]
+            conf_file_path = os.path.realpath(os.path.join(
+                'zulip_bots', 'bots', bot_name, bot_name + '.conf'))
+            section = section or bot_name
```
derAnfaenger: Should section just default to bot_name?

zulip_bots/zulip_bots/zulip_bot_output.py
```diff
@@ -0,0 +1,124 @@
...
+        message_handler.handle_message(
+            message=message,
+            bot_handler=mock_bot_handler,
+            state_handler=StateHandler()
+        )
+        print("On sending ", options.name, " bot the following message:\n\"", options.message, "\"")
+        if mock_bot_handler.send_reply.called:
+            print("\nThe bot gives the following output message:\n\"", list(mock_bot_handler.send_reply.call_args)[0][1], "\"")
+        elif mock_bot_handler.send_message.called:
+            print("\nThe bot sends the following output to zulip:\n\"", list(mock_bot_handler.send_message.call_args)[0][0], "\"")
```
derAnfaenger: I guess you can just assign one message string to some `message` variable and then do 
```
print("\nThe bot sends the following message:\n\"", list(mock_bot_handler.send_reply.call_args)[0][1], "\"")
```

zulip_bots/zulip_bots/run.py
```diff
@@ -73,11 +80,13 @@ def parse_args():
                         help='Install dependencies for the bot.')
     options = parser.parse_args()
 
-    if not options.name and not options.path_to_bot:
+    if (not options.name and not options.path_to_bot) or is_not_equal(options.name, options.path_to_bot):
```
derAnfaenger: This logic fails for me when I try to only specify a bot name: In such a case, `options.name` is not equal to `options.path_to_bot` (which is probably None).

zulip_bots/zulip_bots/run.py
abhijeetkaur: Sure! Makes sense.

zulip_bots/zulip_bots/zulip_bot_output.py
abhijeetkaur: I am not really sure as to what `section` variable does here?

Could you please throw some light on its usage. :)

zulip_bots/zulip_bots/zulip_bot_output.py
abhijeetkaur: As a matter of fact, I have purposefully kept the 2 strings different. Although very slight difference.
`
if mock_bot_handler.send_reply.called:
            print("\nThe bot gives the following output message:\n\"", list(mock_bot_handler.send_reply.call_args)[0][1], "\"")
`
Just outputs the message content.

However,
`
elif mock_bot_handler.send_message.called:
            print("\nThe bot sends the following output to zulip:\n\"", list(mock_bot_handler.send_message.call_args)[0][0], "\"")
` 
prints the entire message with the stream name (for followup bot)

abhijeetkaur: I focussed all my thoughts on the commit message: https://github.com/zulip/python-zulip-api/pull/90/commits/d447d33e79c09fffb1c8522579f75ee3245140a8 .

I'll add a comment for the same in the file. Thanks for pointing it out. :)

derAnfaenger: Hmm, I see. What's the advantage of handling `send_reply` and `send_message` differently in this way?

zulip_bots/zulip_bots/zulip_bot_output.py
derAnfaenger: Oh, I didn't see that `bot_name` is a parameter as well, nevermind. Sorry for the confusion :)

zulip_bots/zulip_bots/zulip_bot_output.py
abhijeetkaur: When `send_reply` is called from any bot's code, it simply calls it by passing (message, response).
Here message is the origin message sent by the user and `response` is the bot's reply after processing message['content']. The bot does not interfere on deciding if this bot response will be private or in a stream, or which topic. The reply is simply sent through the channel we received 'message'.

So, if we want to test bot output, simply the message will suffice in case `send_reply` is called (as the bot does not interfere with any other message fields apart from content.


This is not the case with `send_message`. If `send_message` is directly called by any bot's code, then it can edit/modify all the fields of the response message, hence we need to output all the fields to check its output.
This can simply be understood by taking an example of `follow_up` bot. It uses `send_message` to send message to a different stream with 'sender's name' as the topic.

I hope I have put forward my thoughts clearly. Makes sense?

derAnfaenger: Makes sense, thanks for the detailed explanation! I guess  I'll understand all of this better once I've actually run the command successfully :)

abhijeetkaur: So, if we output all the fields of the reply message for the bot's that use 'send_reply', that would simply increase our effort of inputting them as well. These do not play any role in the bot logic, so it will basically be useless in our testing aka getting the bot output.

However this does play in bots making use of `send_message`, as while testing if the bot is editing/modifying some response fields, we'll want to know that.

Does it make sense? :)

derAnfaenger:
@timabbott tested, LGTM! I think this is stable enough so we can merge this, and others can work on follow-ups. Thanks again to @abhijeetkaur for working through all those suggestions and writing this very helpful script!

derAnfaenger:
Some follow-ups I can think of:
* instead of extra-checking for equality in case of two specified botnames (one path and one name), we might just want to mutually exclude these arguments

* there is a little extra space next to the output's quotation marks because of the list we pass to print,
  should remove that.

* instead of copying ExternalBotHandler's get_config_info method, we should use the original one by setting Mock's wrap argument.

timabbott:
Sounds good, I agree with all of those follow-ups.  I'd add:
* It might make sense to have at least a symlink to the tool under `tools/` so it's more obviously available as a developer tool.
* We should document it.

I merged this, after renaming `is_different` to `name_and_patch_match` for clarity, and shortening the super-long comment about `send_reply` vs. `send_message`.

Thanks @abhijeetkaur and @derAnfaenger!

PR #97, opened 2017-08-25
api: Display help on wrong command-line arguments.

### LINKED ISSUE #93: Make invalid arguments to API examples do something reasonable
If you run one of our API examples with an invalid argument, like this:
```
(zulip-api-py-venv) tabbott@zaset:~/python-zulip-api$ ./zulip/examples/send-message --foo
usage: send-message [options] <recipients>

Sends a test message to the specified recipients.

Example: send-message --type=stream commits --subject="my subject" --message="test message"
Example: send-message user1@example.com user2@example.com
send-message: error: unrecognized arguments: --foo
```

This doesn't explain what options do work, and thus requires that the user know that --help is the magic option for getting help).  Ideally, we'd figure out some argparse configuration option that makes this work properly.

See more discussion in #1 for details on the motivation.

derAnfaenger:
Fixes #93.

Patching ArgumentParser.error is the only way we can call parser.print_help instead of parser.print_usage. The monkeypatch allows us to not have to modify all the various scripts using argparse, although we might want to rename `add_default_arguments` if this is merged.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -33,6 +33,7 @@ import optparse
 import argparse
 import platform
 import random
+import types
 from distutils.version import LooseVersion
 
 from six.moves.configparser import SafeConfigParser
@@ -104,8 +105,16 @@ def _default_client():
     # type: () -> str
     return "ZulipPython/" + __version__
 
-def add_default_arguments(parser):
+def add_default_arguments(parser, patch_error_handling=True):
     # type: (argparse.ArgumentParser) ->  argparse.ArgumentParser
+
+    if patch_error_handling:
+        def custom_error_handling(self, message):
+            import pprint
+            self.print_help(sys.stderr)
+            self.exit(2, '{}: error: {}\n'.format(self.prog, message))
+        parser.error = types.MethodType(custom_error_handling, parser)
+
     group = parser.add_argument_group('Zulip API configuration')
     group.add_argument('--site',
                        dest="zulip_site",


zulip/zulip/__init__.py
```diff
@@ -106,6 +107,12 @@ def _default_client():
 
 def add_default_arguments(parser):
     # type: (argparse.ArgumentParser) ->  argparse.ArgumentParser
+
+    def custom_error_handling(self, message):
+        self.print_help(sys.stderr)
+        self.exit(2, '{prog}: error: {message}\n'.format(self.prog, message))
+    parser.error = types.MethodType(custom_error_handling, parser)
```
timabbott: I would make this behavior be controlled by a kwarg, e.g. if you specify `add_default_arguments(parser, patch_error_handling=False)`, it won't do this.  Because I think this done right will be better than the default for most users.

That said, I think the actual output of this version is busted:

```
$ ./zulip/examples/send-message --asdf
...
  --client-cert-key CLIENT_CERT_KEY
                        Specify a file containing the client certificate's key
                        (if it is in a separate file).
  File "./zulip/examples/send-message", line 41, in <module>
    options = parser.parse_args()
  File "/usr/lib/python2.7/argparse.py", line 1704, in parse_args
    self.error(msg % ' '.join(argv))
  File "/home/tabbott/python-zulip-api/zulip/zulip/__init__.py", line 113, in custom_error_handling
    self.exit(2, '{prog}: error: {message}\n'.format(self.prog, message))
KeyError: 'prog'
```

derAnfaenger:
@timabbott updated.

timabbott:
Merged, after getting rid of the `pprint` import, which I imagine was left over from debugging.  

@derAnfaenger I encourage you to get into the habit of code reviewing your own commits to catch things like this :)

Thanks @derAnfaenger!

derAnfaenger:
@timabbott yeah, I should check them more carefully. Do you think we could add `pprint` to the linter (at least for this repo)? It doesn't seem to be used anywhere.

timabbott:
Yeah, that'd be reasonable.

PR #98, opened 2017-08-25
api: Add script to print path of examples directory.

### LINKED ISSUE #92: Have an easy way to find the examples from our documentation
Because of the move to `pip install zulip` (instead of having people clone the repo or download a tarball), it's now very easy to not find the API examples that come with the package.  We should probably do some combination of:
* Adding something to our API docs that point to the example code (E.g. just a link to the directory on GitHub)
* Having a recommended command that you can run after you've done `pip install zulip` that will find the example code.  This is tricky because the files will be in some crazy path under the virtualenv's share/ directory (if you're using one) or maybe `/usr/local/share` if you installed as root or something.

(Also, I actually can't find the examples in our virtualenvs generated by `tools/provision`; are they not included in the package?)

derAnfaenger:
Partially fixes #92.

I've looked at both `site.getsitepackages()` and `distutils.sysconfig.get_python_lib()`, and both have various in that they don't check all possible paths for the zulip package's installation, or don't work in venvs...

So I thought the easiest and most robust way to do this is to import the zulip package and check it's path.


derAnfaenger:
Updated; Added new commits to move the examples directory to `zulip/examples` and have it installed in the package's directory. Now, the `zulip-api-examples` script prints the correct examples path for both installing with and without `-e`.

derAnfaenger:
I wonder if we should set the threshold for codecov to trigger a little higher than in `zulip`, since this package contains much less code.

timabbott:
Looks good, merged.  @eeshangarg can you double-check that the `pip`  packaging will do the right thing here (i.e. the path is correct no matter how you install it)?

We should probably make this print the path to that example:
```
(zulip-api-py2-venv) tabbott@zaset:~/python-zulip-api$ zulip-api-examples create-user
/home/tabbott/python-zulip-api/zulip/zulip/examples
```

@derAnfaenger as a follow-up, we should document this feature :).

PR #99, opened 2017-08-28
api: Ensure automatic tilde expansion.

derAnfaenger:
When working in vagrant with the zulip botserver, I faced a subtle bug: specifying `--config-file ~/flaskbotrc` did *not* automatically expand the tilde (I don't know why, though). Since it does not get expanded by `os.path.abspath` and other functions, the file coud not be found. The same problem occurs in any terminal when putting the argument in quotes (`--config-file "~/flaskbotrc"`).

This commit:
* automatically expands any tilde with `os.path.expanduser` in the `Client` object and for the botserver.
* prints a comprehensible warning when the specified config file is not found in the botserver (until now, only a BadRequest was being thrown, leading to a `400` error message)

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -246,6 +246,9 @@ class Client(object):
         if client is None:
             client = _default_client()
 
+        # Normalize user-specified path
+        if config_file is not None:
+            config_file = os.path.abspath(os.path.expanduser(config_file))
         # Fill values from Environment Variables if not available in Constructor
         if config_file is None:
             config_file = os.environ.get("ZULIP_CONFIG")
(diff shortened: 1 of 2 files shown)


derAnfaenger:
Another problem I discovered while debugging this:

The botserver returns verbose errors like

```.py
return BadRequest("Can't find the configuration or Bot Handler code for bot {}. "
                          "Make sure that the `zulip_bots` package is installed, and "
                          "that your flaskbotrc is set up correctly".format(bot))
```

These are not printed in the botserver's log (just the status code of the BadRequest is logged). Instead, the message is added to the content of the botserver's response, like this:
```.html
'<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">\n<title>400 Bad Request</title>\n<h1>Bad Request</h1>\n<p>Can\'t find the configuration or Bot Handler code for bot encrypt. Make sure that the `zulip_bots` package is installed, and that your configuration file is set up correctly.</p>\n'
```
We don't log or print these messages, only the error code. This means that the whole text is effectively never able to be read by anyone.

timabbott:
Nice, merged, thanks for finding and fixing this @derAnfaenger!  

Yow, that botserver issue seems clearly broken.  Let's move it to a log file (though I guess that's it's own issue).

I think we're getting to the point where it might be worth starting to build a bit of a unit test suite for both the API bindings and the botserver, so that we can lock down behaviors we like.  E.g. this config parsing stuff is generally super easy to unit-test and super annoying when it's buggy.

PR #100, opened 2017-08-29
zulip_botserver: Add skeleton config file for supervisord.

derAnfaenger:
For https://github.com/zulip/zulip/pull/6301.


timabbott:
Merged, after setting a default for the log file.  Thanks @derAnfaenger!  

I think we might want to think a bit more for a v2 of this about having some sensible defaults for the `--config-file` etc. configuration.  I guess one specific detail is that I think by default this will run the service as root; we may want it to run as a "zulip_botserver" user, document `adduser zulip`, and then have all the logging and config files lives in the `zulip_botserver` user's home directory.

PR #102, opened 2017-08-30
Refactor zulip_bots scripts.

derAnfaenger:
Originally, this was intended to refactor the zulip_bot_output.py script. While working on it, I found parallels to run.py and a little inconsistency in lib.py.

Concerning followups on zulip_bot_output:
* I cannot create a symlink as we need it on Windows, someone else would have to do that.
* I'll work on the documentation tomorrow.

### DIFF
--- a/zulip_bots/zulip_bots/run.py
+++ b/zulip_bots/zulip_bots/run.py
@@ -76,9 +76,9 @@ def parse_args():
     parser.add_argument('--provision',
                         action='store_true',
                         help='Install dependencies for the bot.')
-    options = parser.parse_args()
+    args = parser.parse_args()
 
-    if not options.name and not options.path_to_bot:
+    if not args.name and not args.path_to_bot:
         error_message = """
 You must either specify the name of an existing bot or
 specify a path to the file (--path-to-bot) that contains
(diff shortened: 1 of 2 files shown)


timabbott:
I merged the first commit, after fixing the missing whitespace around `+`.  

The second commit seems to remove the feature @eeshangarg added a few weeks ago to support running the bot framework with an out-of-tree bot (e.g. imagine you want to write your own bot using this framework; you might create a new package that depends on `zulip_bots`in a separate Git repo and wanted to use the `zulip_bots` framework to run it).  I don't think we want tor remove that feature.  We might want to replace that change with documenting the feature's purpose.

derAnfaenger:
@timabbott regarding the `--path-to-bot` feature, it's still present - instead of explicitly adding it, you could now just use a path instead of a bot's name. This would also resolve the issue of a double-specification of bots (with a bot name and path). Or is there a case where we would still need an explicit `--path-to-bot` argument?

timabbott:
Ahh, OK, I missed that.  This may also be a PR where you would benefit from splitting your changes into a series of small, specific changes, each of which explains in the commit message why it is correct if it isn't obvious.  I think you have enough experience with making things work that this should be a focus of your development efforts, since it's the hallmark of an experienced developer to be able to make a series of 5-10 changes in a way that a reviewer can easily understand, rather than having them be sorta squished together, and it will also save both of us time getting changes like this integrated.  I would encourage you to err on the side of too-small commits.  It's pretty easy to split up a branch like this into pieces through the following flow:
* `git rebase -i upstream/master^`
* Select Edit on the top commit (i.e. current master)
* Re-do the changes you want to be the first commit
* `git commit -a` to commit it
* `git rebase --continue`
* When you almost certainly get a merge conflict, use `git checkout <branchname> .` to just set the tree to match the pre-rebase state of your branch (there's a few other ways to do this as well).  I recommend using `gitk` or a similar graphical Git viewer to observe this.

http://zulip.readthedocs.io/en/latest/version-control.html is a helpful guide for this.  

Does that sound good?

derAnfaenger:
Sound very good, I'll try my best! :)

derAnfaenger:
@timabbott updated, I hope this is clearer.

derAnfaenger:
Regarding commit "zulip_bot_output.py: Make --path-to-bot argument positional", I feel it'd be optimal to split the help message refactor from the rest, but `git add --patch` gave me a pretty horrible and complicated changeset, and I felt that it'd be less prone to errors if I just make both changes at once.

timabbott:
Yeah, this is a lot nicer.  Merged, thanks @derAnfaenger!  

I think next we should make sure we have a test for the out-of-tree bot use case; I think we could do some mock file access to pretend to help write that test?  I'm pretty sure there's a nice way to do that in Python, though `mktemp` is always an option.

Also, one detail that seems worth checking: if there's a file in the current directory called "foo" and a bot in the Python namespace named "foo", which wins?  I think we probably want the bot to win if you typed `foo` (so you'd need to type `./foo` to get the directory).

PR #103, opened 2017-08-31
log2zulip: Clean-up and make Windows-proof.

derAnfaenger:


zulip/integrations/log2zulip/log2zulip
```diff
@@ -42,10 +41,10 @@ def send_log_zulip(file_name, count, lines, extra=""):
...
     content = "%s new errors%s:\n```\n%s\n```" % (count, extra, "\n".join(lines))
     zulip_client.send_message({
-        "type": "stream",
-        "to": "logs",
-        "subject": "%s on %s" % (file_name, platform.node()),
-        "content": content,
+        'type': 'stream',
+        'to': 'logs',
+        'subject': "%s on %s" % (file_name, platform.node()),
+        'content': content,
```
timabbott: why did you change quote style here?  There's no advantage to `'` over `"`, and it adds a lot of extra churn to this diff.

zulip/integrations/log2zulip/log2zulip
```diff
@@ -67,47 +66,53 @@ def process_lines(raw_lines, file_name):
 def process_logs():
     # type: () -> None
     for filename in log_files:
-        data_file_path = "/var/tmp/log2zulip.state"
+        data_file_path = os.path.join(tempfile.gettempdir(), 'log2zulip.state')
```
timabbott: So, there's a reason we were using `/var/tmp`, rather than `/tmp`, which is that `/tmp` gets cleared on boot (which would meant the bot would lose its state and thus repeat messages or something if the server reboots)!  

So I think we might just want it to accept a path to its state directory (to support Windows) and have the default remain as-is.

zulip/integrations/log2zulip/log2zulip
```diff
@@ -67,47 +66,53 @@ def process_lines(raw_lines, file_name):
...
-    open(data_file_path, "w").write(ujson.dumps(new_data))
+    open(data_file_path, 'w').write(json.dumps(new_data))
 
-if __name__ == "__main__":
+if __name__ == '__main__':
+    parser = zulip.add_default_arguments(argparse.ArgumentParser())  # type: argparse.ArgumentParser3
+    parser.add_argument('--control-path', default='/etc/log2zulip.conf')
+    args = parser.parse_args()
+    if not args.zulip_config_file and os.name == 'posix':
+        args.zulip_config_file = '/etc/log2zulip.zuliprc'
```
timabbott: Do you think this might be cleaner done by modifying `add_default_arguments` to support a `default_config_file_path` argument?

zulip/integrations/log2zulip/log2zulip
derAnfaenger: Hm, I did it because I usually use single quotes for non-output strings. I guess I was a little over-confident.

zulip/integrations/log2zulip/log2zulip
derAnfaenger: It already does, right? `zulip_config_file` is the parameter for `--config-file`.

derAnfaenger: So what I tried to do here is to use the default values for the `zuliprc` in any case possible: Only if we are on a posix system, the path should default to `/etc/log2zulip.zuliprc`, so that folks updating this script don't have to reconfigure it? Is that what you meant?

derAnfaenger:
@timabbott updated with cleaner commits and `/var/tmp` for posix.

zulip/integrations/log2zulip/log2zulip
```diff
@@ -94,20 +94,28 @@ def process_logs():
...
-    open(data_file_path, "w").write(ujson.dumps(new_data))
+    open(data_file_path, "w").write(json.dumps(new_data))
 
 if __name__ == "__main__":
+    parser = zulip.add_default_arguments(argparse.ArgumentParser())  # type: argparse.ArgumentParser3
+    parser.add_argument("--control-path", default="/etc/log2zulip.conf")
+    args = parser.parse_args()
+    # On posix systems, we set the config directory explicitly for legacy reasons.
+    if not args.zulip_config_file and os.name == "posix":
+        args.zulip_config_file = "/etc/log2zulip.zuliprc"
```
timabbott: Why the `/etc/log2zulip.zuliprc` path?  I guess I don't see the point of this when we already have `/etc/log2zulip.conf`?  If there's a good reason, this sort of thing should go into the commit message...

timabbott: I guess I see the old code did that, but I don't see any real benefit?  Might want to just drop the second path.

derAnfaenger: Hm, I think one benefit of the old code is that you have your `zuliprc` file separated from the `log2zulip` configuration, meaning you could just download it as usual without modifying, or reuse an existing one. So I'd actually opt for keeping it this way.

If we drop it, would the `zuliprc` info rest in `log2zulip.conf`, or the other way around?

timabbott:
Merged, thanks @derAnfaenger!  I added a commit on top removing the (I think unnecessary) second config path.  I don't really see any benefit to having more than 1 file for managing the configuration of this tool.  

I guess we should open an issue to document this as a proper integration?  I don't think it has docs.  Presumably the docs should show how to run it and also have a recommended cron job and sample config for using it.

derAnfaenger:
Sure. Will open one.

PR #104, opened 2017-08-31
zulip package: Add requirements.txt for each integration.

derAnfaenger:


derAnfaenger:
I think we should not include these requirements in `setup.py` for the `zulip` package, since many of these are very  integration-specific and won't be needed for most people installing this.

timabbott:
A few thoughts:
* I think we should probably also not include the dependencies of `zulip` itself in these `requirements.txt` files.  We have those in the requirements files for the zulip package itself, which is where one would get the scripts from anyway. 
* I'm not sure for this application that we want to pin exact versions, as opposed to minimum versions.
* `ujson` lives in the `ujson` package, but we should probably replace use of it in this package with `json` anyway.

derAnfaenger:
Yea, `ujson` isn't used anywhere besides `log2zulip`, which I submitted a PR for to change that.

derAnfaenger:
@timabbott updated to not include zulip and specify minimum versions.

derAnfaenger:
Added two more commits for installing dependencies via `--provision` and irc-mirror as a POC. I don't know how close this is to what you had in mind, but this is what the output looks like now for `python irc-mirror --provision`:
image
The additional exception traceback is a little noisy in case you just wanted to provision without running the script, so I added color highlighting for  `Provisioning successful`.

timabbott:
So, `typing` and `six` are dependencies of `zulip`, so they don't need to be listed either.

zulip/zulip/__init__.py
```diff
@@ -221,6 +228,13 @@ def generate_option_group(parser, prefix=''):
 
 def init_from_options(options, client=None):
     # type: (Any, Optional[str]) -> Client
+
+    if options.provision:
+        requirements_path = os.path.abspath(os.path.join(sys.path[0], 'requirements.txt'))
+        if not pip.main(['install', '--upgrade', '--requirement', requirements_path]):
+            print("{color_green}Provisioning successful.{end_color}".format(
+                color_green='\033[92m', end_color='\033[0m'))
```
timabbott: Do you think we want to move the `import pip` into this block, so that we don't add a dependency of the entire library on `pip` being installed?  I guess I'm not entirely sure whether `pip` is ever going to not be installed on a system, but it's not part of Python core.  I guess my idea for how best to do this block would be to add a `try/except` around importing `pip`, with a nice error message if not.  

@eeshangarg FYI.

derAnfaenger:
@timabbott updated.

* removed the `six` and `typing` dependencies, as well as `simplejson`, since it comes with zulip as well.

* move the `import pip` line and added a check.

* refactored the last commit for allowing provisioning in the irc mirror into three separate commits. For other integrations, I guess this should go easier, since the difficulty here was to outsource the IRCBot class.

derAnfaenger:
I tested the IRC bot only in one direction (IRC->Zulip), but it didn't work for me until removing 
`"forged": "yes",`
from ```python
        print(self.zulip_client.send_message({
            "forged": "yes",
            "sender": sender,
            "type": "stream",
            "to": stream,
            "subject": "IRC",
            "content": content,
        }))
```
I was using a generic bot. What is the purpose of the `forged` key?

timabbott:
The "forged" key lets the IRC bot forge messages as having been sent by the "actual sender" from the IRC world, not the user who's running the bot.  So that's as expected, but I think we might want to change the error message to explain this to people?

You need to make an api_super_user bot to have permission to use it.

timabbott:
Looks great, merged, thanks @derAnfaenger!  

The one thing that I'm thinking we should change in the system here is that probably `--provision` should `sys.exit(0)`, rather than continuing to start the bot afterwards, since if the user hasn't configured it yet, the actual bot startup will crash, which will result in a confusing traceback.

derAnfaenger:
@timabbott thanks for the explanation. 

>  --provision should sys.exit(0), rather than continuing to start the bot afterwards

should that also happen when we specify all the other options necessary to run a bot? Or should we make provision exclusive from all other options?

timabbott:
I think it's probably most in line with expectations for `provision` to be basically exclusive of the other options (maybe not `--verbose`, if we add one of those at some point).  But I'm not sure it's worth adding code to do error handling checking for that.

derAnfaenger:
At least there doesn't seem to be a nice way of doing it with argparse (only provides `add_mutually_exclusive_group`). So I guess we can just exitafter provision, maybe with a little anecdote `you can now run <command name>`?

timabbott:
I think we can just have it end with a `You successfully provisioned the dependencies for converter` notice, and nobody will be confused, especially if the `--help` documentation is clear that the `--provision` option causes it to use `pip` to install the bot's dependencies and the markdown documentation shows running 2 commands, first the `--provision` thing, and then the command to run the bot.

PR #105, opened 2017-09-01
provision: Add color formatting to success output.

derAnfaenger:


timabbott:
I'd consider indenting the `source` line as well 2-4 spaces, but this is definitely better, so merging.

derAnfaenger:
I'm not sure about the indentation: For my terminal, teh command right now fills up the whole line; adding one space leads to a line-break.

timabbott:
OK, that's fair.  Though maybe argues we should have a less long path name for the virtualenv...

PR #108, opened 2017-09-04
zulip package: Add unittest support

derAnfaenger:


tools/test-zulip.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python
```
derAnfaenger: I first had this script named `test-zulip-api.py`, but `test-zulip` feels better when considering the names `test-bots` and `test-botserver` (all names match their corresponding package names)

timabbott:
Looks great, merged, thanks @derAnfaenger!  

In terms of actually doing test coverage work, what I'd do is write a bunch of unit tests for `Client()` / `init_from_options()` that looks vaguely like this:

```
def test_client_cert_option(self):
    c = do_test_options(["--client-cert", "/tmp/a"])
    self.assertEqual(c.client_cert, "/tmp/a")
    c = do_test_options([], dict(ZULIP_CERT="/tmp/b"))
    self.assertEqual(c.client_cert, "/tmp/b")
    c = do_test_options(["--client-cert", "/tmp/a"], dict(ZULIP_CERT="/tmp/b"))
    self.assertEqual(c.client_cert, "/tmp/a")
```

(might make sense to have a third argument for the mock config_file state).

Which would in this case verify that there's an option and an environment variable and the option takes precedence.  Since that's such a common pattern, it might make sense to have a higher-level test function where you just pass it the option name and environment variable name and it just does the above for us.  We'll ideally want a way to provide a mock config_parser as well.  

I'd write tests for exactly 2 of these options so we can see the patterns, and then we can iterate on the framework until it feels really clean, and then fill out the rest.

PR #110, opened 2017-09-07
zulip_bots: Add unittests for run.py.

derAnfaenger:
These are tests that assert the correct argument
parsing; the successful execution of bots is not verified.

Writing these tests made me aware of an import flake for Python 3.0 - 3.4; this is fixed by the first commit.

### DIFF
--- a/zulip_bots/zulip_bots/run.py
+++ b/zulip_bots/zulip_bots/run.py
@@ -10,8 +10,6 @@ from types import ModuleType
 from importlib import import_module
 from os.path import basename, splitext
 
-import six
-
 from zulip_bots.lib import run_message_handler_for_bot
 from zulip_bots.provision import provision_bot
 
@@ -21,7 +19,9 @@ def import_module_from_source(path, name=None):
     if not name:
         name = splitext(basename(path))[0]
 
-    if six.PY2:
+    # importlib.util.module_from_spec is supported from Python3.5
+    py_version = sys.version_info
+    if py_version.major < 3 or (py_version.major == 3 and py_version.minor < 5):
         import imp
         module = imp.load_source(name, path)
     else:
--- a/zulip_bots/zulip_bots/test_run.py
+++ b/zulip_bots/zulip_bots/test_run.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python
+from __future__ import absolute_import
+
+import importlib
+import os
+import zulip_bots.run
+import six
+import unittest
+import zulip
+
+from importlib import import_module
+from unittest import TestCase
+
+if six.PY2:
+    import mock
+    from mock import patch
+else:
+    from unittest import mock
+    from unittest.mock import patch
+
+class TestDefaultArguments(TestCase):
+
+    our_dir = os.path.dirname(__file__)
+    path_to_bot = os.path.abspath(os.path.join(our_dir, 'bots/giphy/giphy.py'))
+
+    @patch('sys.argv', ['zulip-run-bot', 'giphy', '--config-file', '/foo/bar/baz.conf'])
+    @patch('zulip_bots.run.run_message_handler_for_bot')
+    def test_argument_parsing_with_bot_name(self, mock_run_message_handler_for_bot):
+        zulip_bots.run.main()
+        mock_run_message_handler_for_bot.assert_called_with(bot_name='giphy',
+                                                            config_file='/foo/bar/baz.conf',
+                                                            lib_module=mock.ANY,
+                                                            quiet=False)
+
+    @patch('sys.argv', ['zulip-run-bot', path_to_bot, '--config-file', '/foo/bar/baz.conf'])
+    @patch('zulip_bots.run.run_message_handler_for_bot')
+    def test_argument_parsing_with_bot_path(self, mock_run_message_handler_for_bot):
+        zulip_bots.run.main()
+        mock_run_message_handler_for_bot.assert_called_with(
+            bot_name='giphy',
+            config_file='/foo/bar/baz.conf',
+            lib_module=mock.ANY,
+            quiet=False)
+
+if __name__ == '__main__':
+    unittest.main()


zulip_bots/zulip_bots/run.py
```diff
@@ -19,7 +19,9 @@ def import_module_from_source(path, name=None):
     if not name:
         name = splitext(basename(path))[0]
 
-    if six.PY2:
+    # importlib.util.module_from_spec is supported from Python3.5
+    py_version = sys.version_info
+    if py_version.major < 3 or (py_version.major == 3 and py_version.minor < 5):
         import imp
```
eeshangarg: Ah, nice catch! I wrote the original code and I missed this. Thanks for taking care of this! Looks good to me. :)

eeshangarg: One little piece of feedback. I think the commit message for this change could be more informative. `zulip_bots: Fix run.py import for Python < 3.5.` does not seem specific enough. I think a better message would be something along the lines of `zulip_bots: Import imp.load_source for Python < 3.5.`. It would be great if you could please amend the commit message. :)

eeshangarg: Also, I think we can get rid of line 13 `import six` now that we don't use it.

zulip_bots/zulip_bots/test_run_py.py
```diff
@@ -0,0 +1,46 @@
...
+    @patch('zulip_bots.run.run_message_handler_for_bot')
+    def test_argument_parsing_with_bot_path(self, mock_run_message_handler_for_bot):
+        zulip_bots.run.main()
+        mock_run_message_handler_for_bot.assert_called_with(bot_name=None,  # TODO: This should be equal to the bot's name
+                                                            config_file='/foo/bar/baz.conf',
+                                                            lib_module=mock.ANY,
+                                                            quiet=False)
+
+if __name__ == '__main__':
+    unittest.main()
```
eeshangarg: I ran these tests on my machine and they passed (yay)! One thing that concerns me is the output of the tests; it isn't very informative:
```
..
Ran 2 tests in 0.004s

OK
```

I think we should increase the verbosity of the test runner to display the names of the passing (or failing) tests (a convention we seem to follow a lot in our tests). See [here](https://github.com/zulip/python-zulip-api/blob/master/tools/test-bots#L95). On a similar note, I don't see these tests being run anywhere as part of Travis or anything, so I think we should do that too! :)

Perhaps another thing we need to discuss (not as a part of this PR) is whether we need to "discover" these tests in `tools/test-bots` and run it from a master "test all" script for this package. :)

zulip_bots/zulip_bots/test_run_py.py
```diff
@@ -0,0 +1,46 @@
...
+                                                            quiet=False)
+
+    @patch('sys.argv', ['zulip-run-bot', '--path-to-bot', path_to_bot, '--config-file', '/foo/bar/baz.conf'])
+    @patch('zulip_bots.run.run_message_handler_for_bot')
+    def test_argument_parsing_with_bot_path(self, mock_run_message_handler_for_bot):
+        zulip_bots.run.main()
+        mock_run_message_handler_for_bot.assert_called_with(bot_name=None,  # TODO: This should be equal to the bot's name
+                                                            config_file='/foo/bar/baz.conf',
+                                                            lib_module=mock.ANY,
+                                                            quiet=False)
```
eeshangarg: This isn't super important but we should try our best to keep code under 80 characters wherever possible. If doing so isn't super complicated, it makes the code more readable (in my opinion).
```
mock_run_message_handler_for_bot.assert_called_with(
    bot_name=None, # TODO: This should be equal to the bot's ...
    config_file='/foo/bar/baz.conf',
```
and so on.. :)

zulip_bots/zulip_bots/test_run_py.py
derAnfaenger: Travis seems to be running the tests (see [here](https://travis-ci.org/zulip/python-zulip-api/jobs/272814581#L555), lines 542-543), as these tests are already discovered by `tools/test-bots`. The script running all test suites is `test-main` (as opposed to `test-all`, since it doesn't run the linter). We made those changes as part of a refactoring of our test suites. 

Regarding the verbosity on single executions of this test suite, I oriented myself by  similar tests in this repo, e.g. `test_server.py`. We might want to make another PR though, that'll update all test files to be verbose.

derAnfaenger:
@eeshangarg updated with most of the requested changes :) My original reasoning for naming the file `test_run_py.py` was that `test_run` is a rather ambiguous name, so I was curious for feedback on my naming. I see though why `test_run_py` might be even more confusing, so renamed it.

timabbott:
I merged the first commit; the second seems to not actually run:
```
(zulip-api-py2-venv) tabbott@zaset:~/python-zulip-api$ python ./zulip_bots/zulip_bots/test_run.py
.usage: 
        zulip-run-bot <bot_name>
        Example: zulip-run-bot followup
        (This program loads bot-related code from the
        library code and then runs a message loop,
        feeding messages to the library code to handle.)
        Please make sure you have a current ~/.zuliprc
        file with the credentials you want to use for
        this bot.
        See lib/readme.md for more context.
        
zulip-run-bot: error: unrecognized arguments: --path-to-bot
E
======================================================================
ERROR: test_argument_parsing_with_bot_path (__main__.TestDefaultArguments)
  File "/home/tabbott/python-zulip-api/zulip-api-py2-venv/local/lib/python2.7/site-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "./zulip_bots/zulip_bots/test_run.py", line 38, in test_argument_parsing_with_bot_path
    zulip_bots.run.main()
  File "/home/tabbott/python-zulip-api/zulip_bots/zulip_bots/run.py", line 82, in main
    args = parse_args()
  File "/home/tabbott/python-zulip-api/zulip_bots/zulip_bots/run.py", line 76, in parse_args
    args = parser.parse_args()
  File "/usr/lib/python2.7/argparse.py", line 1704, in parse_args
    self.error(msg % ' '.join(argv))
  File "/usr/lib/python2.7/argparse.py", line 2374, in error
    self.exit(2, _('%s: error: %s\n') % (self.prog, message))
  File "/usr/lib/python2.7/argparse.py", line 2362, in exit
    _sys.exit(status)
SystemExit: 2

Ran 2 tests in 0.006s

FAILED (errors=1)

```

(also fails when run via `tools/test-bots`)

derAnfaenger:
@timabbott I see, that's because of the commit we just merged replacing the `--path-to-bot` option. Fixed that in my local branch, but somehow forgot to push it, sorry for that. Should be fixed now.

timabbott:
Merged, thanks @derAnfaenger!  What else should we be testing here?

PR #111, opened 2017-09-08
tools: Add reset-to-pull-request script.

derAnfaenger:
I've had this script from the main repo in the `python_zulip_api` repo for quite a while now and it proved useful, so I thought we should check it in.

### DIFF
--- a/tools/clean-branches
+++ b/tools/clean-branches
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+set -e
+
+# usage: clean-branches
+# Deletes any local branches which are ancestors of origin/master,
+# and also any branches in origin which are ancestors of
+# origin/master and are named like $USER-*.
+
+# usage: clean-branches --reviews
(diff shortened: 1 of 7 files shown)


timabbott:
It seems reasonable to bring the whole suite of those tools.

derAnfaenger:
Copied over all other git utility scripts I found that had no dependencies.

I excluded `pre-commit`, `commit-msg` and `setup-git-repo`, since those were dependent on vagrant. A slight altercation of e.g. `pre-commit` could make it handy to pre-run our linter, though.

derAnfaenger:
Update: Travis is fixed now. Updated the shebang in `tools/review` from `#!/usr/bin/env python3` to `#!/usr/bin/env python` . I guess this should be fine, since it's only a utility script (on top of that, the `python3` shebang fails on Windows).

timabbott:
Merged, thanks @derAnfaenger!

PR #116, opened 2017-09-14
Add mypy.

derAnfaenger:
This adds rudimental mypy support - the script, the requirements, and the travis setup.

* Right now, `mypy` would by default only get executed for some two files in `zulip`, which I had to update in order to not throw errors.

* I removed the `line coverage` option in `run-mypy` for now, so it's not included in travis either; I didn't really understand how codecov would handle the additional coverage file?

* There are a couple issues when we try to include `zulip_bots` or `zulip_botserver` in addition to `zulip`: Mypy detects e.g. multiple modules named `setup.py` . To resolve this, I suggest running a mypy instance for each of these packages.

* With this PR, we have 6 test suites. I think we should merge the test suites for linting and mypy.


derAnfaenger:
Closing, since the commit order is wrong, and for some reason my updated commit order is not reflected in origin.

PR #117, opened 2017-09-14
[WIP]Add mypy. 

derAnfaenger:
*This comes from #116. For some reason, I wasn't able to reorder the commits in origin, even though it worked locally.*

This adds rudimental mypy support - the script, the requirements, and the travis setup.

* Right now, `mypy` would by default only get executed for some two files in `zulip`, which I had to update in order to not throw errors.

* I removed the `line coverage` option in `run-mypy` for now, so it's not included in travis either; I didn't really understand how codecov would handle the additional coverage file?

* There are a couple issues when we try to include `zulip_bots` or `zulip_botserver` in addition to `zulip`: Mypy detects e.g. multiple modules named `setup.py` . To resolve this, I suggest running a mypy instance for each of these packages.

* With this PR, we have 6 test suites. I think we should merge the test suites for linting and mypy.


derAnfaenger:
Hm, so apparently I could have fixed the order issue in the previous PR. Github probably order the commits by the time stamps (why would the order be different from my local branch, though?) Resetting the last commit and recommiting solved the issue.

timabbott:
Yep, that incorrect display ordering thing is a somewhat famous bug in GitHub.

timabbott:
running mypy for each package seems reasonable.  I would pin a fixed version of `mypy`; they sometimes make breaking changes in new versions.

You've still got a few mypy errors; let me know if you need help debugging.

timabbott:
One more thought; we want `mypy` to be a dev-only dependency; I guess is that implied by having it accessed by `provision` but not `setup.py`?

derAnfaenger:
Yup, that was my thought.

derAnfaenger:
Regarding the mypy errors - I wonder why I would get `zulip/zulip/examples/unsubscribe:47: error: "ArgumentParser" has no attribute "usage"`. Since that attribute clearly exists, I tried it on both Python 3.6 and Python 3.5 consoles.

derAnfaenger:
The other errors seem reasonable, which makes me wonder why they don't show up on my system.

derAnfaenger:
Aah, should've rebase my local branch; `examples` moved.

timabbott:
`"ArgumentParser" has no attribute "usage"`: Looks like a bug in the typeshed stubs for ArgumentParser then.  We should open an issue in typeshed for this or just fix it; CC @gnprice to take care of that.

That said, I think we should change our code anyway to just use `parser.print_usage()` anyway, since that's the normal way to write this, right?

derAnfaenger:
Right, and we might also want to use `description=` instead of `usage=` in the future, since the latter overwrites the standard generated usage docopt string.

derAnfaenger:
Okay, updated and rewrote history with the nice command `git commit --amend --no-edit --reset-author`

timabbott:
I get these errors running it locally:
```
zulip/zulip/__init__.py:51: error: "module" has no attribute "__version__"
zulip/setup.py:72: error: "module" has no attribute "__version__"
zulip/zulip/send.py:100: error: "Namespace" has no attribute "message"
zulip/zulip/examples/upload-file:54: error: Incompatible types in assignment (expression has type "StringIO", variable has type IO[Any])
zulip/zulip/examples/upload-file:55: error: Property "name" defined in "IO" is read-only
```

Do you not see those?

derAnfaenger:
Hm, no, for me it passes just fine, and for Travis as well, so it might have something to do with the mypy version? Do you run it in the virtualenv?

timabbott:
Oh, I'm running from a python 2 virtualenv:
```
(zulip-api-py2-venv) tabbott@zaset:~/python-zulip-api$ which mypy
/usr/local/bin/mypy
```

timabbott:
Merged, thanks @derAnfaenger!  I'm not sure I see a super clean solution to the Python 2 venv problem.

gnprice:
:balloon: Thanks @derAnfaenger !

PR #123, opened 2017-09-18
[wip!] Add test-backend from Zulip server.

rht:
test pr


rht:
works!

timabbott:
@rht what's the thinking behind running the Zulip server's backend tests here?

rht:
This is to automate the integration test to check for PRs which require such, such as #18. The caveat is that it will slow down the aggregate CI test time. The other option is to add this commit only for the PRs where this test is needed, so the test is run once only.

The integration test will be required once again to test the stage 2 of Slack data import.

timabbott:
@derAnfaenger @eeshangarg what are your thoughts on this?

timabbott:
I think this might actually be a fine way to test for changes here that impact the integration with the backend...

derAnfaenger:
This is really cool, but it brings up the effective testing time from 1 minute to 11 minutes and goes over the magic five test jobs. Ultimately, I think we should merge this, but if it isn't urgent, maybe with a tweak similar to https://github.com/facebook/react/pull/2000 ? Here, Travis checks the diff of the changes before running a suite - this would allow us to run the backend tests iff e.g. `zulip/zulip` is modified.

rht:
@derAnfaenger agreed on the test duration and the limitation of the parallel tests ! Should the test be triggered in main only when https://github.com/zulip/python-zulip-api/tree/master/zulip/zulip is patched?

I really need this to test the slack data import PR (this (#908) has been in the roadmap for ages who knows when).

derAnfaenger:
Testing when `zulip/zulip` is patched would be a god start, yeah. We could then later refine the exact directories which are affected. For slack, I guess you can just follow your own suggestion for now and add this commit to the PR for test purposes?

derAnfaenger:
@rht would it be sufficient in this case to just run `tools/test-api`?

rht:
Ahh, right, right, let me check how long the test takes with just `tools/test-api`.

derAnfaenger:
@rht what's the state of this? We're possibly doing some iterations on the api code in the next days, so it'd be good to have the api tests set up by then. If you're working on other things right now, feel free to tell and I'd be happy to finish this :)

rht:
The only blocker is that I was testing between various Git clone ssh paths:
1. `sed -i "s|python-zulip-api.git@0.3.4|python-zulip-api.git@pull/$TRAVIS_PULL_REQUEST|" requirements/dev_lock.txt`
2. `sed -i "s|zulip/python-zulip-api.git@0.3.4|$TRAVIS_PULL_REQUEST_SLUG.git@$TRAVIS_PULL_REQUEST_BRANCH|" requirements/dev_lock.txt`
3. `sed -i "s|zulip/python-zulip-api.git@0.3.4|$TRAVIS_PULL_REQUEST_SLUG.git/$TRAVIS_PULL_REQUEST_BRANCH|" requirements/dev_lock.txt`

rht:
Most likely wrong URL formatting. If you have any idea on what's going on, and unblock this, I'm totally up for this.

rht:
I will have to update this to use the circleci script.

timabbott:
We can also mitigate the performance impact here by having the tool only run the subset of backend tests that's related to the subsystem.

timabbott:
@PIG208 I think this would be a good effort to pick up.  We probably won't end up reusing much of the exact code here, because we've moved to GitHub Actions since then, but I do think it'd be really nice if we can run the relevant part of the server's test suites against new python-zulip-api commits in CI.  The main things I'm imagining are:

* `tools/test-api` -- this is the most valuable, since it involves actually running the examples in our API documentation.
* `tools/test-backend x y z` where we pass just the modules like BotStorage that integrate with the API project.

but once we have this setup, we can add other things as well.

PR #126, opened 2017-09-26
Enhance mypy support.

derAnfaenger:
This
* runs mypy and lint in a single travis job (so we have five jobs again, the maximum that can be parallely executed)

* adds support for adding directories to the `lister` exclude list (needed for excluding zulip_bots/zulip_bots/bots

* runs mypy for each package fully and separately

* excludes all files that need to be fixed.

Btw I noticed that we have a `lister` duplicate; one in tools and one in `tools/server_lib`. Which one should we keep? I vote for the `server_lib` one and move `pep8` and `custom_check.py` there.


tools/lister.py
```diff
@@ -66,7 +66,13 @@ def list_files(targets=[], ftypes=[], use_shebang=True, modified_only=False,
...
     # everything is UTF-8.
     repository_root = subprocess.check_output(['git', 'rev-parse', '--show-toplevel']).strip().decode('utf-8')
-    exclude_abspaths = [os.path.normpath(os.path.join(repository_root, fpath)) for fpath in exclude]
+    exclude_abspaths = []
+    for path in exclude:
+        normed_path = os.path.normpath(os.path.join(repository_root, path))
+        if os.path.isdir(normed_path):
+            exclude_abspaths += [os.path.normpath(os.path.join(repository_root, fpath)) for fpath in list_files(targets=[normed_path])]
+        else:
+            exclude_abspaths.append(normed_path)
```
timabbott: Are you sure this didn't already work?  This should work:
```
        if any(absfpath == expath or absfpath.startswith(expath + '/')                               
               for expath in exclude_abspaths):                                                      
            continue                                                                                 
```

.travis.yml
```diff
@@ -8,9 +8,7 @@ env: TEST_SUITE=test-main
 matrix:
   include:
   - python: "3.6"
-    env: TEST_SUITE=lint
-  - python: "3.6"
-    env: TEST_SUITE=run-mypy
+    env: TEST_SUITE=test-static
```
timabbott: maybe `TEST_SUITE=static-analysis`?  I feel like we probably want this to be clearer than "test-static" as to what it is.

tools/run-mypy
```diff
@@ -75,8 +104,11 @@ if args.quick:
...
-else:
-    print("There are no files to run mypy on.")
+for repo, python_files in repo_python_files.items():
+    print("Running mypy for `{}`.".format(repo))
+    if python_files:
+        rc = subprocess.call([mypy_command] + extra_args + python_files)
+        if rc:
+            sys.exit(rc)
+    else:
+        print("There are no files to run mypy on.")
```
timabbott: I think we want it to check all 3 packages even if there are errors in the first one; so we should stored a `failed` variable instead of existing immediately on the first error.

timabbott:
I posted a few comments; thanks for working on this @derAnfaenger!

tools/run-mypy
derAnfaenger: Okay, I was not sure about that bit.

tools/lister.py
derAnfaenger: From my tests and the comment in the code, `excludes_paths` didn't exclude directories - we might want to make the same change to `zulip`?

timabbott: I think the comments are inaccurate.  Looking at the history, we used to have `contrib_bots/bots` in the exclude paths in the server repo.

derAnfaenger: Oh right - I see the issue then, it's Windows paths: 
```
if any(absfpath == expath or absfpath.startswith(expath + '/')
               for expath in exclude_abspaths):
            continue
```
hardcodes the delimiter, so it didn't work for the Windows files. Will update that snippet instead.

timabbott: Cool, yeah, and that fix we should contribute to the main repo too.

derAnfaenger:
Updated. Includes a new commit to remove the duplicate lister.py

timabbott:
This is good enough; I made one change to the last commit to remove the `|=` behavior (it's too hard to read and means the wrong thing, since it's a bitwise or!), and merged.

I'm not totally convinced by the use of `pathlib.PurePath`; why do we need that?  It feels like use of overly complex Python features for something simple :).  But we can address that in a follow-up commit.

derAnfaenger:
I used `pathlib.PurePath` since it makes extracting the first segment of a path very easy. `pathlib.Path` has the same functionality but is more heavy. `os.path` doesn't offer any comparable method. The easiest non-path alternative I can think of is extracting the beginning of the string until the first separator

timabbott:
OK, I guess that's fine.

neiljp:
`pathlib` is 3.4+; are tools generally 3+ only?

PR #129, opened 2017-09-28
bots: Improve and refactor docs to use Markdown macro framework.

eeshangarg:
@timabbott, @showell, @derAnfaenger: This is a starting point for how we can improve and render docs in the main repo and the disadvantages associated with doing so. I think I'd like to start a longer discussion about this tomorrow at the integrations meeting, so this is largely a WIP. Once we reach consensus on how to move forward, I can accelerate this project! :) Thanks!


timabbott:
@eeshangarg did you have more questions about this, or should I merge it?  It looks reasonable.

eeshangarg:
@timabbott: These commits are good to be merged! I was thinking I would push more commits here to the same PR as I work on other bots, but I think I'll make individual PRs for individual bots just to avoid confusion as to when something is a WIP and when it isn't. Thanks! :)

timabbott:
Merged, thanks @eeshangarg!  I think we may want to pause before writing docs for more of these, to figure out whether there are pieces we can automate (e.g. the "--help") piece from @neiljp's PRs.

PR #130, opened 2017-09-30
bots: Mark --message/-m argument to zulip_bot_output.py as required.

neiljp:
A minor fix; `zulip_bot_output.py` does not check for presence of `--message`, so it should be marked as required to avoid a traceback.


derAnfaenger:
Merged, after moving the `require=True` to a new line. Thanks @neiljp !

neiljp:
We might consider this being a positional argument?

derAnfaenger:
Yea, I thought about that as well.

timabbott:
Yeah, I think moving this to positional would be a good idea.

PR #131, opened 2017-10-02
bots: Further improvements to zulip_bot_output.py, beyond PR #130

neiljp:
* Move -m/--message to positional (as per discussion at end of #130)
* Add note regarding quoting messages being optional but necessary with spaces or empty
* Move to built-in argparse formatting (is this considered undesirable?)
* Adjusted examples to show use with/without quotes, and named vs path/to/bot
* Check for failed bot module import (ie. incorrect bot path usage)
* Check for provided module is a bot (ie. contains handler_class)

Re the last point, it suggests the idea of a 'bot_lint', which tests whether a module provides the bot API?


timabbott:
Nice, merged, thanks for the cleanup @neiljp!!

PR #132, opened 2017-10-06
Clarify and fix long-polling code.

derAnfaenger:
The unused long-polling parameter originates from https://github.com/zulip/python-zulip-api/commit/14ee40bf52bf6d21ba036756bcc4975041b372b5.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -491,10 +491,13 @@ class Client(object):
                 if files:
                     kwargs['files'] = req_files
 
+                # Make long-polling api request. When long-polling, set timeout to 90
+                # sec as a balance between a low traffic rate and a still reasonable
+                # latency time in case of a connection failure.
                 res = self.session.request(
                     method,
                     urllib.parse.urljoin(self.base_url, url),
-                    timeout=90,
+                    timeout=90 if longpolling else 30,
                     **kwargs)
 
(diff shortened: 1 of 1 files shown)


zulip/zulip/__init__.py
```diff
@@ -512,6 +512,8 @@ def end_error_retry(succeeded):
                 if longpolling:
                     # When longpolling, we expect the timeout to fire,
                     # and the correct response is to just retry
+                    # after a safety waiting time.
+                    time.sleep(1)
```
timabbott: We don't need to wait here, because fundamentally, we just had a timeout, and timeouts are slow and unlikely to be a DoS risk.

zulip/zulip/__init__.py
```diff
@@ -491,6 +491,9 @@ def end_error_retry(succeeded):
                 if files:
                     kwargs['files'] = req_files
 
+                # Make long-polling api request. Set timeout to 90 sec as a balance
+                # between a low traffic rate and a reasonable latency time in case of
+                # a connection failure.
```
timabbott: We probably actually want the timeout to be like 30 if longpolling is false, and only 90 if longpolling is true.

derAnfaenger:
Updated.

timabbott:
Merged after a few small tweaks to the comments, thanks @derAnfaenger!

PR #133, opened 2017-10-09
bots/giphy: Document the bot.

eeshangarg:
@timabbott: I am not sure how I feel about the way we get config file info in this bot. The way this is implemented [here](https://github.com/zulip/python-zulip-api/blob/master/zulip_bots/zulip_bots/lib.py#L101) seems a bit inconvenient. Should we expect users to go into the package directory and modify a config file for a particular bot there?


timabbott:
Yeah, this is definitely a recommendation that only makes sense if you're running it out of a Git clone (which was sorta the default model).  I think we do want to use the `config_handler` interface, but we should probably modify that interface in the `run.py` system to just require passing a `--config=` option to give the path if the bot is one that requires config... 

While you're working on that, it's probably worth testing that `zulip_botserver` does that reasonably as well, but it's a semi-separate project and can be its own PR.

showell:
Merged, thanks @eeshangarg!

derAnfaenger:
The macro `+{!running-a-bot.md!}` is not supported yet, so we should definitey work on adding macro support quickly now.

PR #134, opened 2017-10-12
virtual_fs bot: Use EmbeddedBot state handler.

derAnfaenger:
This enables `virtual_fs` to store the state in the bot state model when run as an embedded bot.

I use `pickle` to serialize each `fs` object, to store it as a string in the model. I'm aware of the potential security hazards coming with pickle, but I assume its usage here to be fine, since pickle does only deserialize what it once has serialized itself. Every object pickle serializes is a dictionary at its most outer level, so no code injection should be possible.

This will be required for https://github.com/zulip/zulip/pull/6927, where `virtual_fs` is added as an embedded bot.

### DIFF
--- a/zulip_bots/zulip_bots/bots/virtual_fs/test_virtual_fs.py
+++ b/zulip_bots/zulip_bots/bots/virtual_fs/test_virtual_fs.py
@@ -8,42 +8,41 @@ from zulip_bots.lib import StateHandler
 
 class TestVirtualFsBot(BotTestCase):
     bot_name = "virtual_fs"
+    help_txt = ('foo_sender@zulip.com:\n\nThis bot implements a virtual file system for a stream.\n'
+                'The locations of text are persisted for the lifetime of the bot\n'
+                'running, and if you rename a stream, you will lose the info.\n'
+                'Example commands:\n\n```\n'
+                '@mention-bot sample_conversation: sample conversation with the bot\n'
+                '@mention-bot mkdir: create a directory\n'
+                '@mention-bot ls: list a directory\n'
+                '@mention-bot cd: change directory\n'
+                '@mention-bot pwd: show current path\n'
+                '@mention-bot write: write text\n'
+                '@mention-bot read: read text\n'
+                '@mention-bot rm: remove a file\n'
+                '@mention-bot rmdir: remove a directory\n'
+                '```\n'
+                'Use commands like `@mention-bot help write` for more details on specific\ncommands.\n')
 
-    def test_bot(self):
-        help_txt = ('foo_sender@zulip.com:\n\nThis bot implements a virtual file system for a stream.\n'
-                    'The locations of text are persisted for the lifetime of the bot\n'
-                    'running, and if you rename a stream, you will lose the info.\n'
-                    'Example commands:\n\n```\n'
-                    '@mention-bot sample_conversation: sample conversation with the bot\n'
-                    '@mention-bot mkdir: create a directory\n'
-                    '@mention-bot ls: list a directory\n'
-                    '@mention-bot cd: change directory\n'
-                    '@mention-bot pwd: show current path\n'
-                    '@mention-bot write: write text\n'
-                    '@mention-bot read: read text\n'
-                    '@mention-bot rm: remove a file\n'
-                    '@mention-bot rmdir: remove a directory\n'
-                    '```\n'
-                    'Use commands like `@mention-bot help write` for more details on specific\ncommands.\n')
-        expected = {
-            "cd /home": "foo_sender@zulip.com:\nERROR: invalid path",
-            "mkdir home": "foo_sender@zulip.com:\ndirectory created",
-            "pwd": "foo_sender@zulip.com:\n/",
-            "help": help_txt,
-            "help ls": "foo_sender@zulip.com:\nsyntax: ls <optional_path>",
-            "": help_txt,
-        }
+    def test_commands_1(self):
+        expected = [
+            ("cd /home", "foo_sender@zulip.com:\nERROR: invalid path"),
+            ("mkdir home", "foo_sender@zulip.com:\ndirectory created"),
+            ("pwd", "foo_sender@zulip.com:\n/"),
+            ("help", self.help_txt),
+            ("help ls", "foo_sender@zulip.com:\nsyntax: ls <optional_path>"),
+            ("", self.help_txt),
+        ]
         self.check_expected_responses(expected)
 
+    def test_commands_2(self):
         expected = [
-            ("help", help_txt),
+            ("help", self.help_txt),
             ("help ls", "foo_sender@zulip.com:\nsyntax: ls <optional_path>"),
-            ("", help_txt),
+            ("", self.help_txt),
             ("pwd", "foo_sender@zulip.com:\n/"),
             ("cd /home", "foo_sender@zulip.com:\nERROR: invalid path"),
             ("mkdir home", "foo_sender@zulip.com:\ndirectory created"),
             ("cd /home", "foo_sender@zulip.com:\nCurrent path: /home/"),
         ]
-
-        storage = StateHandler()
-        self.check_expected_responses(expected, storage = storage)
+        self.check_expected_responses(expected)
--- a/zulip_bots/zulip_bots/bots/virtual_fs/test_virtual_fs.py
+++ b/zulip_bots/zulip_bots/bots/virtual_fs/test_virtual_fs.py
@@ -45,5 +45,5 @@ class TestVirtualFsBot(BotTestCase):
             ("cd /home", "foo_sender@zulip.com:\nCurrent path: /home/"),
         ]
 
-        state_handler = StateHandler()
-        self.check_expected_responses(expected, state_handler = state_handler)
+        storage = StateHandler()
+        self.check_expected_responses(expected, storage = storage)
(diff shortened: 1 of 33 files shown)


timabbott:
Why not just use JSON rather than pickle for storage?

derAnfaenger:
I tried JSON first, but the native serialization doesn't work here, because the `fs` object contains sets, which the normal parser fails to emulate (only works for dicts and lists). There are some workarounds where you specify a custom evaluation for sets, but those are rather ugly. What do you suggest?

timabbott:
@showell I'd be curious for your thoughts on the protocol question here.  My intuition is that we should avoid pickles just because this bots' main purpose is to be an example, and pickles probably aren't the right approach.  So my leaning would be that we should just change the tooling to turn sets into lists or something, but maybe I'm thinking about this wrong.

showell:
We should use JSON.  It's annoying that json encoders don't just map sets to lists, but we deal with this all the time in server code.  I think we just want clear errors when you try to encode stuff that is not int/string/list/dict, and our demo code should illustrate working with sets by just doing list -> set after getState and set -> list after setState.

derAnfaenger:
Updated to use JSON.

zulip_bots/zulip_bots/bots/virtual_fs/virtual_fs.py
```diff
@@ -18,16 +30,29 @@ def handle_message(self, message, bot_handler, state_handler):
...
+            with state_handler.state({}) as state:
+                if recipient not in state:
+                    state[recipient] = fs_new()
+                fs = state[recipient]
+                if sender not in fs['user_paths']:
+                    fs['user_paths'][sender] = '/'
+                fs, msg = fs_command(fs, sender, command)
+                prependix = '{}:\n'.format(sender)
+                msg = prependix + msg
+                state[recipient] = fs
```
timabbott: This feels like it has a lot of duplicated code flow.  I think what we want is to modify the `state_handler.state` function to accept as an argument the marshalling function for each direction and just have the library take care of the cases; then we can still use the nice `with` logic.

showell: It does feel funny to me that virtual_fs has to care whether it's in an embedded bot or not.  I would like there's always a StateHandler, and the implementations are different on the backend.

derAnfaenger: I'll shortly push an update which reduces the boilerplate here to choosing between `bot_handler.StateHandler` and `state_handler.StateHandler`.  I see this remaining check as unavoidable as long as we don't upgrade the external bot's state handler to use the server state handler.

showell: I've been out of the loop, because I didn't realize we had two state handlers.  I was under the mindset that you'd only ever have one API to write state, and then the deployment option would govern how state got saved on the backend, plus whether state persisted if you restart the bot.

derAnfaenger: We could've gone the way of contuing to use the old statehandler and cranking it so it fits the new format, but I felt that the conceptual differences were too big: The old StateHandler was an object that could be assigned any value, so of type Any, the new StateHandler behaves like a Dict[str, str]. Furthermore, the old statehandler got passed to the bot in handle_message, whereas the new one is an attribute of the bot_handler. I should've probably discussed that change with someone, it felt cleaner to me. Because of these differences, I thought it'd make sense to have a clear cut between both state handlers. eventually, the old one would be rewritten to fit the new model.

showell: I think we might as well eliminate the old one now, since basically nobody uses the old one now, except for the file that you're modifying.

derAnfaenger: As in removing it and making incrementor and virtual_fs embedded_bot-only bots?

showell: No, I'm suggesting that we just have them use a StateHandler that works locally.  I'm still a little confused as to why this is difficult.  The whole reason the original code passed in a state_handler to `handle_message` was to give us the flexibility to have the **runtime** control the implementation, not the virtual_fs bot.  It's a slightly leaky abstraction in cases where you don't have permanent file storage, but the original bot had reasonable persistence.

Most bots deal with so little data--even the tricked out virtual_fs bot--that it's totally reasonable to just save a blob after every message.

And my thought was we were simply making the storage mechanisms more permanent here.

Also, there's no reason why a runtime for a non-embedded bot can't write to the db.  We could totally create a save-user-data endpoint.  That was what I thought we were going to do here, but I obviously either didn't communicate it well, or maybe I just don't understand the tradeoffs well enough.

zulip_bots/zulip_bots/bots/virtual_fs/virtual_fs.py
```diff
@@ -2,6 +2,20 @@
...
+import json
+
+from contextlib import ExitStack
+
+class SetEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, set):
+            return dict(_set_object=list(obj))
+        return super().default(self, obj)
+
```
showell: I talked to @timabbott, and I think we decided that `SetEncoder` is a bit too heavy of a solution here.  99% of bots can just use lists and dictionaries, and even this bot doesn't need the O(1)-ness of sets, since it's just a toy.  (Or even if it did, that could be done more explicitly after your read in the data.)

So I think we might need to abandon this commit.

derAnfaenger:
Marking this WIP until we have a clearer direction on how to proceed.

derAnfaenger:
Pushed some commits that do every discuessed StateHandler change except the conversion to a dict. 
That conversion is nearly finished, will push it tmw.
`zulip_bots: Reify StateHandler testing.` resolves various hazzles I encountered when updating the tests to work with a dict-like StateHandler.

zulip_bots/zulip_bots/bots/virtual_fs/test_virtual_fs.py
```diff
@@ -8,42 +8,64 @@
...
-            "help ls": "foo_sender@zulip.com:\nsyntax: ls <optional_path>",
-            "": help_txt,
-        }
-        self.check_expected_responses(expected)
+    def test_commands_stream_1(self):
+        expected = [
+            ("cd /home", "foo_sender@zulip.com:\nERROR: invalid path"),
+            ("mkdir home", "foo_sender@zulip.com:\ndirectory created"),
+            ("pwd", "foo_sender@zulip.com:\n/"),
+            ("help", TestVirtualFsBot.help_txt),
```
timabbott: this can be `self.help_txt`.

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -82,31 +83,19 @@ def check_expected_responses(self, expectations, expected_method='send_reply',
             trigger_messages.append(stream_message_dict)
 
         for trigger_message in trigger_messages:
-            if storage is None:
-                current_storage = None
-            else:
-                # A new (copy of) the StateHandler is used for each trigger message.
-                # This avoids type="all" failing if state is created in the first iteration.
-                current_storage = deepcopy(storage)
-
```
timabbott: I don't like re-introducing this landmine.  I think instead we should just change the mechanism for `all` to run through the whole block below twice (clearing the storage in between), one for private messages, the other for stream, rather than having it potentially alternate.

timabbott:
Generally looks reasonable; I posted a few comments.  I think we should try to push this through quickly (and update the docs on the server) before too many more people write bots, since this does break the ABI.

zulip_bots/zulip_bots/test_lib.py
derAnfaenger: Okay. I think we should strongly encourage test writers to use this function only once per test case when testing stateful bots: Even if we clear the state between runs through private and stream messages, the exit state is not clear to the user: It could be either the state after sending the stream messages, or the state after the private messages, if he doesn't explicitly look into the function's interior.

derAnfaenger:
@timabbott updated with your comments and added the last couple commits. The last one accomplishes the main load with matching up the StateHandler interfaces. I dropped the context manager for now, since it wasn't compatible and didn't have much use for any except the incrementor bot. I see how the idea of initializing the StateHandler with some initial state can be useful, but maybe we should rediscuss the best implementation for both StateHandlers.

timabbott:
OK, this is definitely progress, so merged, thanks @derAnfaenger!  I think we'll want a few followups:
* Putting a highlight in the changelog about the ABI change.  @eeshangarg is there a good place to make a note of this to ensure the right happens in our next release?  We should have our next release be version 0.4.0 to reflect the interface change.
* We should think a bit more about how we expect bots to interact with state by default.  E.g. I imagine for a tic-tac-toe bot, we'll want its state to be keyed by the thread (stream/topic pair or PM thread) involved, not having just a single game state.  
* One thing that's likely important for performance is that probably a bot should do at most 2 queries to interact with the state per message it receives, to avoid excessive round trips to a server.  So e.g. it should do a single query fetch the `dict` that is the state for a given key (e.g. for our tic-tac-toe bot, key might be the thread) and then when we're done, save it back to the server at the end; not do a query for every small state change.  We should make sure our documentation and examples reflect this.

@derAnfaenger can you open issues for these?

PR #135, opened 2017-10-20
mypy: Fix erroring files.

### LINKED ISSUE #127: mypy: Fix erroring files.
## Background

We have recently added mypy support to this repo, but many files are corrupted and throw errors on mypy.

## Task
#126 updates the `exclude_list` in `tools/run-mypy` Each file in this list errors on mypy. We should fix each file and remove it from the list. **`zulip_bots/zulip_bots/bots` is an exception**, we don't enforce mypy annotations in this directory.

*This task is great for beginners or/and those who want to get comfortable with mypy annotations.*

alenavolk:
Finally, I have a PR ready :) Things to keep in mind while reviewing:

1. I annotated strings as `Text` in files that didn't have any annotations, and as `str` in files that already had `str` annotations.

2. I made a couple of minor changes to the code in places where I thought it made sense. I tested these changes, but it would be great to have someone look over.

3. I tried to minimize using `type: ignore`, but you'll still see it in a couple of places. Please let me know if you can suggest a way to avoid that.

Fixes #127.

### DIFF
--- a/zulip/integrations/irc/irc_mirror_backend.py
+++ b/zulip/integrations/irc/irc_mirror_backend.py
@@ -1,6 +1,7 @@
 import irc.bot
 import irc.strings
 from irc.client import ip_numstr_to_quad, ip_quad_to_numstr, Event, ServerConnection
+from typing import Any, Dict
 
 IRC_DOMAIN = "irc.example.com"
 
--- a/zulip/integrations/log2zulip/log2zulip
+++ b/zulip/integrations/log2zulip/log2zulip
@@ -97,7 +97,7 @@ def process_logs():
     open(data_file_path, "w").write(json.dumps(new_data))
 
 if __name__ == "__main__":
-    parser = zulip.add_default_arguments(argparse.ArgumentParser())  # type: argparse.ArgumentParser3
+    parser = zulip.add_default_arguments(argparse.ArgumentParser())  # type: argparse.ArgumentParser
     parser.add_argument("--control-path", default="/etc/log2zulip.conf")
     args = parser.parse_args()
 
@@ -115,7 +115,7 @@ if __name__ == "__main__":
         zulip_client = zulip.init_from_options(args)
         try:
             log_files = json.loads(open(args.control_path, "r").read())
-        except (json.JSONDecodeError, IOError):
+        except (json.JSONDecodeError, IOError):  # type: ignore # error: Cannot determine type of 'IOError'
             print("Could not load control data from %s" % (args.control_path,))
             traceback.print_exc()
             sys.exit(1)
--- a/zulip/integrations/zephyr/process_ccache
+++ b/zulip/integrations/zephyr/process_ccache
@@ -10,7 +10,7 @@ ccache_data_encoded = sys.argv[3]
 
 # Update the Kerberos ticket cache file
 program_name = "zmirror-%s" % (short_user,)
-with open("/home/zulip/ccache/%s" % (program_name,), "w") as f:
+with open("/home/zulip/ccache/%s" % (program_name,), "wb") as f:
     f.write(base64.b64decode(ccache_data_encoded))
 
 # Setup API key
--- a/zulip/tests/__init__.py
+++ b/zulip/tests/__init__.py
@@ -1,2 +1,4 @@
 import pkgutil
-__path__ = pkgutil.extend_path(__path__, __name__)
+from typing import Iterable, Text
+
+__path__ = pkgutil.extend_path(__path__, __name__)  # type: Iterable[Text]
--- a/zulip/tests/test_default_arguments.py
+++ b/zulip/tests/test_default_arguments.py
@@ -16,9 +16,11 @@ else:
 class TestDefaultArguments(TestCase):
 
     def test_invalid_arguments(self):
+        # type: () -> None
         parser = zulip.add_default_arguments(argparse.ArgumentParser(usage="lorem ipsum"))
-        with self.assertRaises(SystemExit) as cm, patch('sys.stderr', new=six.StringIO()) as mock_stderr:
-            parser.parse_args(['invalid argument'])
+        with self.assertRaises(SystemExit) as cm:  # type: ignore # error: "assertRaises" doesn't match argument types
+            with patch('sys.stderr', new=six.StringIO()) as mock_stderr:
+                parser.parse_args(['invalid argument'])
         self.assertEqual(cm.exception.code, 2)
         # Assert that invalid arguments exit with printing the full usage (non-standard behavior)
         self.assertTrue(mock_stderr.getvalue().startswith("""usage: lorem ipsum
@@ -32,6 +34,7 @@ Zulip API configuration:
 
     @patch('os.path.exists', return_value=False)
     def test_config_path_with_tilde(self, mock_os_path_exists):
+        # type: (bool) -> None
         parser = zulip.add_default_arguments(argparse.ArgumentParser(usage="lorem ipsum"))
         test_path = '~/zuliprc'
         args = parser.parse_args(['--config-file', test_path])
--- a/zulip_bots/generate_manifest.py
+++ b/zulip_bots/generate_manifest.py
@@ -5,6 +5,7 @@ import os
 import glob
 import distutils.cmd
 import distutils.log
+from typing import IO, Iterator
 
 CURRENT_DIR = os.path.abspath(os.path.dirname(__file__))
 BOTS_DIR = os.path.normpath(os.path.join(CURRENT_DIR, 'zulip_bots', 'bots'))
@@ -21,27 +22,30 @@ class GenerateManifest(distutils.cmd.Command):
     ]
 
     def initialize_options(self):
+        # type: () -> None
         self.release = False
 
     def finalize_options(self):
+        # type: () -> None
         pass
 
     def run(self):
+        # type: () -> None
         if self.release:
             generate_release_manifest()
-            self.announce(
+            self.announce(  # type: ignore # error: "GenerateManifest" has no attribute "announce"
                 'Generating a MANIFEST for a PyPA release of zulip_bots.',
-                level=distutils.log.INFO
+                level=distutils.log.INFO  # type: ignore # error: Module has no attribute "INFO"
             )
         else:
             generate_dev_manifest()
-            self.announce(
+            self.announce(  # type: ignore
                 'Generating a MANIFEST for zulip_bots\' development.',
-                level=distutils.log.INFO
+                level=distutils.log.INFO  # type: ignore
             )
 
 def get_test_fixtures():
-    # type: () -> List[str]
+    # type: () -> Iterator[str]
     glob_pattern = os.path.join(BOTS_DIR, '*', 'fixtures', '*.json')
     fixtures_paths = map(
         lambda fp: os.path.join(*fp.split(os.path.sep)[-5:]).replace(os.path.sep, '/'),
@@ -50,7 +54,7 @@ def get_test_fixtures():
     return fixtures_paths
 
 def get_logos():
-    # type: () -> List[str]
+    # type: () -> Iterator[str]
     glob_pattern = os.path.join(BOTS_DIR, '*', 'logo.*')
     logo_paths = map(
         lambda fp: os.path.join(*fp.split(os.path.sep)[-4:]).replace(os.path.sep, '/'),
@@ -59,7 +63,7 @@ def get_logos():
     return logo_paths
 
 def get_docs():
-    # type: () -> List[str]
+    # type: () -> Iterator[str]
     glob_pattern = os.path.join(BOTS_DIR, '*', 'doc.md')
     doc_paths = map(
         lambda fp: os.path.join(*fp.split(os.path.sep)[-4:]).replace(os.path.sep, '/'),
@@ -68,7 +72,7 @@ def get_docs():
     return doc_paths
 
 def get_assets():
-    # type: () -> List[str]
+    # type: () -> Iterator[str]
     glob_pattern = os.path.join(BOTS_DIR, '*', 'assets', '*')
     assets_files = map(
         lambda fp: os.path.join(*fp.split(os.path.sep)[-5:]).replace(os.path.sep, '/'),
@@ -77,7 +81,7 @@ def get_assets():
     return assets_files
 
 def generate_and_write(filepaths, file_obj):
-    # type: () -> None
+    # type: (Iterator[str], IO[str]) -> None
     template = 'include {line}\n'
     lines = map(lambda line: template.format(line=line), filepaths)
 
@@ -99,6 +103,7 @@ def generate_release_manifest():
         generate_and_write(get_assets(), fp)
 
 def parse_args():
+    # type: () -> argparse.Namespace
     usage = """
 To generate a MANIFEST.in for a PyPA release, run:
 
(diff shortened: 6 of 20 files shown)


alenavolk:
I just realized that the changes I've made in `zulip_bots/generate_manifest.py` are causing a problem with `/tools/provision`. I'll think how to fix that.

timabbott:
@alenavolk it looks like Travis is failing on Python 3.7 and 3.4 because we need to add the `typing` package to our `requirements.txt` files.

BTW you should come to our integrations/bots meeting tomorrow at 9AM http://zulip.readthedocs.io/en/latest/chat-zulip-org.html#chat-meetings

zulip_bots/generate_manifest.py
```diff
@@ -68,7 +72,7 @@ def get_docs():
     return doc_paths
 
 def get_assets():
-    # type: () -> List[str]
+    # type: () -> Iterator[str]
```
derAnfaenger: Since `map()` returns a list, can we do `from typing import List` and keep the List types? Also for the other occurrences in this file.

zulip_bots/zulip_bots/provision.py
```diff
@@ -72,7 +74,7 @@ def main():
     # type: () -> None
     current_dir = os.path.dirname(os.path.abspath(__file__))
     bots_dir = os.path.join(current_dir, "bots")
-    bots_subdirs = map(os.path.abspath, glob.glob(bots_dir + '/*'))
+    bots_subdirs = map(lambda d: os.path.abspath(d), glob.glob(bots_dir + '/*'))
```
derAnfaenger: `map` takes a function as the first argument, so the initial version should be fine here.

zulip_bots/zulip_bots/provision.py
```diff
@@ -35,6 +36,7 @@ def provision_bot(path_to_bot, force):
 
 
 def parse_args(available_bots):
+    # type: (Iterator[AnyStr]) -> argparse.Namespace
```
derAnfaenger: What's the reason for using AnyStr instead of str here?

zulip_botserver/tests/__init__.py
```diff
@@ -1,2 +1,2 @@
 import pkgutil
-__path__ = pkgutil.extend_path(__path__, __name__)
+__path__ = pkgutil.extend_path(__path__, __name__)  # type: ignore # error: Cannot determine type of '__path__'
```
derAnfaenger: Same as for `zulip_bots`

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -35,7 +35,7 @@ def get_bot_message_handler(self):
         # handler class. Eventually, we want bot's handler classes to
         # inherit from a common prototype specifying the handle_message
         # function.
-        lib_module = import_module('zulip_bots.bots.{bot}.{bot}'.format(bot=self.bot_name))
+        lib_module = import_module('zulip_bots.bots.{bot}.{bot}'.format(bot=self.bot_name))  # type: Any
```
neiljp: `types.ModuleType` again here?

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -61,7 +61,7 @@ def check_expected_responses(self, expectations, expected_method='send_reply',
         # To test send_message, Any would be a Dict type,
         # to test send_reply, Any would be a str type.
         if isinstance(expectations, dict):
-            expected = [(k, v) for k, v in expectations.items()]
+            expected = [(k, v) for k, v in expectations.items()]  # type: Sequence[Tuple[str, Any]]
```
neiljp: This is a list comprehension, in the implementation, so just type it as that?

zulip_bots/zulip_bots/zulip_bot_output.py
```diff
@@ -75,7 +77,7 @@ def main():
         sys.exit(1)
 
     with patch('zulip.Client') as mock_client:
-        mock_bot_handler = ExternalBotHandler(mock_client, bot_dir)
+        mock_bot_handler = ExternalBotHandler(mock_client, bot_dir)  # type: Any
```
neiljp: Is this necessary since this object gets 'monkey patched' later?

zulip_botserver/setup.py
```diff
@@ -51,14 +52,15 @@
 
     # Manual dependency check
     def check_dependency_manually(module_name, version=None):
+        # type: (str, Optional[str]) -> None
         try:
-            module = import_module(module_name)
+            module = import_module(module_name)  # type: Any
```
neiljp: `types.ModuleType` ?

alenavolk:
Regarding the failing Travis build. It looks like adding `from typing ...` imports to these files:
```
zulip_bots/setup.py
zulip_bots/generate_manifest.py
zulip_botserver/setup.py
```
causes the `tools/provision` script to fail with the `No module named typing` error.

Travis uses this script to recreate the environment every time it runs the tests, that's why the build fails even though all tests have passed locally.

I've tried adding the `typing` package to `requirements.txt`, but it doesn't really solve the problem, because these requirements:
```
-e ./zulip
-e ./zulip_bots
-e ./zulip_botserver
```
are installed from the same place, and `pip` looks at their `setup.py` files before the installation.
 (Besides, `typing` is installed with the `./zulip` package later.)

The only solution I can think of at the moment is to change the problematic imports as follows:
```
if False:
    from typing import Any, Dict, Optional
```
It fixes the problem locally, and I think it'll make the Travis build pass as well.

timabbott:
@alenavolk I think that means we need to add `typing` in the `setup.py` files.

alenavolk:
@timabbott I've added `typing` as a requirement to both files. For example, in `zulip_bots/setup.py`:
```
setuptools_info = dict(
    install_requires=[
        'zulip',
        'mock>=2.0.0',
        'html2text',  # for bots/define
        'typing>=3.5.2.2',
    ],
)
```
The same thing in `zulip_botserver/setup.py`. Is that what you had in mind?

Unfortunately, I'll still have to use the `if False` trick. Otherwise, the provision script fails (it fails before installing the requirements).

zulip/tests/__init__.py
```diff
@@ -1,2 +1,2 @@
 import pkgutil
-__path__ = pkgutil.extend_path(__path__, __name__)
+__path__ = pkgutil.extend_path(__path__, __name__)  # type: ignore # error: Cannot determine type of '__path__'
```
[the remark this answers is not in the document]
alenavolk: Fixed in both cases, thanks!

zulip_bots/setup.py
```diff
@@ -63,14 +64,15 @@
 
     # Manual dependency check
     def check_dependency_manually(module_name, version=None):
+        # type: (str, Optional[str]) -> None
         try:
-            module = import_module(module_name)
+            module = import_module(module_name)  # type: Any
```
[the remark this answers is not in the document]
alenavolk: It looks very logical here, but for some reason it gives me the `Module has no attribute "__version__"` error two lines below.

zulip_bots/zulip_bots/test_lib.py
alenavolk: This one gives me the `Module has no attribute "handler_class"` error.

zulip_botserver/setup.py
alenavolk: Same error as in `zulip_bots/setup.py`: `Module has no attribute "__version__"`.

zulip_bots/zulip_bots/run.py
```diff
@@ -9,13 +9,15 @@
...
 from os.path import basename, splitext
+from typing import Any, Optional, Text
 
 from zulip_bots.lib import run_message_handler_for_bot
 from zulip_bots.provision import provision_bot
 
 current_dir = os.path.dirname(os.path.abspath(__file__))
 
 def import_module_from_source(path, name=None):
+    # type: (Text, Optional[Text]) -> Any
```
alenavolk: @neiljp I think this is the only place where I can use `types.ModuleType` right now.

zulip_bots/zulip_bots/zulip_bot_output.py
alenavolk: I think so, otherwise I get several `Cannot assign to a method` errors right below (and many others later).

alenavolk:
@neiljp Thank you so much for reviewing this! I really appreciate it.

Today I tried annotating the module instances with `types.ModuleType` instead of `Any`, but I got the following errors:
```
zulip_bots/setup.py:73: error: Module has no attribute "__version__"
zulip_bots/zulip_bots/test_lib.py:39: error: Module has no attribute "handler_class"
zulip_botserver/setup.py:61: error: Module has no attribute "__version__"
```
Does it mean I should keep the `type: Any` annotations in these files?

Also, I know I could use `types.ModuleType` in annotating `zulip_bots/zulip_bots/run.py`:
```
def import_module_from_source(path, name=None):
    # type: (Text, Optional[Text]) -> ModuleType
    ...
```
but I'm not sure if it makes sense. Wouldn't it be better to keep `Any` for consistency?

Looking forward to your reply!

timabbott:
@alenavolk if you have a subset of these commits that are fully reviewed and pass tests, it could be good to open those in a new PR so that we can start shrinking this.

Though .. I guess we need to figure out the Travis issue with `typing` first.  @eeshangarg @derAnfaenger do either of you understand the problem?

timabbott:
Actually, looking at the error, I think it's just that `generate_manifest.py`, which is called from `setup.py`, is special, and does need the `if False` hack.  We should only need that in a couple files though.

alenavolk:
@timabbott Unfortunately, the error Travis is showing right now is not the only one, there are actually three files failing the Travis build. It's not a complicated problem or anything, but since we don't have a common understanding of how to fix it, I think the best approach would be to first shrink this PR as much as possible and then continue this discussion. I shouldn't have made a PR this big in the first place, but oh well... I'll work with what I've got now :)

timabbott:
@alenavolk which 3 files are failing?  I think we might want to just do the `if False` thing for them...

alenavolk:
@timabbott `zulip_bots/setup.py`, `zulip_bots/generate_manifest.py`, and `zulip_botserver/setup.py`. `zulip/setup.py` already has the `if False` hack, otherwise it would be on this list too.
 
The problem in a nutshell is that `typing` is included in the standard library only starting from Python 3.5. On the earlier versions we should either install it before installing our packages (`zulip`, `zulip_bots`, and `zulip_botserver`) the same way we install a modern version of `pip`:
```
subprocess.call([pip_path, 'install', 'pip>=9.0', 'typing>=3.5.2.2'])
if subprocess.call([pip_path, 'install', '--prefix', venv_dir, '-r',
                    os.path.join(base_dir, requirements_filename)]):
    raise OSError("The command `pip install -r {}` failed. Dependencies not installed!"
                  .format(os.path.join(base_dir, requirements_filename)))
```
or use the `if False` hack in our setup files.

We could also just leave these files on the `exclude` list for the time being. I honestly don't know which of these solutions is cleaner.

zulip_bots/zulip_bots/provision.py
alenavolk: I was just trying to get rid of the errors below. You're right, `str` works fine here. Fixed.

zulip_bots/zulip_bots/provision.py
alenavolk: It definitely should, but it results in a bunch of weird errors:
```
Argument 1 to "map" has incompatible type Callable[[AnyStr], AnyStr]; expected Callable[[str], AnyStr]
Argument 1 to "isdir" has incompatible type "AnyStr"; expected "Union[bytes, str, _PathLike[Any]]"
Argument 1 to "parse_args" has incompatible type Iterator[AnyStr]; expected Iterator[str]
```
@neiljp FYI :)

zulip/integrations/log2zulip/log2zulip
```diff
@@ -115,7 +115,7 @@ if __name__ == "__main__":
         zulip_client = zulip.init_from_options(args)
         try:
             log_files = json.loads(open(args.control_path, "r").read())
-        except (json.JSONDecodeError, IOError):
+        except (json.JSONDecodeError, IOError):  # type: ignore # error: Cannot determine type of 'IOError'
```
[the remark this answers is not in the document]
alenavolk: @neiljp Just tried that, it gives me the `misplaced type annotation` error. As a matter of fact, any annotation apart from the `type: ignore` does.

zulip/integrations/zephyr/process_ccache
```diff
@@ -10,7 +10,7 @@ ccache_data_encoded = sys.argv[3]
 
 # Update the Kerberos ticket cache file
 program_name = "zmirror-%s" % (short_user,)
-with open("/home/zulip/ccache/%s" % (program_name,), "w") as f:
+with open("/home/zulip/ccache/%s" % (program_name,), "wb") as f:
```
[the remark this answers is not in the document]
alenavolk: Yep, the error is `Argument 1 to "write" of "IO" has incompatible type "bytes"; expected "str"`.

zulip_bots/generate_manifest.py
alenavolk: That's how it was originally. I changed it to avoid this error: `Incompatible return value type (got Iterator[str], expected List[str])`.

@neiljp FYI :)

zulip_bots/zulip_bots/test_lib.py
alenavolk: This line was deleted in `master`.

zulip/integrations/irc/irc_mirror_backend.py
```diff
@@ -1,6 +1,7 @@
 import irc.bot
 import irc.strings
 from irc.client import ip_numstr_to_quad, ip_quad_to_numstr, Event, ServerConnection
+from typing import Any, Dict
```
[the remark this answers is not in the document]
alenavolk: That's a good point, but I don't think it's possible here. If I understand everything correctly, in this file `Any` is used to annotate the instances of the classes defined in the main Zulip repo.

zulip/tests/test_default_arguments.py
```diff
@@ -16,8 +16,11 @@
 class TestDefaultArguments(TestCase):
 
     def test_invalid_arguments(self):
+        # type: () -> None
         parser = zulip.add_default_arguments(argparse.ArgumentParser(usage="lorem ipsum"))
-        with self.assertRaises(SystemExit) as cm, patch('sys.stderr', new=six.StringIO()) as mock_stderr:
+        cm = self.assertRaises(SystemExit)  # type: ignore # error: No variant of "assertRaises" matches argument types
+        mock_stderr = patch('sys.stderr', new=six.StringIO())
+        with cm as cm, mock_stderr as mock_stderr:
```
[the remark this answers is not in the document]
alenavolk: @derAnfaenger Not really, it would make the test fail:
```
ERROR: test_invalid_arguments (tests.test_default_arguments.TestDefaultArguments)
  File "/root/python-zulip-api/zulip/tests/test_default_arguments.py", line 27, in test_invalid_arguments
    self.assertTrue(mock_stderr.getvalue().startswith("""usage: lorem ipsum
AttributeError: '_patch' object has no attribute 'getvalue'
```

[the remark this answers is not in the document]
alenavolk: @neiljp Could you explain again how this line could be improved if `typing` is fixed? The line you suggested looks exactly like the original now :)

timabbott:
Merged, thanks @alenavolk!  This cleaned up a lot of issues.  I think we can probably deal with any remaining `type: ignore` items in follow-up commits.

PR #143, opened 2017-10-30
bots: Access state in external Zulip server.

derAnfaenger:
This depends on [zulip/#7227](https://github.com/zulip/zulip/pull/7227)

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -859,6 +859,38 @@ class Client(object):
             request=request,
         )
 
+    def update_state(self, request):
+        # type: (Dict[str, Any]) -> Dict[str, Any]
+        '''
+            Example usage:
+
+            >>> client.update_state({'state': {"entry 1": "value 1", "entry 2": "value 2", "entry 3": "value 3"}})
+            >>> client.get_state({'keys': ["entry 1", "entry 3"]})
+            {'result': 'success', 'state': {'entry 1': 'value 1', 'entry 3': 'value 3'}, 'msg': ''}
+        '''
+        return self.call_endpoint(
+            url='user_state',
+            method='PUT',
+            request=request,
+        )
+
+    def get_state(self, request=None):
+        # type: (Optional[Dict[str, Any]]) -> Dict[str, Any]
+        '''
+            Example usage:
+
+            >>> client.update_state({'state': {"entry 1": "value 1", "entry 2": "value 2", "entry 3": "value 3"}})
+            >>> client.get_state()
+            {'result': 'success', 'state': {"entry 1": "value 1", "entry 2": "value 2", "entry 3": "value 3"}, 'msg': ''}
+            >>> client.get_state(keys=('entry 1', 'entry 3'))
+            {'result': 'success', 'state': {'entry 1': 'value 1', 'entry 3': 'value 3'}, 'msg': ''}
+        '''
+        return self.call_endpoint(
+            url='user_state',
+            method='GET',
+            request=request,
+        )
+
 class ZulipStream(object):
     """
     A Zulip stream-like object
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -859,13 +859,13 @@ class Client(object):
             request=request,
         )
 
-    def update_state(self, request):
+    def update_storage(self, request):
         # type: (Dict[str, Any]) -> Dict[str, Any]
         '''
             Example usage:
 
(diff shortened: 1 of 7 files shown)


derAnfaenger:
Pushed some more commits, so bots can now actually store state on a Zulip server.

derAnfaenger:
The flask botserver currently creates a new `ExternalBotHandler` instance on each call to handle message. This would lead to StateHandler being reinstantiated over and over, fetching a bot's full state every time. As a follow-up, we should make the flaskserver only keep one instance of each bot's handler for it's runtime.

zulip_bots/zulip_bots/lib.py
```diff
@@ -65,7 +66,8 @@ def put(self, key, value):
 
     def get(self, key):
         # type: (Text) -> Text
-        return self.demarshal(self.state_[key])
+        # Return a copy of the entry to prevent a bot accidentally modifying self.state_ on its own.
```
showell: It seems like we could reorder the commits to have the JSON change happen first, so that we don't need to do the deepcopy here.

showell:
LGTM except for one comment that I made.

Can you open a separate issue for the flaskserver resinstantiation problem?

zulip_bots/zulip_bots/lib.py
derAnfaenger: The problem I see is when bots change the default encoding; The new encoding must be guaranteed to not return any reference to the old object. Maybe we can make an additional check for this?

derAnfaenger: Hmm, but then again I can't see a case where that could occur, except for the original `lamba obj: obj`, which wouldn't work anymore anyway.

derAnfaenger:
@showell updated the PR (merged the two respective commits and removed the deepcopy). Created #147 for the flaskserver issue.

derAnfaenger:
Resolved the merged conflicts.

zulip/zulip/__init__.py
```diff
@@ -859,6 +859,38 @@ def create_user(self, request=None):
...
+            >>> client.get_state({'keys': ["entry 1", "entry 3"]})
+            {'result': 'success', 'state': {'entry 1': 'value 1', 'entry 3': 'value 3'}, 'msg': ''}
+        '''
+        return self.call_endpoint(
+            url='user_state',
+            method='PUT',
+            request=request,
+        )
+
+    def get_state(self, request=None):
```
timabbott: Let's call these e.g. `update_bot_storage`.  We might also want to rename the endpoints to use the `user_storage` name, given what we're calling this feature in the client-side code...

zulip_bots/zulip_bots/lib.py
```diff
@@ -56,15 +56,17 @@ class StateHandler(object):
...
+        self.demarshal = lambda obj: json.loads(obj)
 
     def put(self, key, value):
         # type: (Text, Text) -> None
         self.state_[key] = self.marshal(value)
 
     def get(self, key):
         # type: (Text) -> Text
+        # Ensure demarshal() never returns a reference to state_ to prevent
+        # a bot accidentally modifying self.state_ on its own.
```
timabbott: Is that a TODO, or something else?

timabbott:
I posted a couple comments; the rename could potentially do in a follow-up commit. 

The other thing I'd like to do is only have the bot fetch the storage if the bot has some declaration on the class like `USES_STORAGE = True`, to avoid doing a round-trip for every bot just because some bots support storage.  That can be a follow-up issue though.

@showell feel free to merge once you're satisfied here.

zulip_bots/zulip_bots/lib.py
derAnfaenger: That's a leftover comment, will remove.

derAnfaenger:
@timabbott @showell updated. 
- removed the superfluous comments, 
- renamed get_state() and update_state() to get_storage() and update_storage() (I found these just as meaningful as get_[bot/user]_state(), while shorter.
- added a commit to make the StateHandler optional. I found this to be a nice use case of our META object. Note that this *is* compatible with our embedded bots framework, where the StateHandler is still provided by default.

derAnfaenger:
As a follow-up, should we add the META/bot_details logic to EmbeddedBotHandler?

showell:
Merged!  Thanks @derAnfaenger!

derAnfaenger:
@showell is it possible you commented "merged" without merging?

timabbott:
I bet the push got blocked due to forgetting to rebase.  Happens to me sometimes too :).

showell:
I'm not sure what happened here, but I merged for real now.

PR #144, opened 2017-11-02
Deprecate and remove several bots.

derAnfaenger:
Doing this now, since we are heavily iterating on the bot handler interface, and it's a pain to keep all bots up to date, especially the untested ones.


derAnfaenger:
I removed every github-related bots except for `github_details`. This bot does not depend on the custom `github` module, is tested, and seems to be potentially very useful, especially on servers like `chat.zulip.org`. I think it would make a great candidate for our embedded bots.

timabbott:
Sounds reasonable.  Some of the unmaintained bots we may want to resurrect, but that's OK.  Merged, thanks @derAnfaenger!

PR #145, opened 2017-11-06
Improvements to twitter bot.

derAnfaenger:


zulip/integrations/twitter/twitter-bot
```diff
@@ -33,7 +33,7 @@ import zulip
 VERSION = "0.9"
 CONFIGFILE = os.path.expanduser("~/.zulip_twitterrc")
 INSTRUCTIONS = r"""
-twitter-bot --site=https://zulip.example.com --search="@nprnews,quantum physics"
+twitter-bot --config-file=~/.zuliprc --search="@nprnews,quantum physics"
```
rht: What about for the use case of multiple bots being used simultaneously? E.g. I run various bots with different functionalities and have no choice but to import the config from a relative dir rather than from a global `~/.zuliprc`.

derAnfaenger: You can point the `--config-file` to any file, right?

derAnfaenger: Main point about this change is that the `site` parameter is redundant.

rht: Ah, I see.

rht: Should it be documented that it doesn't necessarily have to be `~/.zuliprc`, perhaps `--config-file=<path to config file>`.

derAnfaenger: So I thought the upper one was supposed to be an example, just like a specific URL was provided. If you go a bit lower, you can find the general forms (also the old general --site argument). Do you think we should should make this clearer?

timabbott:
Merged, thanks @derAnfaenger!

PR #146, opened 2017-11-06
 zulip_bots: Enforce default config file schema.

derAnfaenger:
In the end, I decided against removing the section name. While it is redundant, ConfigParser works only with sections and makes config parsing very convenient. This makes the change for the external bots to comply with embedded bots really straightforward.


showell:
Thanks @derAnfaenger!

PR #149, opened 2017-11-07
Refactor zulip_botserver tests.

derAnfaenger:


zulip_botserver/tests/test_server.py
```diff
@@ -40,18 +40,21 @@ def test_bot_module_not_exists(self):
...
                 'site': 'http://localhost',
             }
         }
-        server.available_bots = available_bots
-        server.load_bot_handlers()
-        mock_LoggingError.assert_called_with("Cannot fetch user profile, make sure you have set up the zuliprc file correctly.")
-
+        # This complains about mismatching argument types, yet they are all correct?
+        # We should investigate and file an issue in mypy.
+        self.assertRaisesRegexp(ImportError,  # type: ignore
```
timabbott: I'm fine with merging this as is, but I bet @neiljp would be up for investigating whether there's a mypy bug here...

timabbott:
Looks good, merged, thanks @derAnfaenger!

PR #150, opened 2017-11-10
Fix XKCD arg parsing

jmandel:


timabbott:
@derAnfaenger can you review this?

jmandel:
(You can probably just treat this a bug report rather than code to merge...)

zulip_bots/zulip_bots/bots/xkcd/xkcd.py
```diff
@@ -49,7 +49,7 @@ class XkcdServerError(Exception):
 
 def get_xkcd_bot_response(message):
     original_content = message['content'].strip()
-    command = original_content.strip()
+    command = original_content.split(' ', 1)[-1].strip()
```
rht: The `.strip()` at this line might be redundant with the previous one.

zulip_bots/zulip_bots/bots/xkcd/xkcd.py
```diff
@@ -68,7 +68,7 @@ def get_xkcd_bot_response(message):
         elif command.isdigit():
             fetched = fetch_xkcd_query(XkcdBotCommand.COMIC_ID, command)
         else:
-            return commands_help % ('xkcd bot only supports these commands:')
+            return commands_help % ("xkcd bot only supports these commands, not '%s':"%command)
```
rht: A lint error: https://travis-ci.org/zulip/python-zulip-api/jobs/299939861#L560

zulip_bots/zulip_bots/bots/xkcd/xkcd.py
jmandel: Technically not since I'm splitting on a single white space and there might be several in between the bottom name and argument. But this is not robust in any case :-) this is probably worth handling in a library that multiple Bots could use.

rht: The logic (removing away the bot @-mention), LGTM.

rht: > ... and there might be several in between the bottom name and argument
Yeah, I see this can be handled via `.split()` (without an explicit `' '`) or regexp split [1]

[1] https://stackoverflow.com/questions/13028120

derAnfaenger:
@jmandel I commited a version of 
```
return commands_help % ("xkcd bot only supports these commands, not '%s':"%command)
```
in #150. The additional tweak to `command` should not be needed anymore, we believe to have fixed the issue in a different commit. Closing this, but feel free to reopen if you encounter the same issue again.

PR #152, opened 2017-11-11
zulip_bots: Fix and simplify extract_query_without_mention.

derAnfaenger:
This fixes quirks related to the regex not covering all potential
weird usernames and adds tests for stripping the @-mentions.


derAnfaenger:
We should make a more general test for the run_message_handler_for_bot flow, where we can integrate the tests created here.

showell:
Merged, thanks @derAnfaenger!

Issue #122, opened 2017-09-16
Get maintained bots to 100% test coverage.


roberthoenig:
**UPDATE Scroll down to the latest comments to find a list of the remaining bot files not at 100% coverage.**

**NOTE: All striked-through bots have been deprecated, all checked bots have been given tests. See the comments for more information on how to proceed.**

We have a couple bots in `zulip_bots/zulip_bots/bots` that are untested:

~~- [ ] commute~~
~~- [ ]  foursquare~~
~~- [ ] git_hub_comment~~
~~- [ ] github~~

- [x] github_issues

~~- [ ] googlesearch~~
~~- [ ] howdoi~~
~~- [ ] john~~
~~- [ ] youtube~~

For each bot, we should **either**

* add tests. You can find help in [our guide for bot testing](http://zulip.readthedocs.io/en/latest/bots-guide.html?highlight=bot%20testing#writing-tests-for-bots).

* deprecate the bot. If a bot seems to be bugged or completely useless, we should discuss it on [chat.zulip.org](chat.zulip.org) and decide whether it is worth keeping.

swekshajha12:
@zulipbot claim

zulipbot:
Welcome to Zulip, @swekshajha12! We just sent you an invite to collaborate on this repository at https://github.com/zulip/python-zulip-api/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](http://zulip.readthedocs.io/en/latest/index.html#code-contribution-guide), as well as the rest of the developer documentation on that site.


See you on the other side (that is, the pull request side)!

zulipbot:
Hello @swekshajha12, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 7 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 3 days.

[...]

PR #153, opened 2017-11-13
zulip_bots: Add common tests to bot testing framework.

derAnfaenger:
This took some experimenting, since unittest doesn't cover this sort of thing. A nice effect is that we have a cheap jump for several bots to 100% coverage.

Because `test_lib.py` gets completely rewritten in the diff, I split out the first commit, which does not alter BotTestCase(TestCase) in any way. This should make it easier to review.

This addresses #122 and moves encrypt, followup, helloworld, help, incrementor, and weather to 100% coverage.

### DIFF
--- a/zulip_bots/zulip_bots/test_lib.py
+++ b/zulip_bots/zulip_bots/test_lib.py
@@ -26,7 +26,7 @@ from types import ModuleType
 
 from copy import deepcopy
 
-class BotTestCase(TestCase):
+class BotTestCaseBase(TestCase):
     bot_name = ''  # type: str
 
     def get_bot_message_handler(self):
@@ -153,3 +153,8 @@ class BotTestCase(TestCase):
         # Strictly speaking, this function is not needed anymore,
         # kept for now for legacy reasons.
         self.call_request(message, expected_method, response)
+
+class BotTestCase(BotTestCaseBase):
+    def test_bot_usage(self):
+        # type: () -> None
+        self.assertNotEqual(self.message_handler.usage(), '')
--- a/tools/test-bots
+++ b/tools/test-bots
@@ -8,6 +8,7 @@ import sys
 import argparse
 import glob
 import unittest
+from unittest import TestCase, TestSuite
 
 def load_tests_from_modules(names, template):
     loader = unittest.defaultTestLoader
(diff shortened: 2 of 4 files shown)


zulip_bots/zulip_bots/test_lib.py
```diff
@@ -26,130 +26,135 @@
...
-                    mock_get.assert_called_with(http_request['api_url'], headers=headers)
-                else:
-                    mock_get.assert_called_with(http_request['api_url'])
-
-    def assert_bot_response(self, message, response, expected_method):
-        # type: (Dict[str, Any], Dict[str, Any], str) -> None
-        # Strictly speaking, this function is not needed anymore,
-        # kept for now for legacy reasons.
-        self.call_request(message, expected_method, response)
+class BotTestCase:
```
timabbott: I think it might be better to keep the existing code structure as `BotTestCaseBase`, and then have this double-layer class just inherit from that and contain a bit of code for the actual tests. That avoids indenting everything twice, and also provides an interface for a bot to inherit from BotTestCaseBase if it doesn't want the common usage tests.

Does that make sense?

derAnfaenger: Yeah, that'd make it a lot nicer!

derAnfaenger:
Updated. Just moving the common tests to a new class `class BotTestCase(BotTestCaseBase)` actually didn't work, because unittest imports all module-level classes that inherit somehow from TestCase. So I added some code to our test runner `test-bots` that manually excludes all occurences of pure BotTestCase tests in the TestSuite.

timabbott:
Merged, with these changes for clarity and avoiding fragility:
```
diff --git a/tools/test-bots b/tools/test-bots
index 7db3c2e..c298bdb 100755
--- a/tools/test-bots
+++ b/tools/test-bots
@@ -98,7 +98,7 @@ def main():
         for test in tests:
             if isinstance(test, TestCase):
                 # Exclude test base class from being tested.
-                if 'BotTestCase' not in test.__class__.__name__:
+                if test.__class__.__name__ not in ['BotTestCase', 'BotTestCaseBase']:
                     filtered_tests.addTest(test)
             else:
                 filtered_tests.addTest(filter_tests(test))
diff --git a/zulip_bots/zulip_bots/test_lib.py b/zulip_bots/zulip_bots/test_lib.py
index f363cf7..48dd05d 100755
--- a/zulip_bots/zulip_bots/test_lib.py
+++ b/zulip_bots/zulip_bots/test_lib.py
@@ -27,6 +27,7 @@ from types import ModuleType
 from copy import deepcopy
 
 class BotTestCaseBase(TestCase):
+    """Test class for common Bot test helper methods"""
     bot_name = ''  # type: str
 
     def get_bot_message_handler(self):
@@ -155,6 +156,8 @@ class BotTestCaseBase(TestCase):
         self.call_request(message, expected_method, response)
 
 class BotTestCase(BotTestCaseBase):
+    """Test class extending BotTestCaseBase to add common default tests
+    that should be run (by default) for all our bots"""
     def test_bot_usage(self):
         # type: () -> None
         self.assertNotEqual(self.message_handler.usage(), '')
```

PR #156, opened 2017-11-14
zulip_bots: Update README.md

derAnfaenger:
This stubs the zulip_bots README.md
and moves architectural information
to architecture.md.

@showell FYI. 

The architecture.md file is probably not optimal yet. It's mainly composed by moving over stuff from the old doc and deleting the outdated/redundant parts.

### DIFF
--- a/zulip_bots/README.md
+++ b/zulip_bots/README.md
@@ -1,143 +1,25 @@
-# Contrib bots
+# Zulip bots
 
-This is the documentation for an experimental new system for writing
-bots that react to messages. It explains how to run the code, and also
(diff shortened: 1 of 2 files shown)


timabbott:
This is an improvement, merged, thanks @derAnfaenger!  I think in terms of high-value documentation work, we should focus on moving the ReadTheDocs bots documentation to /api...

derAnfaenger:
Right, will work on that tomorrow.

showell:
@derAnfaenger Thanks for keeping the architecture stuff!

Issue #142, opened 2017-10-27
mypy: Clean out or report issues in mypy/typeshed for `type: ignores`


timabbott:
Now that we've finished #127, it's worth doing a bit of work on the `type: ignores` in our mypy annotations, trying to fix upstream in typeshed anything where that's possible.

I don't want this issue to remain open forever, so we should happily close this as soon as we've done a sweep through our `type: ignore` entries.

@neiljp I feel like you'd probably be pretty effective at doing this.

neiljp:
@zulipbot claim

neiljp:
#157 Should cover most of this, I think?

I think I had a copy/paste issue, as the work on generate_manifest.py I intended to link to 
python/mypy#1747, which was my reported issue; the corresponding PR I just worked on is python/mypy#1748.

timabbott:
Yeah, this seems good enough for this issue (though of course you're welcome to contribute fixes upstream to clean this out further!).  Thanks @neiljp!

neiljp:
For the record, the correct links I reported and which are now fixed were in python/typeshed, not python/mypy. I'm just putting together a PR which removes these ignores and upgrades to mypy 0.560.

PR #157, opened 2017-11-15
mypy: Improve 'type: ignore' annotations.

neiljp:
Related to #142.


timabbott:
This looks reasonable to me assuming Travis passes.

neiljp:
Pushed version without "python3" in comment, which triggers lint error.

PR #166, opened 2017-11-22
Report client.get_profile() errors nicely (e.g. API key).

showell:
This patch is particularly useful in the scenario that your
API key is wrong or out of date, but it's targeted more
generally at any error that `client.get_profile()` reports.


timabbott:
Is this something we should add a unit test for?  Generally lgtm.

showell:
We eventually should get a unit test here, but for now I just want to get it fixed, and I don't think there's huge risk of people breaking this going foward.  So about to merge.

PR #167, opened 2017-11-23
Fix provisioning Python version mixup.

derAnfaenger:
@showell can you test this with Py 2 and 3?


derAnfaenger:
I decided against ditching Python 2 completely, since developers still want to test the `zulip` package with Python 2.

PR #169, opened 2017-11-25
Bots: Add default (and other) command dispatch

neiljp:
This is an alternative implementation of #66 based on discussion in #integrations regarding perceived issues (@showell, @timabbott)

I avoided amending existing bots and instead added a demonstration as a 'helloworld_defaults' bot, which demonstrates the increasing features in subsequent commits:
- empty ("") and "about" command responses
- "commands" (uses default commands and those passed in)
- "help" (default commands, and those passed in with help text)
- allowing dispatch of all commands through the system
- a decorator which can be applied to functions, to indicate the command name and help text, and which causes the function to be dispatched if the command matches.

I attempted to develop things in tandem in the test system, but am not fully satisfied with the current system (as per the comments in test_helloworld_defaults).

There is no doc.md file as yet for the new bot, though I can add this easily.

I'm completely open to renaming aspects of the system, but functionally it appears to work fine in tests and with zulip_bot_output.py (or did until I rebased, and now have some kind of storage error? @roberthoenig?)


zulip_bots/zulip_bots/lib.py
```diff
@@ -82,6 +84,23 @@ def contains(self, key):
...
 
+class NotADefaultCommand(Exception):
+    def __init__(self, command, supported):
+        # type: (Text, Sequence[Text])
+        self.msg = "'{}' is not a supported default command; options are: {}".format(
+            command, ", ".join([repr(s) for s in supported]))
+    def __str__(self):
+        return self.msg
+
+to_dispatch = {}
```
showell: This doesn't feel right, having `to_dispatch` being a module-level variable.

neiljp: Would you prefer this in eg. `ExternalBotHandler`? Or in a separate class?

Is this approach preferable to #66?

PR #170, opened 2017-11-27
Get 100% test coverage for Giphy bot.

roberthoenig:
This comes along with a few improvements to the testing system (and bot improvements).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,13 @@
 language: python
 python:
-  - "2.7"
   - "3.4"
   - "3.5"
   - "3.6"
 env: TEST_SUITE=test-main
 matrix:
   include:
+  - python: "2.7"
+    env: TEST_SUITE=test-py2-legacy-support
   - python: "3.6"
     env: TEST_SUITE=test-static-analysis
 install:
--- a/zulip_bots/zulip_bots/test_lib.py
+++ b/zulip_bots/zulip_bots/test_lib.py
@@ -137,7 +137,7 @@ class BotTestCaseBase(TestCase):
             http_headers = http_data.get('response-headers')
             with patch('requests.get') as mock_get:
                 mock_result = requests.Response()
-                mock_result._content = json.dumps(http_response).encode()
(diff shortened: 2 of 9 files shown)


roberthoenig:
This fails because of Python 2 incompatibility. I'm gonna update Travis tests so `zulip_bots` is excluded from the Python 2 tests.

.travis.yml
```diff
@@ -1,12 +1,13 @@
 language: python
 python:
-  - "2.7"
   - "3.4"
   - "3.5"
   - "3.6"
 env: TEST_SUITE=test-main
 matrix:
   include:
+  - python: "2.7"
```
showell: Do we actually need this?  If so, I would clarify the commit message.

timabbott: Looks correct to me.

timabbott: (But I think the commit message could benefit here, yeah)

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -136,12 +136,9 @@ def mock_http_conversation(self, test_name):
             http_response = http_data.get('response')
             http_headers = http_data.get('response-headers')
             with patch('requests.get') as mock_get:
-                mock_result = mock.MagicMock()
-                mock_result.json.return_value = http_response
-                if 'text' in http_response:
-                    mock_result.text = http_response.get('text', None)
+                mock_result = requests.Response()
```
showell: I would explain with a comment how it's safe to use a "real" `Response` object in a `mock_result` context.

timabbott:
@eeshangarg FYI, if we're dropping Python 2 support for zulip_bots and zulip_botserver, there will be corresponding packaging changes we want to make (I'd just removing the packaging tags, and not do a bunch of migrations to python-3-only stuff, for now).

showell:
Ok. I merged this.  I'm a little unsure on a couple things, but I don't want this to get stale when Robert's on vacation and when I get pulled away to other stuff.  If my concerns are valid, we can always fix forward.

Thanks @roberthoenig!

timabbott:
The main concern is that we need to update the release tools to not try releasing a Python 2.7 copy.

PR #176, opened 2017-11-28
Improve bots (misc)

showell:


roberthoenig:
@showell we should update the documentation for this (and remove the zulip-bot-output section): https://zulipchat.com/api/writing-bots#testing-a-bots-output

showell:
Yep, we should.  We should create a follow-up issue.

PR #178, opened 2017-11-28
bots: Clean up terminal.py for running bots directly.

abhijeetkaur:
Make terminal.py exit gracefully with a message.
Modify output bot reply for better understanding for multi-line bot output. For example:
```
>python3 terminal.py --bot-config-file bots/weather/weather.conf weather 
Enter your message: New York

            reply:
            Weather in New York, US:
51.53 F / 10.85 C
Scattered Clouds
```

On providing no config file flag in simple_lib.py for a bot that uses conf, throws the following error:
`NameError: name 'sys' is not defined`
Corrected this.

### DIFF
--- a/zulip_bots/zulip_bots/terminal.py
+++ b/zulip_bots/zulip_bots/terminal.py
@@ -56,18 +56,23 @@ def main():
 
     sender_email = 'foo_sender@zulip.com'
 
-    while True:
-        content = input('Enter your message: ')
+    try:
+        while True:
+            content = input('Enter your message: ')
 
-        message = dict(
-            content=content,
-            sender_email=sender_email,
-            display_recipient=sender_email,
-        )
-        message_handler.handle_message(
-            message=message,
-            bot_handler=bot_handler,
-        )
+            message = dict(
+                content=content,
+                sender_email=sender_email,
+                display_recipient=sender_email,
+            )
+            message_handler.handle_message(
+                message=message,
+                bot_handler=bot_handler,
+            )
+    except KeyboardException:
+        print("\n\nOk, if you're happy with your terminal-based testing, try it out with a Zulip server.",
+              "\nYou can refer to https://zulipchat.com/api/running-bots#running-a-bot.")
+        sys.exit(1)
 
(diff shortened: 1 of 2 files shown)


showell:
@abhijeetkaur This looks great, but when you have time to work on this, there are two remaining tasks:

* debug the test failure
* catch `KeyboardException` specifically

Thanks!

zulip_bots/zulip_bots/terminal.py
```diff
@@ -56,18 +56,22 @@ def main():
...
+            message = dict(
+                content=content,
+                sender_email=sender_email,
+                display_recipient=sender_email,
+            )
+            message_handler.handle_message(
+                message=message,
+                bot_handler=bot_handler,
+            )
+    except:
```
neiljp: The build is failing for exactly what I would request as a change, namely a more-specific `except` statement here :)

zulip_bots/zulip_bots/terminal.py
```diff
@@ -56,18 +56,22 @@ def main():
...
+        print("\n\nOk, if you're happy with your terminal-based testing, try it out with a Zulip server.\nYou can refer to https://zulipchat.com/api/running-bots#running-a-bot.")
```
neiljp: Can this line be adjusted? At least on here, it looks very long.

showell:
Merged, thanks @abhijeetkaur!

zulip_bots/zulip_bots/terminal.py
```diff
@@ -56,18 +56,23 @@ def main():
...
+                content=content,
+                sender_email=sender_email,
+                display_recipient=sender_email,
+            )
+            message_handler.handle_message(
+                message=message,
+                bot_handler=bot_handler,
+            )
+    except KeyboardException:
+        print("\n\nOk, if you're happy with your terminal-based testing, try it out with a Zulip server.",
```
showell: @abhijeetkaur We should use triple-quoting for this kind of stuff in the future.  It indents the actual output, which is sort of unintentional, but it actually looks fine to the user.

PR #179, opened 2017-11-29
docs: Mention Python 3 in the README.

skunkmb:
As mentioned in [this Zulip conversation](https://chat.zulip.org/#narrow/stream/GCI.20tasks/subject/Python.202.20or.203.20in.20python-zulip-api/near/347329), the docs don't mention whether or not to use Python 2 or 3, so I made a commit to make it more clear.


showell:
Merged!  Thanks @skunkmb!

PR #183, opened 2017-11-30
Tests: Extend test-lib script to support coverage parameter & add to test-main

neiljp:


showell:
Merged, thanks @neiljp!

PR #189, opened 2017-11-30
Move tests toward more transparent stubbing.

showell:
This is still a work in progress.


showell:
This is actually ready for review now.  I can explain the big picture on chat, but the goal is to essentially remove some of the complexity from BotTestCase.  I think I have a pretty good plan not to have to duplicate too much code.

PR #190, opened 2017-12-01
Slack importer

rheaparekh:
This is based on  https://github.com/zulip/python-zulip-api/pull/114.
Note: This had only users and channels import implementation.
# Usage

See https://my.slack.com/services/export for the step to export a Slack data.

- To convert a slack data, let the zip file be `slack_data.zip`, then use `./slackdata2zulipdata.py slack_data.zip` 
(sample slack data: https://github.com/houstondatavis/slack-export)

- This outputs `zulip_data.tar.gz`. Use `tar xf zulip_data` to uncompress into a folder.

- To deploy the data onto an existing fresh Zulip repo,
`./manage.py import --destroy-rebuild-database <path to zulip_data folder>`


showell:
Merged!  Thanks @rheaparekh!

I haven't given this a thorough review yet, but I think it's good to have the code merged, so that other folks can easily start to test it and suggest improvements.

PR #192, opened 2017-12-02
zulip_bots: Create Link Shortener Bot.

skunkmb:
Hello! Here is the Link Shortener Bot I created for the *Create your own URL shortener bot* Google Code-in task. I have tested it both manually and with the automatic `BotTestCase` class.

Before use and testing, the `key` property must be set to a goo.gl API key in `link_shortener.conf` and in the `mock_config_info` call in `test_link_shortener.py`.

In use, for example, you could write

 > @**Link Shortener Bot** @**Joe Smith** See https://github.com/zulip/python-zulip-api/tree/master/zulip_bots/zulip_bots/bots for a list of all Zulip bots.

and LS Bot would respond

 > https://github.com/zulip/python-zulip-api/tree/master/zulip_bots/zulip_bots/bots: https://goo.gl/NjLZZH

Please let me know if there is something I should improve or fix! Thank you!

### DIFF
--- a/zulip_bots/zulip_bots/bots/link_shortener/link_shortener.py
+++ b/zulip_bots/zulip_bots/bots/link_shortener/link_shortener.py
@@ -1,0 +1,80 @@
+import re
+import requests
+
+class LinkShortenerHandler(object):
+    '''A Zulip bot that will shorten URLs ("links") in a conversation using the
+    goo.gl URL shortener.
+    '''
+
+    def usage(self):
+        return (
+            'Mention the link shortener bot in a conversation and then enter '
+            'any URLs you want to shorten in the body of the message. \n\n'
+            '`key` must be set in `link_shortener.conf`.')
+
+    def initialize(self, bot_handler):
+        self.config_info = bot_handler.get_config_info('link_shortener')
+
+    def handle_message(self, message, bot_handler):
+        REGEX_STR = (
+            '('
+            '(?:http|https):\/\/'  # This allows for the HTTP or HTTPS
+                                   # protocol.
+            '[^"<>#%\{\}|\\^~[\]` ]+'  # This allows for any character except
+                                       # for certain non-URL-safe ones.
+            ')'
+        )
+
+        content = message['content']
+
+        if content.strip() == 'help':
+            bot_handler.send_reply(
+                message,
+                (
+                    'Mention the link shortener bot in a conversation and '
+                    'then enter any URLs you want to shorten in the body of '
+                    'the message.'
+                )
+            )
+            return
+
+        link_matches = re.findall(REGEX_STR, content)
+
+        shortened_links = [self.shorten_link(link) for link in link_matches]
+        link_pairs = [
+            (link_match + ': ' + shortened_link)
+            for link_match, shortened_link
+            in zip(link_matches, shortened_links)
+            if shortened_link != ''
+        ]
+        final_response = '\n'.join(link_pairs)
+
+        if final_response == '':
+            bot_handler.send_reply(
+                message,
+                'No links found. Send "help" to see usage instructions.'
+            )
+            return
+
+        bot_handler.send_reply(message, final_response)
+
+    def shorten_link(self, long_url):
+        '''Shortens a link using goo.gl Link Shortener and returns it, or
+        returns an empty string if something goes wrong.
+
+        Parameters:
+            long_url (str): The original URL to shorten.
+        '''
+
+        body = {'longUrl': long_url}
+        params = {'key': self.config_info['key']}
+
+        request = requests.post(
+            'https://www.googleapis.com/urlshortener/v1/url',
+            json=body,
+            params=params
+        )
+
+        return request.json().get('id', '')
+
+handler_class = LinkShortenerHandler
--- a/zulip_bots/zulip_bots/bots/link_shortener/test_link_shortener.py
+++ b/zulip_bots/zulip_bots/bots/link_shortener/test_link_shortener.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python
+
+from zulip_bots.test_lib import BotTestCase
+
+class TestLinkShortenerBot(BotTestCase):
+    bot_name = "link_shortener"
+
+    def test_bot(self):
+        MESSAGE = 'Shorten https://www.github.com/zulip/zulip please.'
+        RESPONSE = 'https://www.github.com/zulip/zulip: https://goo.gl/6uoWKb'
+
+        with self.mock_config_info({'key': 'qwertyuiop'}), \
+                self.mock_http_conversation('test_normal'):
+            self.initialize_bot()
+
+            self.assert_bot_response(
+                message = {'content': MESSAGE},
+                response = {'content': RESPONSE},
+                expected_method='send_reply'
+            )
+
+    def test_bot_empty(self):
+        MESSAGE = 'Shorten nothing please.'
+        RESPONSE = 'No links found. Send "help" to see usage instructions.'
+
+        # No `mock_http_conversation` is necessary because the bot will
+        # recognize that no links are in the message and won't make any HTTP
+        # requests.
+        with self.mock_config_info({'key': 'qwertyuiop'}):
+            self.initialize_bot()
+
+            self.assert_bot_response(
+                message = {'content': MESSAGE},
+                response = {'content': RESPONSE},
+                expected_method='send_reply'
+            )
+
+    def test_bot_help(self):
+        MESSAGE = 'help'
+        RESPONSE = (
+            'Mention the link shortener bot in a conversation and then enter '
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -13,6 +13,8 @@ This repository contains the source code for Zulip's PyPI packages:
 * `zulip_botserver`: [PyPI package](https://pypi.python.org/pypi/zulip-botserver)
   for Zulip's Flask bot server.
 
+The source code is written in *Python 3*.
```
neiljp: While this is a minor update, and is in a separate commit, I would suggest submitting this in a separate PR.

You should be able to cherry-pick the commit into another branch, though it is just one line :)

zulip_bots/zulip_bots/bots/link_shortener/link_shortener.py
```diff
@@ -0,0 +1,68 @@
...
+
+        request = requests.post(
+            'https://www.googleapis.com/urlshortener/v1/url',
+            json=body,
+            params=params
+        )
+
+        response = request.json()
+
+        if 'id' in response:
```
neiljp: Could you use `.get()` here to simplify the return?

README.md
skunkmb: Hmm. I actually *did* mention this in a separate PR: https://github.com/zulip/python-zulip-api/pull/179 but I guess it got carried over to this one too. I will remove it from the PR now.

skunkmb:
Hi @neiljp, I removed the commit about Python 3 (it is in a [separate PR](https://github.com/zulip/python-zulip-api/pull/179)) and changed the 

```python
response = request.json()

if 'id' in response:
    return response['id']

return ''
```

to just

```python
return request.json().get('id', '')
```

as you mentioned. Are there any other changes I should make?

zulip_bots/zulip_bots/bots/link_shortener/link_shortener.py
```diff
@@ -0,0 +1,63 @@
...
+        body = { 'longUrl': long_url }
+        params = { 'key': self.config_info['key'] }
+
+        request = requests.post(
+            'https://www.googleapis.com/urlshortener/v1/url',
+            json=body,
+            params=params
+        )
+
+        return request.json().get('id', '').
```
aero31aero: There is a syntax error on this line. See the trailing '.'.

skunkmb: OK, I fixed it in commit 17ac82d3e930700e2d26d7bdf63d407f5cb99ed8. Thanks!

zulip_bots/zulip_bots/bots/link_shortener/test_link_shortener.py
```diff
@@ -0,0 +1,22 @@
...
+    bot_name = "link_shortener"
+
+    def test_bot(self):
+        with self.mock_config_info(
+            { 'key': '<your API key>' }
+        ):
+            self.initialize_bot()
+
+        message = 'Shorten https://www.github.com/zulip/zulip please.'
+        response = 'https://www.github.com/zulip/zulip: https://goo.gl/6uoWKb'
```
aero31aero: Is the bot making an http request to the API here? You should stub that request as done in the tests for giphy bot.

skunkmb: No request to the API is being made on lines 14 and 15. Since goo.gl always returns the same shortened link for the same url (as long as the REST API is being used), https://www.github.com/zulip/zulip will always be converted to https://goo.gl/6uoWKb. Because of this, `test_bot` just tests if given that `message`, the `response` is `'https://www.github.com/zulip/zulip: https://goo.gl/6uoWKb'`.

I'm not sure what "stubbing" is, can you please explain it? I looked in the giphy test and did a search for "stub" but there were no results. I also looked at some articles online, and they mentioned using "mock configurations," but I'm not entirely sure how to use those either. Thank you very much!

skunkmb:
Thanks, @aero31aero! I think the reason the Travis CI build fails is because no API key is configured in the `test_link_shortener.py`. On line 10, I put a placeholder for an API key (`<your API key>`), but when a request is made with that placeholder, an error is returned. How would you recommend I proceed? Should I hard-code a key?

zulip_bots/zulip_bots/bots/link_shortener/test_link_shortener.py
aero31aero: Stubbing is replacing an actual function with a fake function so that we can test it in isolation. See the call to `mock_http_conversation` in the giphy tests. Also, look at the `fixtures/*.json` files that show mock http conversations.

So, yes, the line 14-15 are making a request to the API but getting the same result back. This means that we can use this result to create a mock http conversation now

aero31aero:
You shouldn't have to hardcode a key. This is why we want to mock the http conversation and test your bot's code in isolation. I left a comment above explaining this briefly. Feel free to message on the chat if you have any doubts.

skunkmb:
I just pushed a new commit using a mock. However, since the goo.gl API uses POST requests, I had to add a POST request feature to the `mock_http_conversation` function in `test_lib.py`, as discussed here: https://goo.gl/fuWjf9. I made a PR with those changes at #194. Until that PR is merged, though, I don't think Travis is going to say it passes the tests, because the POST request won't be detected by `mock_http_conversation`.

zulip_bots/zulip_bots/bots/link_shortener/fixtures/test.json
```diff
@@ -0,0 +1,19 @@
+{
```
aero31aero: Consider renaming this to `test_succes.json`.

zulip_bots/zulip_bots/bots/link_shortener/link_shortener.py
```diff
@@ -0,0 +1,63 @@
...
+        shortened_links = [self.shorten_link(link) for link in link_matches]
+        link_pairs = [
+            (link_match + ': ' + shortened_link)
+                for link_match, shortened_link
+                in zip(link_matches, shortened_links)
+                if shortened_link != ''
+        ]
+        final_response = '\n'.join(link_pairs)
+
+        if final_response == '':
```
aero31aero: Consider adding some text like: `No links found. Send 'help' to see usage instructions.`. Also, add a test for the case where we don't send any link.

zulip_bots/zulip_bots/bots/link_shortener/requirements.txt
```diff
@@ -0,0 +1 @@
+request
```
aero31aero: This file shouldn't be needed. Aside from the fact that as of now we don't install from each bot's requirements files, the requests module is already a dependency.

You should add it to `zulip_bots/setup.py` instead, for future proofing as follows:


```python
setuptools_info = dict(
    install_requires=[
        'zulip',
        'mock>=2.0.0',
        'html2text',  # for bots/define
    ],
)
```

zulip_bots/zulip_bots/bots/link_shortener/test_link_shortener.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python
+
+from zulip_bots.test_lib import BotTestCase
+
+class TestMessageInfoBot(BotTestCase):
```
aero31aero: I think you forgot to rename this class.

skunkmb:
@aero31aero Thanks for the feedback! I made some new commits to accommodate for those changes.

showell:
Looks like the tests are still failing.

skunkmb:
@showell I just pushed a commit to fix some linter errors (b0e0042), and now the tests pass! I think Travis just needed a new commit to make it check again, now that PR #194 has been merged.

Is there anything else I should change, or is this ready to be merged? Thanks!

showell:
LGTM once you squash the 2nd commit.

skunkmb:
@showell OK, squashed. Thanks!

neiljp:
I was thinking of suggesting you could annotate the bot for types, but let's save that for another PR/commit once that's discussed further elsewhere (as bots aren't checked by default by mypy, as it stands)

skunkmb:
@neiljp OK, let me know if you change your mind. Thanks!

neiljp:
@skunkmb This has been discussed and may lead to other GCI tasks:
  https://chat.zulip.org/#narrow/stream/integrations/topic/Partial.20Mypy.20annotation

It is certainly not currently a requirement for bots, and I believe your bot has been provisionally included to be typed in the new GCI tasks, but if you wanted to add type annotations I don't imagine there'd be objections - though you could keep them in a separate commit to make it easier to cherry-pick into another task/PR later.

skunkmb:
@neiljp Thanks! Maybe after the task I'm working on now (Chess Bot), I'll create a new PR adding the type annotations.

showell:
Merged, thanks @skunkmb!

Issue #187, opened 2017-11-30
Fully move zulip_bots to Python 3.


roberthoenig:
We have dumped Python 2 support for the zulip_bots and zulip_botserver packages. I am opening this issue to track progress on converting remaining code shreds from Python 2 legacy support to Python 3 only.

- [x] - move all shebangs from `#!/usr/bin/env python` to `#!/usr/bin/env python3`
- [ ] - move remaining six imports
- [ ] - remove any python 2-related checks.

Feel free to comment additional steps to be taken here.

roberthoenig:
@fredfishgames assigning you for the first checkbox :) When you submit your PR, mention this issue in the PR description.

sibasish14:
@roberthoenig can I work on this issue for the first check?
I have to change the `python` to `python3` in all shebangs of `zulip_bots` and `zulip_botserver` packages right?

fredfishgames:
@sibasish14 We decided to do this for `zulip_bots`, but not the test scripts. @showell did it in [this commit](https://github.com/zulip/python-zulip-api/commit/fd069dff82b98785ea23183f7f5313c782351474).

sibasish14:
Ok @fredfishgames
Can you suggest any other issue though? I'm a beginner by the way.
Or should I look for a beginner issue in the main zulip repo?

showell:
We can actually change the test scripts that pertain to zulip_bots.

(My earlier comments were probably me just being cautious until we were sure, but I'm confident now that we want to make the move.)

sibasish14:
@showell  you want `python3` shebangs to be introduced in the `zulip_bot` test scripts? So can I do that?

sibasish14:
@showell waiting for your reply :)
If you could please tell me if you wanted to proceed ahead, I could start working. Else I'll look for some other issue :(

neiljp:
@sibasish14: @showell may well be away for the holidays right now, so you might want to wait a few days at least - is there something else you can work on?

sibasish14:
@neiljp : may be I'll move on to some other issue for now. But I'll start working on this one once @showell gives me the green signal ;)

showell:
yes, python3 shebangs for the zulip_bot test scripts would be good

sibasish14:
@showell I'll get to work.

sibasish14:
Can someone please have a look at https://github.com/zulip/python-zulip-api/pull/259 ?

showell:
[...]

PR #193, opened 2017-12-02
Fully move zulip_bots to Python 3.

fredfishgames:
Fixing issue #187

### DIFF
--- a/tools/custom_check.py
+++ b/tools/custom_check.py
@@ -164,7 +164,7 @@ def build_custom_checkers(by_lang):
          'exclude': set(['tools/provision']),
          'include_only': set(['zulip/', 'tools/', 'zulip_botserver/']),
          'description': 'Explicit python invocations should not include a version'},
-        {'pattern': '#!/usr/bin/env python[ 2]',
+        {'pattern': '#!/usr/bin/env python$',
          'include_only': set(['zulip_bots/']),
          'description': 'Python shebangs must be python3'},
         {'pattern': '(^|\s)open\s*\(',
--- a/tools/custom_check.py
+++ b/tools/custom_check.py
@@ -162,7 +162,11 @@ def build_custom_checkers(by_lang):
         # and comments which should be rewritten to avoid use of "python2", "python3", etc.
         {'pattern': 'python[23]',
          'exclude': set(['tools/provision']),
+         'include_only': set(['zulip/', 'tools/', 'zulip_botserver/']),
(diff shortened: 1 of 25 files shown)


tools/custom_check.py
```diff
@@ -161,7 +161,7 @@ def custom_check_file(fn, identifier, rules, skip_rules=None, max_length=None):
         # This rule might give false positives in virtualenv setup files which should be excluded,
         # and comments which should be rewritten to avoid use of "python2", "python3", etc.
         {'pattern': 'python[23]',
-         'exclude': set(['tools/provision']),
+         'exclude': set(['tools/provision', 'zulip_bots/']),
```
roberthoenig: I think you need to remove the `/` in `zulip_bots/` for this to work.

fredfishgames:
Hi,
I made the changes you suggested. It still failed the tests...
I've also added a test to make sure python versions are python3

fredfishgames:
Ok, I got rid of that test.. It was creating a bunch of false positives in comments etc.

rht:
@fredfishgames see https://github.com/rht/python-zulip-api/commit/d11ad48342544f0bc10d8e8bd0dd5e3c8b7bb47c for the pattern needed for excluding `zulip_bots/` in 'python[23]' custom check. I ran the static analysis test locally and did not observe any failing.

fredfishgames:
@rht Thanks! Didn't think of that

showell:
@roberthoenig Can you merge this when you're happy with it and plan to be online for a while?  (I want to make sure you're available to revert it quickly if people have unexpected problems.)

rht:
(Here is a case in zulip/zulip for having both `include_only` and `exclude` https://github.com/zulip/zulip/blob/4fd3492e737b38c79e781a24f1c0415cd824e6c7/tools/linter_lib/custom_check.py#L621-L633 -- also note there is a trailing slash at the end.)

fredfishgames:
I've added a rule to the linter which makes sure any shebangs are python3, however I have not pushed it

fredfishgames:
I'll be back in an hour to push it

showell:
This commit got kind of stale, but I just merged a fix, borrowing the lint code from @fredfishgames and then manually fixing the lint errors myself (there were lots of conflicts due to recent activity, and we wanted to just remove shebangs from the test modules, so it was easiest just to re-do it).  Closing.

Thanks @fredfishgames!

PR #194, opened 2017-12-02
zulip_bots: Test POST requests in `mock_http_conversation`.

skunkmb:
As of now, only GET requests are supported by `mock_http_conversation`. However, this creates problems as some APIs utilize POST requests instead of GET ones. Therefore, create a system for using POST requests. (See https://goo.gl/p77qab for a discussion of this on the Zulip #integrations stream.)

To use a POST request with `mock_http_conversation`, specify
`"method": "POST"` in the `"request"` options of a fixture. For example,

    "request": {
      "api_url": "https://example.com/api",
      "method": "POST",
      "params": {
        "foo": "bar",
        "baz": "qwerty"
      },
      "json": {
        "baz": "qwerty"
      }
    },

### DIFF
--- a/zulip_bots/zulip_bots/test_lib.py
+++ b/zulip_bots/zulip_bots/test_lib.py
@@ -159,6 +159,13 @@ def read_bot_fixture_data(bot_name, test_name):
     http_data = json.loads(content)
     return http_data
 
+def get_response(http_response, http_headers):
+    # type: (Dict[str, Any], Dict[str, Any]) -> Any
+    mock_result = requests.Response()
+    mock_result._content = json.dumps(http_response).encode()  # type: ignore # This modifies a "hidden" attribute.
+    mock_result.status_code = http_headers.get('status', 200)
+    return mock_result
+
 @contextmanager
 def mock_http_conversation(http_data):
     # type: (Dict[str, Any]) -> Any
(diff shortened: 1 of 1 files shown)


zulip_bots/zulip_bots/test_lib.py
```diff
@@ -159,6 +159,13 @@ def read_bot_fixture_data(bot_name, test_name):
     http_data = json.loads(content)
     return http_data
 
+def patch_response(http_response, http_headers):
```
aero31aero: I believe we should rename this function to `get_response` or `get_response_object`. Patching can be confused with [this](https://stackoverflow.com/questions/5626193/what-is-monkey-patching).

aero31aero:
Great. Now you just need to wait for a final review. :)

showell:
Overall this looks reasonable, but I think you need to start over on the latest copy of master and mostly squash this into one commit.

skunkmb:
@showell OK, I squashed it into one commit. Is there anything else I should change? Thanks!

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -159,6 +159,13 @@ def read_bot_fixture_data(bot_name, test_name):
     http_data = json.loads(content)
     return http_data
 
+def get_response(http_response, http_headers):
+    # type: (Dict[str, Any], Dict[str, Any]) -> Any
+    mock_result = requests.Response()
+    mock_result._content = json.dumps(http_response).encode()  # type: ignore # This modifies a "hidden" attribute.
+    mock_result.status_code = http_headers.get('status', 200)
+    return mock_result
```
roberthoenig: Since this method is only used in `mock_http_conversation`, it probably makes sense to define it directly in `mock_http_conversation`.

roberthoenig: That way, we also avoid the trouble with finding a decent name for this function that is not  too confusing for the general namespace.

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -174,20 +181,36 @@ def mock_http_conversation(http_data):
...
+            if 'params' in http_request:
+                params = http_request.get('params', None)
+                mock_get.assert_called_with(http_request['api_url'], params=params)
+            elif 'headers' in http_request:
+                headers = http_request.get('headers', None)
+                mock_get.assert_called_with(http_request['api_url'], headers=headers)
+            else:
+                mock_get.assert_called_with(http_request['api_url'])
+    else:
+        with patch('requests.post') as mock_post:
```
roberthoenig: The two codepaths in each `with` statement are almost identical. I actually think that the codepath for `mock_post` should also work with `mock_get`. Can you try de-duplicating code by putting it all into a function inside `mock_http_conversation` that is then called in each with statement?

skunkmb:
@roberthoenig OK thanks! I made some changes in this commit: d078f2f0a8896fbd484e649230ed06a7270d4138. I moved `get_response` inside of `mock_http_conversation`, and I also created a new function in `mock_http_conversation`: `assert_called_with_fields`. 

`assert_called_with_fields` takes a mock object, HTTP request, and list of fields as parameters, and then calls the `assert_called_with` method on the mock using each field listed along with it's value from the HTTP request. That let me remove almost all of the duplicated code between the GET `with` block and POST `with` block. Thanks!

showell:
Merged!  Thanks @skunkmb 

@roberthoenig If you think there are opportunities for further refinement here, feel free to submit a follow-up PR.  Looks pretty clean now!

PR #195, opened 2017-12-04
tools/run-mypy: Allow individual bots to opt in to type-checking locally

neiljp:
Adding a 'type_check' file in a bot directory will cause run-mypy to check all the files in that directory.


neiljp:
Alternative implementation already merged, so closing.

PR #196, opened 2017-12-04
Add mypy annotations for hello world bot.

aero31aero:
This is a sample PR for the GCI task mypy-annotations Type B.

### DIFF
--- a/tools/run-mypy
+++ b/tools/run-mypy
@@ -32,6 +32,12 @@ exclude = [
     "tools/test-lib",
 ]
 
+# These files will be included even if excluded by a rule above.
+force_include = [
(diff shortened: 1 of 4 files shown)


roberthoenig:
@aero31aero thanks for working on this! I'm wary of modifying the `lister` code, since our medium-term goal is to use a common linter repo for all repos - diverting would make that more difficult.

I think there is a really simple solution at hand: Create a `force_include` list of all files we want to include, no matter what. Then, append that list to `python_files`. This shouldn't require modifying anything other than `run-mypy`.

showell:
Merged!  Thanks @aero31aero @roberthoenig 

@neiljp fyi

PR #197, opened 2017-12-04
interactive bots: Create googletranslate bot.

fredfishgames:
The Google translate bot translates messages using Google translate.
It uses the [google translate web API](https://cloud.google.com/translate/docs/reference/rest), so it requires a Google Cloud auth key to work. It used to use an unofficial python API, but the developer of that said that it could break at any moment, and I couldn't get any of the tests to work with it.

### DIFF
--- a/zulip_bots/zulip_bots/bots/googletranslate/googletranslate.conf
+++ b/zulip_bots/zulip_bots/bots/googletranslate/googletranslate.conf
@@ -1,0 +1,2 @@
+[googletranslate]
+key=your_api_key_here
--- a/zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
+++ b/zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
@@ -1,0 +1,103 @@
+# To use this plugin, you need to set up the Google Cloud API key for this bot in
+# googletranslate.conf in this (zulip_bots/bots/googletranslate/) directory.
+
+from __future__ import absolute_import
+from __future__ import print_function
+
+import requests
+
+class GoogleTranslateHandler(object):
+    '''
+    This bot will translate any messages sent to it using google translate.
+    Before using it, make sure you set up google api keys, and enable google
+    cloud translate from the google cloud console.
+    '''
+    def usage(self):
+        return '''
+            This plugin allows users translate messages
+            Users should @-mention the bot with the format
+            @-mention "<text_to_translate>" <target-language> <source-language(optional)>
+            '''
+
+    def initialize(self, bot_handler):
+        self.config_info = bot_handler.get_config_info('googletranslate')
+        self.supported_languages = get_supported_languages(self.config_info['key'])
+
+    def handle_message(self, message, bot_handler):
+        bot_response = get_translate_bot_response(message['content'],
+                                                  self.config_info,
+                                                  message['sender_full_name'],
+                                                  self.supported_languages)
+        bot_handler.send_reply(message, bot_response)
+
+api_url = 'https://translation.googleapis.com/language/translate/v2'
+
+help_text = '''
+Google translate bot
+Please format your message like:
+`@-mention "<text_to_translate>" <target-language> <source-language(optional)>`
+Visit [here](https://cloud.google.com/translate/docs/languages) for all languages
+'''
+
+language_not_found_text = '{} language not found. Visit [here](https://cloud.google.com/translate/docs/languages) for all languages'
+
+def get_supported_languages(key):
+    parameters = {'key': key, 'target': 'en'}
+    response = requests.get(api_url + '/languages', params = parameters)
+    if response.status_code == requests.codes.ok:
+        languages = response.json()['data']['languages']
+        return {lang['name'].lower(): lang['language'].lower() for lang in languages}
+    raise TranslateError(response.json()['error']['message'])
+
+class TranslateError(Exception):
+    pass
+
+def translate(text_to_translate, key, dest, src):
+    parameters = {'q': text_to_translate, 'target': dest, 'key': key}
+    if src != '':
+        parameters.update({'source': src})
+    response = requests.post(api_url, params=parameters)
+    if response.status_code == requests.codes.ok:
+        return response.json()['data']['translations'][0]['translatedText']
+    raise TranslateError(response.json()['error']['message'])
+
+def get_code_for_language(language, all_languages):
+    if language.lower() not in all_languages.values():
+        if language.lower() not in all_languages.keys():
+            return ''
+        language = all_languages[language.lower()]
+    return language
+
+def get_translate_bot_response(message_content, config_file, author, all_languages):
+    message_content = message_content.strip()
+    if message_content == 'help' or message_content is None or not message_content.startswith('"'):
+        return help_text
+    split_text = message_content.rsplit('" ', 1)
+    if len(split_text) == 1:
+        return help_text
+    split_text += split_text.pop(1).split(' ')
+    if len(split_text) == 2:
+        # There is no source language
+        split_text.append("")
+    if len(split_text) != 3:
+        return help_text
+    (text_to_translate, target_language, source_language) = split_text
+    text_to_translate = text_to_translate[1:]
+    target_language = get_code_for_language(target_language, all_languages)
+    if target_language == '':
+        return language_not_found_text.format("Target")
+    if source_language != '':
+        source_language = get_code_for_language(source_language, all_languages)
+        if source_language == '':
+            return language_not_found_text.format("Source")
+    try:
+        translated_text = translate(text_to_translate, config_file['key'], target_language, source_language)
+    except requests.exceptions.ConnectionError as conn_err:
+        return "Could not connect to Google Translate. {}.".format(conn_err)
+    except TranslateError as tr_err:
+        return "Translate Error. {}.".format(tr_err)
+    except Exception as err:
+        return "Error. {}.".format(err)
+    return "{} (from {})".format(translated_text, author)
+
+handler_class = GoogleTranslateHandler
--- a/zulip_bots/zulip_bots/bots/googletranslate/test_googletranslate.py
+++ b/zulip_bots/zulip_bots/bots/googletranslate/test_googletranslate.py
@@ -1,0 +1,135 @@
+#!/usr/bin/env python
+
+from __future__ import absolute_import
+from __future__ import print_function
+
+import json
+
+from unittest.mock import patch
+from requests.exceptions import HTTPError, ConnectionError
+
+from zulip_bots.test_lib import BotTestCase
+from zulip_bots.bots.googletranslate.googletranslate import TranslateError
+
+help_text = '''
+Google translate bot
+Please format your message like:
+`@-mention "<text_to_translate>" <target-language> <source-language(optional)>`
+Visit [here](https://cloud.google.com/translate/docs/languages) for all languages
+'''
+
+class TestGoogleTranslateBot(BotTestCase):
+    bot_name = "googletranslate"
+
+    def test_normal(self):
+        with self.mock_config_info({'key': 'abcdefg'}), \
+                self.mock_http_conversation('test_normal'):
+                    with self.mock_http_conversation('test_languages'):
+                        self.initialize_bot()
+                    self.assert_bot_response(
+                        message = {'content': '"hello" de', 'sender_full_name': 'tester'},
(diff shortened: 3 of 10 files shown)


zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
```diff
@@ -0,0 +1,92 @@
...
+    raise TranslateError(r.json()['error']['message'])
+
+def get_translate_bot_response(text, config_file, author):
+    text = text.strip()
+    if (text == 'languages'):
+        return languages_text
+    if (text == 'help' or text is None or not text.startswith('"')):
+        # Return the help message
+        return help_text
+    split_text = []
```
neiljp: You overwrite this assignment on the line below?

zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
```diff
@@ -0,0 +1,92 @@
...
+    if (len(split_text) == 1):
+        return help_text
+    split_text += split_text.pop(1).split(' ')
+    if (len(split_text) == 2):
+        # There is no source language
+        split_text.append("")
+    if (len(split_text) != 3):
+        return help_text
+    split_text[0] = split_text[0].replace('"', '')
+    if (not split_text[1] in utils.google_langs.keys()):
```
neiljp: If you want to refactor `utils.py` then you could consider making this a short function, to allow the `google_langs` to be varied.

zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
```diff
@@ -0,0 +1,92 @@
...
+        # This means this is not a language code, change it to the code
+        try:
+            split_text[1] = utils.get_code(split_text[1])
+        except Exception as e:
+            return "Language not found. Visit [here](https://cloud.google.com/translate/docs/languages) for all languages"
+    try:
+        translated = translate(split_text[0], config_file['key'], split_text[1], split_text[2])
```
neiljp: It is very obvious what `config_file['key']` means; I can understand what's happening with the other arguments, but you might consider adjusting the call to use more explicit variable names, eg. setting `text_to_translate` on line 75, and perhaps similarly for the other variables.

fredfishgames:
Hello
I think I've made all the changes you requested, however, instead of using `utils.py` I'm now getting the list of languages from the API itself.

zulip_bots/zulip_bots/bots/googletranslate/test_googletranslate.py
```diff
@@ -0,0 +1,50 @@
...
+from __future__ import print_function
+
+import json
+
+from unittest.mock import patch
+from requests.exceptions import HTTPError, ConnectionError
+
+from zulip_bots.test_lib import BotTestCase
+
+from zulip_bots.bots.googletranslate.googletranslate import TranslateError
```
neiljp: This is unused; you don't have a test for the `TranslateError` case?
Maybe add a general `Exception` case too, since you handle that in the bot?

zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
```diff
@@ -0,0 +1,103 @@
...
+`@-mention "<text_to_translate>" <target-language> <source-language(optional)>`
+To see a list of languages, @-mention this bot with `languages`
+'''
+
+languages_text_to_translate = '''Please see [here](https://cloud.google.com/translate/docs/languages) for all languages and their codes.
+You can use these codes, or their names.'''
+
+def get_supported_languages(key):
+    parameters = {'key': key, 'target': 'en'}
+    r = requests.get(api_url + '/languages', params = parameters)
```
neiljp: While `r` is an obvious abbreviation if you know what the function does, perhaps choose a more descriptive name for those who initially might not. Good names also indicate to those reading the code later what it is intended to do, which can be useful not just for understanding, but checking that the intent matches what is actually happening :)

zulip_bots/zulip_bots/bots/googletranslate/googletranslate.conf
```diff
@@ -0,0 +1,2 @@
+[googletranslate]
+key=your_api_key_here
```
aero31aero: Add a newline at the end.

zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
```diff
@@ -0,0 +1,103 @@
...
+        self.supported_languages = get_supported_languages(self.config_info['key'])
+
+    def handle_message(self, message, bot_handler):
+        bot_response = get_translate_bot_response(message['content'],
+                                                  self.config_info,
+                                                  message['sender_full_name'],
+                                                  self.supported_languages)
+        bot_handler.send_reply(message, bot_response)
+
+api_url = 'https://translation.googleapis.com/language/translate/v2'
```
aero31aero: I am a little vary of having all the bot code outside of the bot class. I would recommend creating a TranslateHelper class or similar for having this code instead. @roberthoenig what are your views on this?

zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
```diff
@@ -0,0 +1,103 @@
...
+    parameters = {'key': key, 'target': 'en'}
+    r = requests.get(api_url + '/languages', params = parameters)
+    languages = r.json()['data']['languages']
+    language_dictionary = {}
+    for lang in languages:
+        language_dictionary.update({lang['name'].lower(): lang['language']})
+    return language_dictionary
+
+class TranslateError(Exception):
+    def __init__(self, value):
```
neiljp: I believe you could simplify the body of the class to just `pass` - give it a try :)

aero31aero:
Hey! Getting the list of languages from the API sounds very neat. Also, from a static reading only, the PR looks good. I was unable to create an API key so I haven't done the manual testing yet.

It would be great if you could run the bot on chat.zulip.org and drop a message in #integrations stream regarding testing it. :smiley:

zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
neiljp: @aero31aero I'm not sure this is necessary, at least based on previous bots, though I would be interested to hear more thoughts. Ultimately the `Handler` class just needs the appropriate functions to be present, to be called from the library, so the implementation itself can go anywhere else that those functions can call?

fredfishgames:
Also I've added a few more tests

zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
```diff
@@ -0,0 +1,104 @@
...
+    if language.lower() not in all_languages.values():
+        if language.lower() not in all_languages.keys():
+            return ''
+        language = all_languages[language.lower()]
+    return language
+
+def get_translate_bot_response(message_content, config_file, author, all_languages):
+    message_content = message_content.strip()
+    if message_content == 'help' or message_content is None or not message_content.startswith('"'):
+        # Return the help message
```
neiljp: You follow this comment by `return help_txt`, which makes it pretty self-explanatory without the comment :)

zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
```diff
@@ -0,0 +1,104 @@
...
+        return help_text
+    (text_to_translate, target_language, source_language) = split_text
+    text_to_translate = text_to_translate[1:]
+    target_language = check_language(target_language, all_languages)
+    if target_language == '':
+        return language_not_found_text
+    if source_language != '':
+        source_language = check_language(source_language, all_languages)
+        if source_language == '':
+            return language_not_found_text
```
neiljp: If the user provides source and target languages and neither are supported, how do they know which one is "not found"?

You should test for both cases too.

zulip_bots/zulip_bots/bots/googletranslate/test_googletranslate.py
```diff
@@ -0,0 +1,72 @@
...
+                self.mock_http_conversation('test_403'):
+                    with self.mock_http_conversation('test_languages'):
+                        self.initialize_bot()
+                    self.assert_bot_response(
+                        message = {'content': '"hello" german english', 'sender_full_name': 'tester'},
+                        response = {'content': 'Translate Error. Invalid API Key..'},
+                        expected_method = 'send_reply'
+                    )
+
+    def test_help(self):
```
neiljp: Re the name of this test, this is really just testing getting help text in response to an empty string - there are many cases which also return the help text. These tests would be great to add, whether in this test function (so `test_help` remains descriptive), or in individual tests (in which case renaming this test may be appropriate).

zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
```diff
@@ -0,0 +1,104 @@
...
+    try:
+        translated_text = translate(text_to_translate, config_file['key'], target_language, source_language)
```
neiljp: This is so much clearer to read, and I hope you agree makes the rest of the function easier to understand too?

zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
```diff
@@ -0,0 +1,104 @@
...
+'''
+
+language_not_found_text = 'Language not found. Visit [here](https://cloud.google.com/translate/docs/languages) for all languages'
+
+def get_supported_languages(key):
+    parameters = {'key': key, 'target': 'en'}
+    response = requests.get(api_url + '/languages', params = parameters)
+    if response.status_code == requests.codes.ok:
+        languages = response.json()['data']['languages']
+        return {lang['name'].lower(): lang['language'].lower() for lang in languages}
```
neiljp: This looks better to me, but do you find this change clearer too?

zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
```diff
@@ -0,0 +1,104 @@
...
+def translate(text_to_translate, key, dest, src):
+    parameters = {'q': text_to_translate, 'target': dest, 'key': key}
+    if src != '':
+        parameters.update({'source': src})
+    response = requests.post(api_url, params=parameters)
+    if response.status_code == requests.codes.ok:
+        return response.json()['data']['translations'][0]['translatedText']
+    raise TranslateError(response.json()['error']['message'])
+
+def check_language(language, all_languages):
```
neiljp: Perhaps this function would be more self-explanatory with a different name? It does check the language, but its a simple function so you could even include the returned data in the name, ie. along the lines of `get_[returnedthing]_for_[importantinput]` (or using 'for' instead of 'from' - or something else along these lines)

Then using it further down, you could even consider (though it is not necessary to do so) renaming variables to take into account that the 'returnedthing' means something different to the 'importantinput'.

zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
```diff
@@ -0,0 +1,104 @@
...
+    target_language = check_language(target_language, all_languages)
+    if target_language == '':
+        return language_not_found_text
+    if source_language != '':
+        source_language = check_language(source_language, all_languages)
+        if source_language == '':
+            return language_not_found_text
+    try:
+        translated_text = translate(text_to_translate, config_file['key'], target_language, source_language)
+    except requests.exceptions.ConnectionError as conn_err:
```
neiljp: Did you intend to use `conn_err`?

zulip_bots/zulip_bots/bots/googletranslate/googletranslate.py
fredfishgames: Yeah, Just didn't know how to do it before

fredfishgames:
Hello
I made all the changes you suggested, and I agree that it is now easier and clearer to read
Thanks

aero31aero:
@fredfishgames can you squash the commits now into one? This looks good to merge to me. :tada:

neiljp:
I agree with @aero31aero, this looks ready! :D

fredfishgames:
Hi
I've squashed all the commits into one

aero31aero:
@showell this LGTM for merging. Requesting a final review. Thanks!

showell:
merged!  Thanks @fredfishgames @neiljp and @aero31aero!

tommyip:
GCI task approved.

Issue #161, opened 2017-11-20
Migrate Google calendar bot.


derAnfaenger:
https://github.com/zulip/zulip/pull/2771 is a PR for adding a Google Calendar bot.
The PR is big, and has potentially the most comments any PR in the Zulip project ever had.

We should:
* Copy the PR's code and create a new Google Calendar bot in this repo.
* Test the bot. If it works, we should just merge it. Then, people can collaborate on code-style issues, adding tests, etc.

Ivche1337:
@zulipbot claim

zulipbot:
Welcome to Zulip, @Ivche1337! We just sent you an invite to collaborate on this repository at https://github.com/zulip/python-zulip-api/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

Ivche1337:
https://github.com/zulip/python-zulip-api/pull/198

PR #198, opened 2017-12-04
Added google calendar bot.

Ivche1337:
I did this https://github.com/zulip/python-zulip-api/issues/161

### DIFF
--- a/zulip/integrations/google/oauth.py
+++ b/zulip/integrations/google/oauth.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python
+from __future__ import print_function
+import argparse
+import httplib2
+import os
+
+# Uses Google's Client Library
+#   pip install --upgrade google-api-python-client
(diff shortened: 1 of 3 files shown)


aero31aero:
Hey @Ivche1337 thanks for working on this! There is still a lot to do before we can merge this PR. What we need to ensure that the bot is working is tests as we have for other bots. Take a look at the giphy bot as an example for testing bots that send requests to a remote server.

In order to write those tests, and in general making sure that the bot runs, its required to generate API keys for Google Calendar and get the bot up and running. That way, we can also capture sample payloads(requests) to the API and start mocking them in our tests.

You can run the tests locally by `tools/test-bots`.

When you make further commits to this PR, please make new commits instead of amending this existing commit. That way, it would be easier to review the subsequent changes only.

Please join us on https://chat.zulip.org/ if you need help in writing the tests. Great work so far. :smiley:

Issue #174, opened 2017-11-27
followup: Allow changing default stream.


aero31aero:
Currently, the followup bot expects to have a stream `#followup` with write access. This parameter should be modifiable via either a config file like in github_detail bot or some other method.

Pseudo code for the logic:

```
try:
    load config
    defaultsttream = config.stream
except:
    defaultstream = "followup"
```

aero31aero:
@zulipbot label "bite size"

nkmishra1997:
@zulipbot claim

zulipbot:
Welcome to Zulip, @nkmishra1997! We just sent you an invite to collaborate on this repository at https://github.com/zulip/python-zulip-api/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

nkmishra1997:
I tried to follow what has been done in github_detail bot.
I created a stream_detail.config file and in the followup.py, I tried something like this:
```
def initialize(self, bot_handler):
    try:
        self.config_info = bot_handler.get_config_info('stream_detail', optional=True)
        self.default_stream = self.config_info.get("stream", False)
    except:
        self.default_stream = 'followup
```

But, in the `handle_message` function, when i did `default_stream = self.default_stream`, the tests failed with error `AttributeError: 'FollowupHandler' object has no attribute 'default_stream'`

So I decided to load config inside the `handle_message` function:
```
try:
    self.config_info = bot_handler.get_config_info('stream_detail', optional=True)
    stream_name = self.config_info.get("stream", False)
except:
    stream_name = 'followup'

bot_handler.send_message(dict(
    type='stream',
    to=stream_name,
    subject=message['sender_email'],
    content=bot_response,
))
```
How shall I finally go about it?

nkmishra1997:
[...]

Issue #173, opened 2017-11-27
followup: Provide help text as a reply.


aero31aero:
Unlike most other bots, followup bot doesn't supply a help text on running `@botname help`.

On sending `help`, we should respond only with the help text of the bot stop the control flow there.

aero31aero:
@zulipbot label "bite size"

nkmishra1997:
@zulipbot claim

nkmishra1997:
The current Help Text is

> This plugin will allow users to flag messages as being follow-up items.  Users should preface messages with "@mention-bot". Before running this, make sure to create a stream called "followup" that your API user can send to.

I think we can update this as well once the #200 is merged:

> Before running this, make sure to create a stream called "followup" that your API user can send to OR you can specify a custom stream in the stream_detail.config file.

aero31aero:
Sounds good to me. :)

PR #200, opened 2017-12-05
 followup: Allow changing default stream.

nkmishra1997:
#174 
I have created a stream_detail.config file which allows changing the default "followup" stream. I have updated the test_followup.py for the same.


zulip_bots/zulip_bots/bots/followup/followup.py
```diff
@@ -22,15 +23,21 @@ def usage(self):
             called "followup" that your API user can send to.
             '''
 
+    def initialize(self, bot_handler):
+            self.config_info = bot_handler.get_config_info('stream_detail', optional=True)
+            self.default_stream = self.config_info.get("stream", False)
+            self.default_stream = 'followup'
```
aero31aero: Hey. It seems like this line would nullify the previous line. Nice call deciding that followup should be the default stream.

You should rewrite this as `self.default_stream = self.config_info.get("stream", "folllowup")` as the second argument to `get` is the default value to return in case we do not have the specified key in config.

zulip_bots/zulip_bots/bots/followup/stream_detail.config
```diff
@@ -0,0 +1,2 @@
+[stream_detail]
+stream = <stream name>
```
aero31aero: Can you add a newline at the end of this file?

zulip_bots/zulip_bots/bots/followup/test_followup.py
```diff
@@ -26,4 +27,6 @@ def test_bot(self):
               'subject': 'foo_sender@zulip.com',
               'content': 'from foo_sender@zulip.com: I have completed my task'}),
         ]
-        self.check_expected_responses(expected_send_message, expected_method='send_message')
+        with self.mock_config_info(self.mock_config):
+            self.initialize_bot()
+            self.check_expected_responses(expected_send_message, expected_method='send_message')
```
aero31aero: Can you add a test case where we have a different stream than default? That way we can be certain that the code works as expected. For example, right now, the code wasn't working but the test was passing because the testcase was only testing the default stream name.

nkmishra1997:
@aero31aero I have made the changes and created two tests: 1. to test the default stream 2. to test the specified stream.
The tests are running successfully.

zulip_bots/zulip_bots/bots/followup/followup.py
```diff
@@ -1,5 +1,6 @@
 # See readme.md for instructions on running this code.
 
+import six.moves.configparser
```
showell: We no longer need to support python2, so no need to use "six" here.

zulip_bots/zulip_bots/bots/followup/followup.py
```diff
@@ -22,15 +23,20 @@ def usage(self):
             called "followup" that your API user can send to.
             '''
 
+    def initialize(self, bot_handler):
+            self.config_info = bot_handler.get_config_info('stream_detail', optional=True)
+            self.default_stream = self.config_info.get("stream", "folllowup")
```
showell: There's a typo here (three l's).

zulip_bots/zulip_bots/bots/followup/test_followup.py
```diff
@@ -7,8 +7,10 @@
 
 class TestFollowUpBot(BotTestCase):
     bot_name = "followup"
+    mock_config_default_stream = {'stream': 'followup'}
```
showell: I'm not sure having these as class variables is necessary.  You can probably define them directly in the tests.  And it seems like "followup" **is** the default, so you should be mocking config_info to return None or something, right?

nkmishra1997:
I have made the changes as mentioned by @showell  and I have also tried to solve #173 .

I am facing some issue with writing tests. I reverted back everything I did in `test_followup.py` and now when I run tests, I am getting this error. 

```
======================================================================
ERROR: test_bot (zulip_bots.bots.followup.test_followup.TestFollowUpBot)
  File "/home/nikhil/ossTRY/python-zulip-api/zulip_bots/zulip_bots/bots/followup/test_followup.py", line 15, in test_bot
    self.check_expected_responses(expected_send_reply, expected_method='send_reply')
  File "/home/nikhil/ossTRY/python-zulip-api/zulip_bots/zulip_bots/test_lib.py", line 292, in check_expected_responses
    self.assert_bot_responses(message, (response, expected_method))
  File "/home/nikhil/ossTRY/python-zulip-api/zulip_bots/zulip_bots/test_lib.py", line 344, in assert_bot_responses
    self.call_request(message, *response_list)
  File "/home/nikhil/ossTRY/python-zulip-api/zulip_bots/zulip_bots/test_lib.py", line 302, in call_request
    self.message_handler.handle_message(message, self.mock_bot_handler)
TypeError: handle_message() missing 1 required positional argument: 'config_info'

```

I tried to do what @showell did in #206 but I am not able to proceed. Help me out with this. I have understood that I can now directly use `self.mock_config_info({'stream': 'any stream here'})` to mock the stream and test it. I need not declare them as class variables. But, I was unable to solve the above-mentioned error.

PR #201, opened 2017-12-06
Mypy annotations for followup and giphy

thejwzhang:


roberthoenig:
Nice work @thejwzhang! These changes, however, won't be tested automatically unless you add all the files you've just annotated to the `force_include` list in `tools/run-mypy`. Can you do that and see if everything still works?

eeshangarg:
@thejwzhang: Thanks for working on this! I have a few comments that I'd like you to address:
* Try squashing the commits for each bot, so the changes for the followup bot should be in one commit called `bots/followup: Use Python 3 type mypy annotations.` and so on for the giphy bot too.
* Travis is failing because you have to specify the full path to the individual files in `force_include`, not just the path to the directory, so I'd like you to fix that as well and then run `tools/run-mypy` to make sure it runs smoothly before submitting the PR for review again. :)

Apart from that, this looks great! If you have any questions, feel free to post on `chat.zulip.org`! :) Thanks!

zulip_bots/zulip_bots/bots/giphy/giphy.py
```diff
@@ -43,16 +44,17 @@ class GiphyNoResultException(Exception):
...
     query = {'s': keyword,
              'api_key': api_key}
     try:
         data = requests.get(GIPHY_TRANSLATE_API, params=query)
-    except requests.exceptions.ConnectionError as e:  # Usually triggered by bad connection.
-        logging.warning(e)
+    except requests.exceptions.ConnectionError as e:
+        # Usually triggered by bad connection.
+        logging.warning('{}'.format(e))
         raise
```
eeshangarg: Actually, I think it might be better to do this (this just occured to me):
```
logging.exception('Bad connection')
raise
```

See [this documentation](https://docs.python.org/3/library/logging.html#logging.Logger.exception), the exception info should be added to the message, if I'm not wrong.

eeshangarg:
@thejwzhang: This looks great! I left a comment. I would like you to address that and then we can probably merge this! I would recommend rebasing and editing the `giphy` commit to make the change I suggested. It would look something like this:

```
git rebase -i HEAD~2
```
Replace `pick` with `edit` for the `giphy` commit. Make your changes, `git add` the file you changed, and then run:
```
git commit --amend
git rebase --continue
```
And then push the changes (add `--force` to the push command). Also, `git commit --amend` will ask you if you want to change the commit message, don't change it, just leave it as it is. :) Let me know how it goes :)

eeshangarg:
@roberthoenig: Did you have any thoughts on this btw? :)

eeshangarg:
@thejwzhang: Nice work! :) I am going to go ahead and merge this!

PR #204, opened 2017-12-06
zulip_bots: Create Google Calendar Bot.

YagoGG:
This addresses #161, as discussed on Zulip with @Ivche1337.

Adapting the tests and the extra fixes from the original PR will be made by him.


YagoGG:
Updated the PR with some fixes and minor improvements.

I haven't added a unit tests because the bot needs a decent amount of setup in order to run, which cannot be prepared with the current testing framework. If we were to make quite a decent amount of mocking in order to test it, I could offer to make a followup PR.

Cheers!

roberthoenig:
Nonetheless, I feel this PR has been stuck for too long and should just get merged and then iterated upon.

However, in light of especially the last bullet point, I wouldn't recommend directly adding this bot to the main directory. Instead, I'd add a commit to move the google_calendar bot to `bots_unmaintained` and then merge it. Then, we can close this PR, improve the bot, and eventually move it to the proper bots directory.

YagoGG:
Thanks for reviewing, @roberthoenig!

I don't mind waiting a bit more in order to have this merged properly. I'll update the PR in a few hours :)

Regarding the parser tests, I forgot to make some comments about that: I tried to export that to a separate testing file, but I noticed the following:
- As soon as you create the test file, the 3 standard checks that are built into the testing system are run (for instance, sending the bot an empty message). This causes some problems, since it involves starting the server —which can't be done due to the complex setup I mentioned earlier. A possible fix would be overriding those predefined tests.
- Moving the stuff in `test()` to a separate file would require importing the bot's code from the new test file. Yesterday, when I played around with that idea, I had to do the trick of adding the bot's directory to the interpreter's path. Not sure if this would end up being necessary after some iterations, but just something worth keeping in mind.

YagoGG:
I addressed for now the first two comments @roberthoenig made.

Let me know what you think about the tests and if you consider that further improvements are needed.

Thanks!

snowe2010:
Hi, what is blocking this PR from being merged?

timabbott:
It would benefit from being rebased and tested by a new person, just to make sure it still works.

PIG208:
I think this is worth picking up. We might need to do some extra testings on this to see if everything still works fine.

PR #208, opened 2017-12-07
Added comments, added more descriptive responses, added usage....

Ivche1337:
There wasn't much that could be done to this bot, i made sure everything is running properly, reorganized the code so it looks a bit better and added some documentation.
There is no need for a configuration file since there are no API keys or anything from that nature included.
Also i didn't have any idea if there is some feature than i can implement. 
Let me know if i need to do something else :D

### DIFF
--- a/zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
+++ b/zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
@@ -75,10 +75,10 @@ class WikipediaHandler(object):
                 url1 = 'https://en.wikipedia.org/wiki/' + search_string1
                 url2 = 'https://en.wikipedia.org/wiki/' + search_string2
                 url3 = 'https://en.wikipedia.org/wiki/' + search_string3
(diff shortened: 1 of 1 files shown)


aero31aero:
Hey, thanks for working on this! I see a couple of easy targets here. It would be useful to print multiple results for a query, formatting them as a list of links like:

```
Here are the results:
1. [Result One](https://en.wikipedia.org/one)
2. [Result Two](https://en.wikipedia.org/two)
3. [Result Three](https://en.wikipedia.org/three)
```

What do you feel about this?

aero31aero:
Also, the tests fail on the PR as of now. See https://travis-ci.org/zulip/python-zulip-api/jobs/313139932 for details.

neiljp:
@Ivche1337 I suggest you rename the PR to indicate the area/bot you're working on, as that isn't very clear :)

showell:
I'm closing this, as I think we merged the gist of this PR on another PR.

PR #209, opened 2017-12-07

fredfishgames:
This adds mypy annotations for:
github_detail
googlesearch
help


showell:
merged, thanks @fredfishgames 

FYI I think we don't really need `self: Any` in most places, but we can fix that later.

aero31aero:
Approved GCi task.

PR #210, opened 2017-12-08

KaiUsesThis:
Adds annotations for the following bots: 
- converter
- define
- encrypt

Accidentally added `.vscode` in gitignore, can remove if needed.


zulip_bots/zulip_bots/bots/converter/converter.py
```diff
@@ -10,7 +10,11 @@
 import re
 from zulip_bots.bots.converter import utils
 
-def is_float(value):
+from typing import Any, List
+
+from zerver.lib.type_debug import print_types
```
showell: Do we need this import any more?

KaiUsesThis: Whoops, my bad. I've believe corrected it now.

aero31aero:
Hey! Thanks for working on this. It looks like you have added 4 extra commits on this PR. Can you do an interactive rebase and remove all commits except for the three where you've added the annotations?

zulip_bots/zulip_bots/bots/converter/converter.py
```diff
@@ -10,7 +10,9 @@
 import re
 from zulip_bots.bots.converter import utils
 
-def is_float(value):
+from typing import Any
+
+def is_float(value: Any) -> bool:
```
neiljp: I believe this would be more appropriate as `object`? Do other reviewers agree?

zulip_bots/zulip_bots/bots/converter/converter.py
```diff
@@ -46,11 +49,11 @@ def usage(self):
                all supported units.
                '''
 
-    def handle_message(self, message, bot_handler):
+    def handle_message(self, message: Any, bot_handler: Any) -> None:
```
neiljp: You should be able to tell from the way in which `message` is used, that it can be typed more strongly than this (particularly in how the same passed-through variable is used below), and also from `lib.py`.

tools/run-mypy
```diff
@@ -37,7 +37,10 @@ force_include = [
     # Include bots that we migrate to mypy.
     "zulip_bots/zulip_bots/bots/helloworld/helloworld.py",
     "zulip_bots/zulip_bots/bots/converter/converter.py",
-    "zulip_bots/zulip_bots/bots/converter/test_converter.py"
+    "zulip_bots/zulip_bots/bots/converter/test_converter.py",
+    "zulip_bots/zulip_bots/bots/define/define.py",
+    "zulip_bots/zulip_bots/bots/define/test_define.py"
+
```
neiljp: We don't need this extra line.

zulip_bots/zulip_bots/bots/define/define.py
```diff
@@ -17,19 +19,19 @@ class DefineHandler(object):
...
 
-    def usage(self):
+    def usage(self) -> str:
         return '''
             This plugin will allow users to define a word. Users should preface
             messages with @mention-bot.
             '''
 
-    def handle_message(self, message, bot_handler):
+    def handle_message(self, message: Any, bot_handler: Any) -> None:
```
neiljp: See my review of converter bot typing - `message` should be able to be typed more specifically than this.

zulip_bots/zulip_bots/bots/define/define.py
```diff
@@ -17,19 +19,19 @@ class DefineHandler(object):
...
         original_content = message['content'].strip()
         bot_response = self.get_bot_define_response(original_content)
 
         bot_handler.send_reply(message, bot_response)
 
-    def get_bot_define_response(self, original_content):
+    def get_bot_define_response(self, original_content: Any) -> str:
```
neiljp: Can you see how `original_content` could be typed more specifically, based on how it's used in the function, or the values that are passed in?

KaiUsesThis:
@neiljp Thanks for the feedback! I changed the commits to include your suggestions.

PR #214, opened 2017-12-08
interactive bots: Create lunch bot.

LiTO773:
This is a simple lunch organizer bot, it can create a list of items of what each user wants to eat.

### DIFF
--- a/zulip_bots/zulip_bots/bots/lunch/doc.md
+++ b/zulip_bots/zulip_bots/bots/lunch/doc.md
@@ -1,0 +1,34 @@
+# Lunch Bot 🍛
+A small Zulip bot for organizing lunch with your friends and coworkers. To call
+the bot simply use:
+
+```
+@lunch <user if applicable> <command>
+```
+
+## Usage
+Most of lunch bot's commands can be applied to both the sender or another user
+in the same server. To do so, write the user's @ before the command like in the
+following examples:
+
+`I want <item>` - Adds <item> to the sender's list of items
+
+`<user> wants <item>` - Adds <item> to the <user>'s list of items
+
+`<user> I want that` - Adds every item in <user>'s list of items to the sender's
+list of items
+
+`<user> remove <item>` - Removes <item> from the sender's (or <user> if
+specified) list of items
+
+`<user remove all` - Removes all items from the sender's (or <user> if specified
+) list of items
+
+`orders` - Show all orders
+
+`<user> order` - Shows all items in the sender's (or <user> if specified) list
+of items
+
+`who should <verb>(?)` - Randomly selects a user from the orders list
+
+`help` - Shows help (this section)
--- a/zulip_bots/zulip_bots/bots/lunch/lunch.py
+++ b/zulip_bots/zulip_bots/bots/lunch/lunch.py
@@ -1,0 +1,185 @@
+import random
+
+ORDERS = {}
+
+class LunchBotHandler(object):
+
+    def usage(self):
+        return '''
+        This is a small Zulip bot for organizing lunch with your friends and
+        coworkers.
+        '''
+
+    def handle_message(self, message, bot_handler):
+        content = self.keyword_selection(message)
+        bot_handler.send_reply(message, content)
+
+    def keyword_selection(self, message):
+        msg = message['content'].split()
+
+        if len(msg) == 0:
+            return "I don't think I can do that. :frowning:"
+
+        user = message['sender_full_name']
+        if msg[0][0] == '@':
+            user = msg[0][1:].replace('**', '')
+            msg  = msg[1:]
+
+        if len(msg) == 0:
+            return
+
+        if msg[-1] == 'that':
+            return copy_user_list(user, message['sender_full_name'])
+
+        if msg[0:2] == ['I', 'want']:
+            return add_to_user_list(user, msg[2:])
+
+        if msg[0] == 'wants':
+            return add_to_user_list(user, msg[1:])
+
+        if msg == ['remove', 'all']:
+            return remove_all(user)
+
+        if msg[0] == 'remove':
+            return remove_from_user_list(user, msg[1:])
+
+        if msg[0] == 'orders':
+            return show_all_orders()
+
+        if msg[0] == 'order':
+            return show_order(user)
+
+        if msg[0:2] == ['who', 'should']:
+            return random_user(msg[2:])
+
+        if msg[0] == 'help':
+            return '''
+                <user|optional> (I) want(s) <item>
+                Adds <item> to the sender's (or <user>'s if specified) list of items
+
+                <user> I want that
+                Adds every item in <user>'s list of items to the sender's list of items
+
+                <user|optional> remove <item>
+                Removes <item> to the sender's (or <user>'s if specified) list of items
+
+                <user|optional> remove all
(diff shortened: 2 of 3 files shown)


LiTO773:
I'm having a problem with bot testing.
Is there a way to specify the 'sender_full_name' in the test file? I'm not finding it in the docs

zulip_bots/zulip_bots/bots/lunch/test_lunch.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python
+from zulip_bots.test_lib import StubBotTestCase
+
+class TestLunch(StubBotTestCase):
+    bot_name = 'lunch'
+
+    def test_bot(self):
+        msg = dict(
```
showell: I know you copied this style from another test, but it's not clear to me that some of the items in this dictionary make the test more readable.  I think it's better if we can read the tests and clearly see how the food items get added and removed.

zulip_bots/zulip_bots/bots/lunch/lunch.py
```diff
@@ -0,0 +1,219 @@
...
+        return FIND_USER(user, True)
+
+    # Removes user
+    del ORDERS[user]
+
+    return 'Cleared @**' + user + '** list! :ok_hand:'
+
+# Shows all orders
+def show_all_orders():
+    order = str(ORDERS)
```
showell: This feels like a strange way to build the list of orders.  You can do something more like `', '.join(ORDERS)` to produce the list.

LiTO773:
Fixed! :)

zulip_bots/zulip_bots/bots/lunch/lunch.py
```diff
@@ -0,0 +1,214 @@
...
+                item_str += item
+            else:
+                item_str += ' ' + item
+
+    return item_str[1:].split(',')
+
+# Copies user list
+def copy_user_list(original_user, user):
+    # Verification
+    if FIND_USER(original_user, True):
```
showell: We should change the name of `FIND_USER` to be just `find_user` (lowercase).  More importantly, we want to figure out a nicer protocol for error handling, where you don't have to call back to the same function to get the error message.  We can have `find_user` return a tuple like `(True, ['ice cream'])` or `(False, 'Could not find Alice').  Or you may want to consider having a separate helper function to format error messages.

LiTO773:
Changed `FIND_USER` to `find_user`,
Function is now only called once,
Returns a tuple

zulip_bots/zulip_bots/bots/lunch/doc.md
```diff
@@ -0,0 +1,34 @@
...
+Show all orders
+
+`<user|optional> order`
+Shows all items of the sender's (or <user>'s if specified) list of items
+
+`who should <verb>(?)`
+Randomly selects a user from the orders list
+
+`help`
+Shows help (this section)
```
eeshangarg: Could you please modify the `doc.md` to follow a structure similar to [the xkcd bot's docs](https://github.com/zulip/python-zulip-api/blob/master/zulip_bots/zulip_bots/bots/xkcd/doc.md)? We're hoping that's the structure we can follow moving further! :)

zulip_bots/zulip_bots/bots/lunch/lunch.py
```diff
@@ -0,0 +1,220 @@
...
+        return '''
+        This is a small Zulip bot for organizing lunch with your friends and
+        coworkers.
+        '''
+
+    # MESSAGE RECEIVERS
+    def handle_message(self, message, bot_handler):
+        content = self.keyword_selection(message)
+        print(message)
+        bot_handler.send_reply(message, content)
```
eeshangarg: Could you please explain what's the purpose of the `print()` here? Was this just for debugging and perhaps you forgot to remove it? :)

zulip_bots/zulip_bots/bots/lunch/lunch.py
```diff
@@ -0,0 +1,220 @@
...
+                Shows help (this section)
+            '''
+
+        return "I don't think I can do that. :frowning:"
+
+# User finder
+# Finds if a user exists
+# error is True, return an error message
+# error is False, create the user line
+def find_user(user, error):
```
eeshangarg: I don't think you need these comments! I think it would be better if you removed them! :)

zulip_bots/zulip_bots/bots/lunch/lunch.py
```diff
@@ -0,0 +1,220 @@
...
+    for item in items:
+        if len(item_str) == 0:
+            item_str += ' ' + item
+        else:
+            if item_str[-1] == ',':
+                item_str += item
+            else:
+                item_str += ' ' + item
+
+    return item_str[1:].split(',')
```
eeshangarg: I am bit a confused with this one, your comment says that you are `transforming an array of items into a string`, but you call `split` in `item_str` at the end, which returns a list, not a string. And that's not what the comment says. Am I correct? :) Either way, you can probably get rid of these comments too. :)

zulip_bots/zulip_bots/bots/lunch/lunch.py
```diff
@@ -0,0 +1,220 @@
...
+    # Checks if the ORDERS.keys() isn't empty
+    if len(users) == 0:
+        return "**No one wants to " + verb + "**. :anguished:"
+
+    # Selects the user
+    user = random.choice(users)
+
+    return "**I think @**" + user + "** should " + verb + ".** :stuck_out_tongue:"
+
+handler_class = LunchBotHandler
```
eeshangarg: I think this goes for all the functions for this module that most of the comments are redundant and describe what the code itself sufficiently indicates. I would recommend removing all of the comments! :)

eeshangarg:
@LiTO773: This looks good! :) I left some comments that I'd like you to address. They are pretty minor. :) I am not sure as to whether @roberthoenig and @showell have any more thoughts! :)

showell:
@eeshangarg Thanks for the thorough review!  @LiTO773 I would start by addressing Eeshan's comments, and then we'll go from there.  I think there are some general themes in his comments about things like string manipulation that might be good for you to ask about on chat.

zulip_bots/zulip_bots/bots/lunch/lunch.py
LiTO773: Ups, debugging :)

zulip_bots/zulip_bots/bots/lunch/lunch.py
LiTO773: The function works as expected, but it previously used to return a string (however I forgot to edit the comment :P). The comment has been removed now

zulip_bots/zulip_bots/bots/lunch/lunch.py
LiTO773: Removed, nice to know the code explains itself :)

showell:
@LiTO773 What is the current status of this PR?  Are you still making improvements or are you ready for a final review?

showell:
@LiTO773 Thanks for working on this!  I'm closing this for now, since it seems like you have moved on to other things, and we are actually trying to streamline how many bots we have in the repo.

I hope you had fun working on this.  The code is pretty clean.  Happy Python programming!

PR #215, opened 2017-12-08

fredfishgames:
Added annotations for
virtual_fs
weather


fredfishgames:
I've removed any `self: Any` annotations

showell:
Merged!  Thanks @fredfishgames!

PR #223, opened 2017-12-08
followup: Allow changing default stream, provide help text

nkmishra1997:
Solves #173 and #174 
I have created a new function `get_response()` in `test_lib.py` which basically returns the bot response to a given message.
I have created a config file, `followup.config` in which, the default config stream is 'followup' stream. It can be modified to some other stream as well.
I have also created different test cases for the same using `StubBotTestCase`


zulip_bots/zulip_bots/bots/followup/followup.py
```diff
@@ -23,15 +23,24 @@ def usage(self) -> str:
...
+
         if message['content'] == '':
             bot_response = "Please specify the message you want to send to followup stream after @mention-bot"
             bot_handler.send_reply(message, bot_response)
         else:
             bot_response = self.get_bot_followup_response(message)
             bot_handler.send_message(dict(
                 type='stream',
-                to='followup',
+                to= 'followup',
```
showell: We don't put a space here.  I agree with you it sometimes looks nicer, but it's our convention.

zulip_bots/zulip_bots/bots/followup/followup.py
```diff
@@ -23,15 +23,24 @@ def usage(self) -> str:
             called "followup" that your API user can send to.
             '''
 
+    def initialize(self, bot_handler: Any) -> None:
+        self.config_info = bot_handler.get_config_info('followup', optional=False)
+        self.stream = self.config_info.get("stream", False)
```
showell: Shouldn't the default here be 'followup', not `False`?

zulip_bots/zulip_bots/bots/followup/test_followup.py
```diff
@@ -3,29 +3,34 @@
...
+class TestFollowUpBot(StubBotTestCase):
     bot_name = "followup"
 
-    def test_bot(self) -> None:
-        expected_send_reply = [
-            ("", 'Please specify the message you want to send to followup stream after @mention-bot')
-        ]
-        self.check_expected_responses(expected_send_reply, expected_method='send_reply')
-
-        expected_send_message = [
```
showell: I think we want to keep the spirit of this test, which is that it handles the "normal" case.

zulip_bots/zulip_bots/bots/followup/test_followup.py
```diff
@@ -34,3 +34,25 @@ def test_help_text(self) -> None:
...
+
+    def test_followup_stream(self) -> None:
+        message = dict(
+            content='foo',
+            type='stream',
+            sender_email='foo@example.com',
+        )
+        with self.mock_config_info({'stream': 'followup'}):
+            response = self.get_response(message)
+            self.assertEqual(response['content'], 'from foo@example.com: foo')
```
showell: It's generally a good idea to limit what you put inside a `with` statement to only those operations that require the change in context. So, stylistically, I would put the `assertEqual` statements after the `with` block.

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -100,6 +100,17 @@ class StubBotTestCase(TestCase):
 
     bot_name = ''
 
+    def get_response(self, message):
```
showell: We probably want to split this out to be the first commit in your PR.  The commit message would be something `bot tests: Add get_response() helper.'

showell:
thanks for the progress, @nkmishra1997 !

I made a few comments that you'll want to address, and when you re-organize the commits, I would work on your commit messages.

Also, the tests failed due to a mypy error:

image

zulip_bots/zulip_bots/bots/followup/test_followup.py
nkmishra1997: I have handled this case in test_no_text()

zulip_bots/zulip_bots/bots/followup/test_followup.py
nkmishra1997: but if I remove the `response = self.get_response(message)` after the `with` block, I get 
`AttributeError: 'NoneType' object has no attribute 'get'` because I need to pass the config before I call `get_response()`

neiljp: @showell referred specifically to the `assertEqual` statements; the setting of the `response` variable can remain in the `with` block, but will still be accessible after the `with` is complete, so the testing of `response` can be done outside of the `with` block.

nkmishra1997: ohh I got it, thanks a lot.

zulip_bots/zulip_bots/bots/followup/followup.py
```diff
@@ -23,15 +23,24 @@ def usage(self) -> str:
...
 
+    def initialize(self, bot_handler: Any) -> None:
+        self.config_info = bot_handler.get_config_info('followup', optional=False)
+        self.stream = self.config_info.get("stream", 'followup')
+
     def handle_message(self, message: Dict[str, str], bot_handler: Any) -> None:
+        # Send help message
+        if message['content'] == 'help':
+            bot_handler.send_reply(message, self.usage())
+            return
```
neiljp: Just a minor thought, that this new block could be part of the if/else block below (elif), since it's part of the same tests; then you don't need the return either :)

nkmishra1997: Thanks for the review. I will move it to the if/else block itself. It makes more sense.

zulip_bots/zulip_bots/bots/followup/test_followup.py
showell: What about below?..."I have completed my task"

showell:
I am going to rebase this.

showell:
@nkmishra1997 Thanks for working on this!

One thing that would be helpful on future tasks is to try to keep your commits a bit more organized.  If you are solving a bunch of issues for a bot, it's good to have each commit address a single issue.  This helps the reviewers a lot, for example.  And it also allows us to merge pieces of your PR before conflicts occur.

A really good skill to grow is using interactive rebase.

Please review this to make sure I captured all of your changes:

https://github.com/zulip/python-zulip-api/commit/81b207795f09bf1bcea3bef2b9da394cfb714f07

nkmishra1997:
Thanks a lot, @showell . I will surely keep these things in mind. Looks like everything is covered in the merged commit. Sorry for the trouble.

showell:
Thanks for the followup @nkmishra1997!

PR #224, opened 2017-12-09

KaiUsesThis:
Adds annotations for the following bots:

- converter
- define
- encrypt

re-opened after closing last one.

### DIFF
--- a/tools/run-mypy
+++ b/tools/run-mypy
@@ -52,7 +52,12 @@ force_include = [
     "zulip_bots/zulip_bots/bots/weather/weather.py",
     "zulip_bots/zulip_bots/bots/youtube/youtube.py",
     "zulip_bots/zulip_bots/bots/youtube/test_youtube.py",
-
+    "zulip_bots/zulip_bots/bots/converter/converter.py",
+    "zulip_bots/zulip_bots/bots/converter/test_converter.py",
+    "zulip_bots/zulip_bots/bots/define/define.py",
+    "zulip_bots/zulip_bots/bots/define/test_define.py",
+    "zulip_bots/zulip_bots/bots/encrypt/encrypt.py",
+    "zulip_bots/zulip_bots/bots/encrypt/test_encrypt.py",
 ]
 
 parser = argparse.ArgumentParser(description="Run mypy on files tracked by git.")
--- a/zulip_bots/zulip_bots/bots/converter/converter.py
+++ b/zulip_bots/zulip_bots/bots/converter/converter.py
@@ -10,7 +10,9 @@ from math import log10, floor
 import re
 from zulip_bots.bots.converter import utils
 
-def is_float(value):
+from typing import Any, Dict, List
+
+def is_float(value: Any) -> bool:
     try:
         float(value)
         return True
@@ -22,7 +24,8 @@ def is_float(value):
 # fractional decimals, e.g. 0.00045 would become 0.0.
 # 'round_to()' rounds only the digits that are not 0.
 # 0.00045 would then become 0.0005.
-def round_to(x, digits):
+
+def round_to(x: float, digits: int) -> float:
     return round(x, digits-int(floor(log10(abs(x)))))
 
 class ConverterHandler(object):
@@ -35,7 +38,7 @@ class ConverterHandler(object):
     the plugin, along with a list of all supported units.
     '''
 
-    def usage(self):
+    def usage(self) -> str:
         return '''
                This plugin allows users to make conversions between
                various units, e.g. Celsius to Fahrenheit,
@@ -46,11 +49,11 @@ class ConverterHandler(object):
                all supported units.
                '''
 
-    def handle_message(self, message, bot_handler):
+    def handle_message(self, message: Dict[str, str], bot_handler: Any) -> None:
         bot_response = get_bot_converter_response(message, bot_handler)
         bot_handler.send_reply(message, bot_response)
 
-def get_bot_converter_response(message, bot_handler):
+def get_bot_converter_response(message: Dict[str, str], bot_handler: Any) -> str:
     content = message['content']
 
     words = content.lower().split()
@@ -72,8 +75,9 @@ def get_bot_converter_response(message, bot_handler):
                 results.append('`' + number + '` is not a valid number. ' + utils.QUICK_HELP)
                 continue
 
-            number = float(number)
-            number_res = copy.copy(number)
+            # cannot reassign "number" as a float after using as string, so changed name
+            convert_num = float(number)
+            number_res = copy.copy(convert_num)
 
             for key, exp in utils.PREFIXES.items():
                 if unit_from.startswith(key):
@@ -83,14 +87,14 @@ def get_bot_converter_response(message, bot_handler):
                     exponent -= exp
                     unit_to = unit_to[len(key):]
 
-            uf_to_std = utils.UNITS.get(unit_from, False)
-            ut_to_std = utils.UNITS.get(unit_to, False)
+            uf_to_std = utils.UNITS.get(unit_from, [])  # type: List[Any]
+            ut_to_std = utils.UNITS.get(unit_to, [])  # type: List[Any]
 
-            if uf_to_std is False:
+            if not uf_to_std:
                 results.append('`' + unit_from + '` is not a valid unit. ' + utils.QUICK_HELP)
-            if ut_to_std is False:
+            if not ut_to_std:
                 results.append('`' + unit_to + '` is not a valid unit.' + utils.QUICK_HELP)
-            if uf_to_std is False or ut_to_std is False:
+            if not uf_to_std or not ut_to_std:
                 continue
 
             base_unit = uf_to_std[2]
--- a/zulip_bots/zulip_bots/bots/converter/test_converter.py
+++ b/zulip_bots/zulip_bots/bots/converter/test_converter.py
@@ -5,15 +5,15 @@ from zulip_bots.test_lib import StubBotTestCase
 class TestConverterBot(StubBotTestCase):
     bot_name = "converter"
 
-    def test_bot(self):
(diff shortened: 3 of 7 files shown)


.gitignore
```diff
@@ -47,3 +47,6 @@ flaskbotrc
 
 # mypy
 .mypy_cache
+
```
neiljp: I would remove this part of the commit; this file could be added to the .gitignore, but it would be better to discuss elsewhere and place into a separate PR.

tools/run-mypy
```diff
@@ -36,6 +36,12 @@ exclude = [
 force_include = [
     # Include bots that we migrate to mypy.
     "zulip_bots/zulip_bots/bots/helloworld/helloworld.py",
+    "zulip_bots/zulip_bots/bots/converter/converter.py",
+    "zulip_bots/zulip_bots/bots/converter/test_converter.py",
+    "zulip_bots/zulip_bots/bots/define/define.py",
+    "zulip_bots/zulip_bots/bots/encrypt/encrypt.py",
+    "zulip_bots/zulip_bots/bots/encrypt/test_encrypt.py"
+
```
neiljp: You don't have test_define.py here?

zulip_bots/zulip_bots/bots/converter/converter.py
```diff
@@ -83,14 +87,14 @@ def get_bot_converter_response(message, bot_handler):
                     exponent -= exp
                     unit_to = unit_to[len(key):]
 
-            uf_to_std = utils.UNITS.get(unit_from, False)
-            ut_to_std = utils.UNITS.get(unit_to, False)
+            uf_to_std = utils.UNITS.get(unit_from, [])
+            ut_to_std = utils.UNITS.get(unit_to, [])
 
-            if uf_to_std is False:
+            if uf_to_std is []:
```
neiljp: I'm aware that I suggested moving to the empty-list approach here, and you have the right idea I think.
However, your use of `is` here will not work as the objects being compared cannot be the same exact object (this also indicates how we are not testing the converter bot sufficiently, as I think this should cause a test to fail).
You'll need a different approach for the check; let me or a mentor know about this - I know we chatted about the mypy issue before on Zulip.

KaiUsesThis:
@neiljp Got it, thanks. I wasn't too sure about changing the original code of the converter bot too much so I simply replaced all instances of `False` with `[]`. I understand where you are coming from and I'll discuss it.

This converter bot is giving a lot more trouble than I expected!

zulip_bots/zulip_bots/bots/converter/converter.py
```diff
@@ -8,9 +8,11 @@
 from math import log10, floor
 
 import re
-from zulip_bots.bots.converter import utils
+from zulip_bots.zulip_bots.bots.converter import utils
```
eeshangarg: This import is wrong, the previous import was correct! :) So I would suggest changing this back to what it was. :)

eeshangarg: This is also why Travis is failing! :)

zulip_bots/zulip_bots/bots/converter/converter.py
```diff
@@ -72,8 +75,9 @@ def get_bot_converter_response(message, bot_handler):
                 results.append('`' + number + '` is not a valid number. ' + utils.QUICK_HELP)
                 continue
 
-            number = float(number)
-            number_res = copy.copy(number)
+            # cannot reassign "number" as a float after using as string, so changed name
+            convertNum = float(number)
```
eeshangarg: The variable `convertNum` is what you call camelCase, which violates [the PEP8 coding standard](https://www.python.org/dev/peps/pep-0008/). The variable should be `convert_num` instead! :)

zulip_bots/zulip_bots/bots/converter/converter.py
```diff
@@ -83,14 +87,14 @@ def get_bot_converter_response(message, bot_handler):
...
+            ut_to_std = utils.UNITS.get(unit_to, [])
 
-            if uf_to_std is False:
+            if uf_to_std == []:
                 results.append('`' + unit_from + '` is not a valid unit. ' + utils.QUICK_HELP)
-            if ut_to_std is False:
+            if ut_to_std == []:
                 results.append('`' + unit_to + '` is not a valid unit.' + utils.QUICK_HELP)
-            if uf_to_std is False or ut_to_std is False:
+            if uf_to_std == [] or ut_to_std == []:
```
eeshangarg: I think its more Pythonic to say `if not uf_to_std` instead of explicitly checking whether it is equal to an empty list. I would recommend using `if not uf_to_std` for all of these. This is again a PEP8 thing, which I highly recommend reading. :) Here's [a link](https://www.python.org/dev/peps/pep-0008/).

zulip_bots/zulip_bots/bots/converter/test_converter.py
```diff
@@ -5,8 +5,8 @@
 class TestConverterBot(StubBotTestCase):
     bot_name = "converter"
 
-    def test_bot(self):
-        dialog = [
+    def test_bot(self) -> None:
+        expected = [
             ("", 'Too few arguments given. Enter `@convert help` '
                  'for help on using the converter.\n'),
             ("foo bar", 'Too few arguments given. Enter `@convert help` '
```
eeshangarg: You forgot to change `dialog` to `expected` at the end of this test, which is why tests are failing. Also, I recommend running `./tools/test-bots` and `./tools/run-mypy` before pushing your commits! :)

zulip_bots/zulip_bots/bots/define/define.py
```diff
@@ -66,7 +68,8 @@ def get_bot_define_response(self, original_content):
 
             except Exception as e:
                 response += self.REQUEST_ERROR_MESSAGE
-                logging.exception(e)
+                # logging.exception expects a string
+                logging.exception(str(e))
```
eeshangarg: I know what's causing this, in Python3, you have to pass in a str to `logging.exception`. I would recommend removing the comment you added and doing `logging.exception("")` instead. `logging.exception` should take care of embedding the exception info in the logged message automatically! :)

eeshangarg:
@KaiUsesThis: I think you are on the right track, however, this needs a little bit of work before we can merge it. I left a few comments and it would be awesome if you could please address those! :)

Additionally, I would highly recommend running `./tools/test-bots` and `./tools/run-mypy` and making sure that both of them pass before pushing your commits. That way you won't have to wait for a review or Travis to let you know that something is amiss! :) Thanks and good luck! :)

KaiUsesThis:
@eeshangarg Thanks! I've fixed issues you brought up. I've ran all tests locally, too. Also made some changes that I'm not too sure about, so please take a look at them.

zulip_bots/zulip_bots/bots/converter/converter.py
```diff
@@ -83,14 +87,14 @@ def get_bot_converter_response(message, bot_handler):
                     exponent -= exp
                     unit_to = unit_to[len(key):]
 
-            uf_to_std = utils.UNITS.get(unit_from, False)
-            ut_to_std = utils.UNITS.get(unit_to, False)
+            uf_to_std = utils.UNITS.get(unit_from, [])  # type: Any
```
KaiUsesThis: If `type: any` is not specified, mypy gives an error: 
conversion-error

Is specifying that okay?

zulip_bots/zulip_bots/bots/converter/test_converter.py
```diff
@@ -5,15 +5,15 @@
...
                         'for help on using the converter.\n'),
-            ("2 m cm", "2.0 m = 200.0 cm\n"),
+            ("2 m cm", "2 m = 200.0 cm\n"),
             ("12.0 celsius fahrenheit", "12.0 celsius = 53.600054 fahrenheit\n"),
             ("0.002 kilometer millimile", "0.002 kilometer = 1.2427424 millimile\n"),
-            ("3 megabyte kilobit", "3.0 megabyte = 24576.0 kilobit\n"),
+            ("3 megabyte kilobit", "3 megabyte = 24576.0 kilobit\n"),
```
eeshangarg: Why did you make these changes?

zulip_bots/zulip_bots/bots/converter/converter.py
eeshangarg: Hmm.. weird, I don't get any errors when I pulled your PR and ran the tests locally. @neiljp: Any ideas? :)

eeshangarg: @KaiUsesThis could you try running mypy again and make sure you are on the correct branch? :)

eeshangarg: @KaiUsesThis: Maybe try `# type: List[Any]`, since that seems to be the correct annotation! :)

eeshangarg:
@KaiUsesThis: Great work! I left a couple of comments/questions! :)

zulip_bots/zulip_bots/bots/converter/test_converter.py
KaiUsesThis: The tests were expecting it to be `3 megabyte` instead of `3.0 megabyte`.

zulip_bots/zulip_bots/bots/converter/converter.py
KaiUsesThis: Ran mypy again, no errors when ran with `type: any`, but there were errors when `type: any` was removed. I am on the correct branch

KaiUsesThis: @eeshangarg Okay, I've changed the type to `List[Any]`

showell:
Merged!  Thanks @KaiUsesThis and @eeshangarg!

PR #226, opened 2017-12-09

skunkmb:
Add annotations for xkcd, Wikipedia, and Yoda Bot.

Also, make a few minor changes inside of some functions to accommodate for the new typing:
 1. Use `return None` instead of just `return` in [`get_bot_wiki_response` for Wikipedia Bot](https://github.com/zulip/python-zulip-api/compare/master...skunkmb:add-bots-mypy-6?expand=1#diff-f466b0b628423593f706fa17fa3e1601R67).
 2. Use `logging.warning(str(e))` instead of `logging.warning(e)` in [`fetch_xkcd_query` for xkcd Bot](https://github.com/zulip/python-zulip-api/compare/master...skunkmb:add-bots-mypy-6?expand=1#diff-2414b1dc9f66a1edc5cc2950c177e52fR119).
 3. Use `response.json()` instead of `response.text` in [`send_to_yoda_api` for Yoda Bot.](https://github.com/zulip/python-zulip-api/compare/master...skunkmb:add-bots-mypy-6?expand=1#diff-d20394f5dc43134ad0ac41457d27cb23R81)

### DIFF
--- a/zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
+++ b/zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
@@ -4,6 +4,8 @@ import requests
 import logging
 import re
 from six.moves import urllib
+from zulip_bots.lib import ExternalBotHandler
+from typing import Optional
 
 # See readme.md for instructions on running this code.
 
@@ -24,7 +26,7 @@ class WikipediaHandler(object):
         'description': 'Searches Wikipedia for a term and returns the top 3 articles.',
     }
 
-    def usage(self):
+    def usage(self) -> str:
         return '''
             This plugin will allow users to directly search
             Wikipedia for a specific key term and get the top 3
@@ -32,11 +34,11 @@ class WikipediaHandler(object):
             should preface searches with "@mention-bot".
             @mention-bot <name of article>'''
 
-    def handle_message(self, message, bot_handler):
+    def handle_message(self, message: dict, bot_handler: ExternalBotHandler) -> None:
         bot_response = self.get_bot_wiki_response(message, bot_handler)
         bot_handler.send_reply(message, bot_response)
 
-    def get_bot_wiki_response(self, message, bot_handler):
+    def get_bot_wiki_response(self, message: dict, bot_handler: ExternalBotHandler) -> Optional[str]:
         '''This function returns the URLs of the requested topic.'''
 
         help_text = 'Please enter your search term after @mention-bot'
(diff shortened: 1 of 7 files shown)


zulip_bots/zulip_bots/bots/xkcd/xkcd.py
```diff
@@ -111,7 +116,7 @@ def fetch_xkcd_query(mode, comic_id=None):
 
         xkcd_json = fetched.json()
     except requests.exceptions.ConnectionError as e:
-        logging.warning(e)
+        logging.warning(str(e))
         raise
```
eeshangarg: Okay, I know what's causing this and I think it would be better to do the following:
```
logging.exception("Connection Error")
```
`logging.exception` automatically takes care of including the exception info in the logged message! :)

zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
```diff
@@ -24,19 +26,27 @@ class WikipediaHandler(object):
...
         bot_response = self.get_bot_wiki_response(message, bot_handler)
         bot_handler.send_reply(message, bot_response)
 
-    def get_bot_wiki_response(self, message, bot_handler):
+    def get_bot_wiki_response(
+        self,
+        message: dict,
+        bot_handler: ExternalBotHandler
+    ) -> Optional[str]:
         '''This function returns the URLs of the requested topic.'''
```
eeshangarg: One minor concern I have is I don't think we need to break the annotation over multiple lines if they can fit on one line without sacrificing readability! Maybe take a look into seeing which one of these can be fit on one line? I think the only time we should make an annotation span over multiple lines is when there are more than half a dozen arguments or something! :)

eeshangarg:
@skunkmb: Thanks for working on this! This is close to completion! I left a couple of comments, I'd be happy to merge this once you address those! :)

zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
skunkmb: Hi @eeshangarg! The reason I initially spread definitions over multiple lines is if they went over 80 characters. 

```
    def get_bot_wiki_response(self, message: dict, bot_handler: ExternalBotHandler) -> Optional[str]:
```

is 101 characters (with the indent), which is why I spread it over multiple lines. Is that not a concern? If so, I'd be happy to make it go on only one line.

zulip_bots/zulip_bots/bots/xkcd/xkcd.py
skunkmb: @eeshangarg OK, changed. Thanks!

digi0ps:
Looks great @skunkmb. Thanks for working on this!
Ping @eeshangarg for merging.

showell:
Merged!  Thanks @skunkmb @digi0ps @eeshangarg 

FYI there was a merge conflict on the last commit after I merged some work from @neiljp, so I just fixed those myself.  See https://github.com/zulip/python-zulip-api/commit/3c794b384b4183c390e86e5802b50c1dda4b4ed2

PR #232, opened 2017-12-11
Improvements to wikpedia bot and tests for define bot

SivagiriVisakan:
This PR brings 100%  test coverage define bot.
Also when exploring the  wikipedia bot , I found that the bot replied with `null` in case of an error which is not  the very best response for the a normal user. So I made a change to return an error message incase of an error.

### DIFF
--- a/zulip_bots/zulip_bots/bots/define/test_define.py
+++ b/zulip_bots/zulip_bots/bots/define/test_define.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 from zulip_bots.test_lib import StubBotTestCase
+from unittest.mock import patch
 
 class TestDefineBot(StubBotTestCase):
     bot_name = "define"
@@ -46,3 +47,10 @@ class TestDefineBot(StubBotTestCase):
         # Empty messages are returned with a prompt to reply. No request is sent to the Internet.
         bot_response = "Please enter a word to define."
         self.verify_reply('', bot_response)
+
+    def test_connection_error(self) -> None:
+        with patch('requests.get', side_effect=Exception), \
+                patch('logging.exception'):
+            self.verify_reply(
+                'aeroplane',
+                '**aeroplane**:\nCould not load definition.')
--- a/zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
+++ b/zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
@@ -56,12 +56,15 @@ class WikipediaHandler(object):
 
         except requests.exceptions.RequestException:
             logging.error('broken link')
-            return None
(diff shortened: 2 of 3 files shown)


zulip_bots/zulip_bots/bots/define/test_define.py
```diff
@@ -46,3 +47,10 @@ def test_bot(self) -> None:
         # Empty messages are returned with a prompt to reply. No request is sent to the Internet.
         bot_response = "Please enter a word to define."
         self.verify_reply('', bot_response)
+
+    def test_connection_error(self) -> None:
+        with patch('requests.get', side_effect=Exception), \
```
showell: Would request_test_lib.mock_request_exception() work here?

zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
```diff
@@ -56,12 +56,15 @@ def get_bot_wiki_response(self, message: dict, bot_handler: ExternalBotHandler)
 
         except requests.exceptions.RequestException:
             logging.error('broken link')
-            return None
+            return 'Uh-Oh ! Sorry ,couldn\'t process the request right now.:slightly_frowning_face:\n' \
```
showell: We should probably add test coverage for this.

SivagiriVisakan: I already have the tests done [here](https://github.com/zulip/python-zulip-api/pull/232/files#diff-418787db0bb7c23c3cf2cf65150abf5eR62)

zulip_bots/zulip_bots/bots/define/test_define.py
SivagiriVisakan: It would , but since the code was trying to catch any exception, I added it like this as `Exception`.
Code that is being tested here : 
```python
except Exception as e:
        response += self.REQUEST_ERROR_MESSAGE
        logging.exception("")
```
Should I change it to use `request_test_lib.mock_request_exception()` ?

zulip_bots/zulip_bots/bots/wikipedia/wikipedia.py
showell: oh, gotcha

zulip_bots/zulip_bots/bots/define/test_define.py
showell: Ah...I understand now...maybe `mock_request_exception` is not a perfect fit.

showell:
merged, thanks @SivagiriVisakan 

(I think both of my comments were just misunderstandings on my part.  Thanks for clarifying both.)

PR #234, opened 2017-12-11
mypy: Add --disallow-any=generics & extend typing accordingly.

neiljp:
Also reset typing of ExternalBotHandler to Any after discussion.

This should also cause mypy to complain in future if `dict` or `list` are used, cf `Dict` and `List`.

**Update**: rebased to fix recent commits.

**Outstanding issue**: jabber_mirror_backend.py has a commented-out line, which appears to be inconsistent with the typing, in how it's currently used.

**Issue resolved**.


showell:
merged, thanks @neiljp!

PR #235, opened 2017-12-12
bots: Simplify & add annotations Link shortener & tests.

neiljp:


showell:
Nice!  merged, thanks @skunkmb @neiljp

PR #236, opened 2017-12-14
Fix tests on Windows and make BotTestCase the only helper library.

roberthoenig:


showell:
merged, thanks @roberthoenig!

PR #243, opened 2017-12-17
Editing: Add .editorconfig from zulip/zulip (with note).

neiljp:
This is a new file for this repo, but taken from the amended file presented in zulip/zulip#7806 (adding .pyi support).


showell:
merged thanks @neiljp

PR #245, opened 2017-12-18
Create `welcome-message` api example.

fredfishgames:


zulip/zulip/examples/welcome-message
```diff
@@ -0,0 +1,70 @@
...
+whether you will get selected or not. You will learn a lot contributing to \
+Zulip in course of next few months and if you do a good job at that you \
+will get selected too :)\n \
+* Most important of all welcome to the Zulip family :octopus:'
+
+def handle_event(event):
+    storage = client.get_storage()
+    watchlist = list(storage['storage'].values())
+    if event['type'] == 'realm_user' and event['op'] == 'add':
+        watchlist.append(event['person']['email'])
```
showell: You can just do an early return here, to keep the code "flat."

showell:
I added a comment inline, plus two other things to fix:

* Can you restrict this to just the "new members" stream?  (We can ignore anything they post elsewhere for now, although maybe if they send anything to "development help" first, take them off the watchlist.
* Can you clean up your commit message?

zulip/zulip/examples/welcome-message
fredfishgames: Wouldn't returning here mean that the storage won't be updated?
Should I put client.update_storage here?

showell: and it's fine to do the storage save there...the trivial code duplication is worth it to avoid making the "main" code get buried in an "else" clause, I think

showell:
You'll want to capitalize "Create" in the commit message.

showell:
Merged!  Thanks @fredfishgames

PR #246, opened 2017-12-19
interactive bots: Create idonethis bot.

thejwzhang:


zulip_bots/zulip_bots/bots/idonethis/idonethis.py
```diff
@@ -0,0 +1,78 @@
...
+        data = requests.get(url, headers=auth)
+    except requests.exceptions.ConnectionError as e:  # Usually triggered by bad connection.
+        logging.exception('Bad connection')
+        raise
+    data.raise_for_status()
+    try:
+        counter = 0
+        while (counter < number):
+            entries += data.json()[counter]['body']
+            counter = counter + 1
```
sampritipanda: Here, if the user asks for the last 10 entries but there are only 5 entries, it will raise the NoResultException right?
We should return the 5 results we get back instead I think?

PR #248, opened 2017-12-19
Update dialogflow bot to show user friendly messages

fredfishgames:
These changes improve the dialogflow bot by:

 - Instead of sending the actual error e.g. `error 204: error 500 internal server error.`, the bot will now send the message sent, if available, e.g. `could not get the weather.`
 - Fixed mistake where if the dialogflow sent an empty message with no alternate response, it would raise a KeyError.


showell:
merged, thanks @fredfishgames

PR #249, opened 2017-12-21
mypy: Upgrade to 0.560 & improve type checking.

neiljp:


showell:
merged thanks @neiljp !

PR #250, opened 2017-12-23
mypy: Include strict-optional by default when running tools/run-mypy

neiljp:
Outstanding issues are within the api (zulip/zulip/__init__.py) and various integrations, totalling 21 errors locally. Pending review, all but the tools/run-mypy change should be mergeable now.

**Jan 8th 2018** Initial patches were merged in #251, so I rebased here and added commits which should resolve all outstanding mypy+strict-optional issues. The tools commit is now the last commit, but github does not show the correct order.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -337,6 +337,7 @@ class Client(object):
             raise RuntimeError("api_key or email not specified and file %s does not exist"
                                % (config_file,))
 
+        assert(api_key is not None and email is not None)
         self.api_key = api_key
         self.email = email
         self.verbose = verbose
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -588,7 +588,7 @@ class Client(object):
                 else:
                     res = self.register(event_types=event_types, narrow=narrow)
 
-                if 'error' in res.get('result'):
+                if 'error' in res['result']:
                     if self.verbose:
                         print("Server returned error:\n%s" % res['msg'])
                     time.sleep(1)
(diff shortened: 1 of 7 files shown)


zulip/zulip/__init__.py
```diff
@@ -588,12 +588,16 @@ def do_register():
                 else:
                     res = self.register(event_types=event_types, narrow=narrow)
 
-                if 'error' in res.get('result'):
-                    if self.verbose:
-                        print("Server returned error:\n%s" % res['msg'])
-                    time.sleep(1)
-                else:
-                    return (res['queue_id'], res['last_event_id'])
+                try:
```
showell: I think we can make these diffs smaller and the exception handling less aggressive as described here:

https://chat.zulip.org/#narrow/stream/integrations/subject/Strict-optional.20mypy.20.23P250/near/433795

zulip/zulip/__init__.py
```diff
@@ -604,36 +608,40 @@ def do_register():
...
-                # Add a pause here to cover against potential bugs in this library
-                # causing a DoS attack against a server when getting errors.
-                # TODO: Make this back off exponentially.
-                time.sleep(1)
-                continue
-
-            for event in res['events']:
-                last_event_id = max(last_event_id, int(event['id']))
-                callback(event)
+            try:
```
showell: same suggestion for this section as above

showell:
Except for the very first commit (see inline comments), all these LGTM, although I wouldn't bet my life on some of the more obscure ones, haha.  (I didn't make time to test out our subversion integration.)

neiljp:
As per further discussion on that thread, I've removed the try/except blocks and am shortly going to force push with that change.

I've left the Zephyr check_mirroring changes in for now, which are very similar, but for which there is a pre-existing exception block around similar code.

neiljp:
Now updated the Zephyr change to be minimal.

showell:
Merged, thanks @neiljp!

PR #251, opened 2017-12-23
Strict optional mypy patches without tool change from #250

neiljp:
This should pass tests.


showell:
Merged!  Thanks @neiljp!

PR #253, opened 2017-12-25
Improve connect four bot

sColin16:
These changes address @showell requests to create a general game framework, by introducing a GameAdapter class that can be easily used to create bots that manage single player and two player games. The class currently is located within the connect_four bot's folder, although it should eventually be moved to another folder to allow any bot to use it. 

The changes also allow superusers to be added to the config file. These 'superusers' are moderators who can reset the bot in case users abuse the fact that the bot can only handle a single game at a time.

### DIFF
--- a/zulip_bots/zulip_bots/bots/connect_four/connect_four.conf
+++ b/zulip_bots/zulip_bots/bots/connect_four/connect_four.conf
@@ -1,0 +1,2 @@
+[connect_four]
+superusers = ["user@example.com"]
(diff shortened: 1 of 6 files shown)


showell:
Merged thanks @sColin16!

showell:
I think at this point the next step is to try to find another simple two-player game to test out the framework.

sColin16:
@showell I think I should be able to do that. Would another tic-tac-toe bot be O.K, or should I think of another game?

showell:
Let's find another game than tic-tac-toe.

PR #257, opened 2017-12-25
Test run_message_handler_for_bot() and add a 'full_message' field for bots.

roberthoenig:
Note that currently, the test for `run_message_handler_for_bot()` validates only one message. The framework is built though, and adding new test messages is straightforward.


showell:
merged thanks @roberthoenig

PR #259, opened 2017-12-26
python3 shebangs for zulip_bots test scrpits

sibasish14:
Added python3 shebangs in zulip_bot test scripts.
Fixes https://github.com/zulip/python-zulip-api/issues/187


showell:
There may have been some miscommunication here.  We don't want to add shebangs to non-executable files like the ones in this PR.

Instead, we want to update python3 shebangs for scripts like `tools/test-bots`.

I'm closing this.

PR #269, opened 2017-12-29
update run-mypy: Add Baremetrics bot files.

ViRu-ThE-ViRuS:
Added the baremetrics bot files to `run-mypy.py`.


showell:
Ah, this was an easy change but important one.  Good catch, @ViRu-ThE-ViRuS!

ViRu-ThE-ViRuS:
Thanks!

PR #271, opened 2017-12-29
presence: Make a request to `mark_bot_listening` when a bot starts.

skunkmb:
(This pull request should only be merged after/if zulip/zulip#7938 is merged.)

zulip/zulip#7938 creates a mechanism for marking a bot as `listening` and displaying it in the presence sidebar if it is. This PR calls that endpoint whenever a bot is started/stopped so that it displayed.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -673,6 +673,17 @@ class Client(object):
             request=message_data,
         )
 
+    def mark_bot_listening(self, bot_name, is_listening):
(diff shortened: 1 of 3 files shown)


showell:
@skunkmb Thanks for working on this!  I'm going to close this, because I believe the backend can automatically detect when bots are present, and I've advised @roberthoenig to look into this.  It's likely that we will use some of your code from the other PR that you mention above.

PR #275, opened 2017-12-30
Add ExternalBotHandler.quit()

roberthoenig:
@showell can you review the approach I took here?

If you agree, I'll update the other bots currently using sys.exit(), and the ones using logging.error() where quit() should be used.


showell:
Merged, thanks @roberthoenig!  I like how you decided to have the test lib throw an exception.  I sorta wonder whether the other library should do the same thing and then have `run.py` catch it, but that would be more moving parts, so I'm inclined to see how this change feels at first.

PR #276, opened 2017-12-31
Improve test coverage for Mention bot.

ViRu-ThE-ViRuS:
Issue #122 
Added more tests.


zulip_bots/zulip_bots/bots/mention/test_mention.py
```diff
@@ -30,3 +30,11 @@ def test_get_alert_id(self) -> None:
...
         with self.mock_http_conversation('get_alert_id'):
             self.assertEqual(bot_test_instance.get_alert_id('TEST'), 'TEST')
+
+    def test_get_mentions(self) -> None:
+        bot_test_instance = MentionHandler()
+        bot_test_instance.access_token = 'TEST'
+        bot_test_instance.account_id = 'TEST'
+
+        with self.mock_http_conversation('get_mentions'):
+            self.assertEqual(bot_test_instance.get_mentions('TEST'), [])
```
showell: Can we test this with more data than an empty list?

ViRu-ThE-ViRuS: I tried. It was difficult getting demo data with a free account.

showell: What I would do in this case is add mentions with text like "simulated mention for test purposes #1" and "simulated mention for test purposes #2", just so we can exercise the code a bit.  (The test still won't really ensure that we are correct for the actual API, but it will at least ensure that the code works as intended.)

ViRu-ThE-ViRuS: Ok. Will do that today itself.

ViRu-ThE-ViRuS:
@showell I tried a number of approaches... I settled with just testing one element in the returned array with the expected values (same as in the fixture). It closely resembles real world situations.

ViRu-ThE-ViRuS:
The other bots will need to be restructured to get higher code coverage. How shall we proceed with this?
Quite a few of the uncovered areas are the code sections with `try/except` blocks.

showell:
Merged, thanks @ViRu-ThE-ViRuS!

PR #289, opened 2018-01-06
Improve test coverage for Zulip Bots.

ViRu-ThE-ViRuS:
Issue #122 
(Not a GCI Task.)

### DIFF
--- a/zulip_bots/zulip_bots/bots/baremetrics/test_baremetrics.py
+++ b/zulip_bots/zulip_bots/bots/baremetrics/test_baremetrics.py
@@ -1,5 +1,7 @@
 from unittest.mock import patch
 from zulip_bots.test_lib import BotTestCase
+from zulip_bots.test_lib import StubBotHandler
+from zulip_bots.bots.baremetrics.baremetrics import BaremetricsHandler
 
 class TestBaremetricsBot(BotTestCase):
     bot_name = "baremetrics"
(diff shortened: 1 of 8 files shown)


zulip_bots/zulip_bots/bots/yoda/yoda.py
```diff
@@ -72,7 +72,7 @@ def send_to_yoda_api(self, sentence: str) -> str:
             error_message = response.json()['message']
             logging.error(error_message)
             error_code = response.status_code
-            error_message = error_message + 'Error code: ' + error_code +\
+            error_message = error_message + 'Error code: ' + str(error_code) +\
```
ViRu-ThE-ViRuS: Reason: `error_code` is an int according to the fixtures.

zulip_bots/zulip_bots/bots/baremetrics/test_baremetrics.py
```diff
@@ -66,3 +67,29 @@ def test_list_subscriptions_command(self) -> None:
         with self.mock_config_info({'api_key': 'TEST'}):
             with self.mock_http_conversation('list_subscriptions'):
                 self.verify_reply('list-subscriptions TEST', r)
+
+    def test_exception_when_api_key_is_invalid(self)-> None:
+        with self.mock_config_info({'api_key': 'TEST'}):
+            with self.assertRaises(StubBotHandler.BotQuitException):
+                self.verify_reply('help', '''
```
showell: If you expect the "help" command to raise an exception here, the long expected response for `verify_reply` seems distracting here.  Also, it's not clear to me why the `TEST` fixture should cause us to quit, and should this be happening in the `intialize` method, where we can have finer grain testing?

zulip_bots/zulip_bots/bots/mention/test_mention.py
```diff
@@ -43,3 +44,8 @@ def test_get_mentions(self) -> None:
             bot_response = bot_test_instance.get_mentions('TEST')[0]
             self.assertEqual(bot_response['title'], 'TEST')
             self.assertEqual(bot_response['original_url'], 'TEST')
+
+    def test_exception_when_api_key_is_invalid(self) -> None:
+        with self.mock_config_info({'access_token': 'TEST'}):
+            with self.assertRaises(StubBotHandler.BotQuitException):
```
showell: See my other comment.  Why are supplying an expected reply here if we expect this to raise an exception, and why aren't we testing a more specific thing like just initializing?

ViRu-ThE-ViRuS: Did this to to use `context manager` to cover line 33 in `baremetrics.py`.  Just initializing was not covering that line.
[Code Coverage Report](https://codecov.io/gh/zulip/python-zulip-api/src/master/zulip_bots/zulip_bots/bots/baremetrics/baremetrics.py)

zulip_bots/zulip_bots/bots/baremetrics/test_baremetrics.py
ViRu-ThE-ViRuS: Yea, you are right. `help` here is just a filler command; will replace the response with a simpler and shorter one since it is supposed to raise an exception. The case should quit because of the invalid api key in the fixture.
I did this in the way it is because I needed to use `context manager` to get line 33 in `baremetrics.py` actually covered. [Report](https://codecov.io/gh/zulip/python-zulip-api/src/master/zulip_bots/zulip_bots/bots/baremetrics/baremetrics.py)

zulip_bots/zulip_bots/bots/mention/test_mention.py
ViRu-ThE-ViRuS: You are right here as well. Will change it so that we test just the initialization, and notify. Thanks for your review!

ViRu-ThE-ViRuS:
@showell Done!

showell:
Merged! @ViRu-ThE-ViRuS Thanks for following up here.  I like how the tests just directly call `initialize` on the bot handlers, so there's very little setup, which should make failures nice and clear if people change the underlying code here.

PR #292, opened 2018-01-08
zulip_bots: Use package_data instead of MANIFEST.in.

eeshangarg:
This commit removes generate_manifest.py and package data files
are now included using the package_data kwarg to setup().

This is because, in certain situations, MANIFEST.in is a bit
finicky. For instance installing a package using:

python setup.py install

doesn't include files specified in MANIFEST.in, while using

pip install ./zulip_bots

does. package_data doesn't pose this problem, ergo it's better
for us.

@roberthoenig: FYI :)


roberthoenig:
@eeshangarg this is really cool!

PR #294, opened 2018-01-09
requirements: Remove runtime dependency on mock.

### LINKED ISSUE #293: zulip 0.4.0 lists runtime dependency on `mock`
0.3.9 didn't; this should probably not be required except for testing.

eeshangarg:
The zulip and zulip_botserver packages specify mock as a runtime
dependency, which is only needed for testing during development.
So, it made more sense to move it to requirements.txt so that
it can be installed by ./tools/provision during development.

Fixes #293.

@roberthoenig FYI :)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,7 @@ crayons
 twine
 coverage>=4.4.1
 pycodestyle==2.3.1
+mock
 -e ./zulip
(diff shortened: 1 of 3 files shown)


roberthoenig:
Thanks for the quick fix @eeshangarg!

PR #296, opened 2018-01-11
zulip: Change remove_subscriptions to update_subscriptions.

eeshangarg:
The underlying endpoint that remove_subscriptions called was
zerver.streams.view.update_subscriptions_backend, which is more
for bulk removing *and* adding subscriptions. Also there is a
separate endpoint for removing subscriptions that gives more
control than this one. So this method should be more aptly
called update_subscriptions and should allow one to add
subscriptions as well.

Note that once we make the next PyPI release and update the main
repo to use it, ./zulip/tools/test-api should fail, and would
need to be updated in the PR that updates the main repo to use
the latest release of the packages.

@roberthoenig, @showell: Could I please get a quick review? Thanks! :)

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -810,12 +810,15 @@ class Client(object):
             request=request,
         )
 
-    def remove_subscriptions(self, streams):
-        # type: (Iterable[str]) -> Dict[str, Any]
+    def update_subscriptions(self, add=[], delete=[]):
+        # type: (Optional[Iterable[Dict[str, Any]]], Optional[Iterable[str]]) -> Dict[str, Any]
         '''
             See examples/unsubscribe for example usage.
         '''
-        request = dict(delete=streams)
+        request = dict(
+            delete=delete,
+            add=add,
+        )
         return self.call_endpoint(
             url='users/me/subscriptions',
             method='PATCH',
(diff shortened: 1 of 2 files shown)


showell:
@eeshangarg I'm actually happy to hide the underlying complexity here.  I think the `remove_subscriptions` API is what users really need, and the more complex underlying endpoint should be deprecated.  I'm not sure we ever really need to simultaneous delete/remove subscriptions with a single endpoint.

So I'm closing this for now, unless you can convince me I'm thinking about this wrong.

timabbott:
Just a random bit of commentary since I saw this: There are some use cases for bulk-removing subscriptions via the API, but I don't expect them to be common.

eeshangarg:
@timabbott: From what I remember, in either case, bulk-removing shouldn't be a problem. But just from a general perspective, advertising `update_subscriptions` as `remove_subscriptions` just didn't seem right, since there was already `remove_subscriptions` that gave users a little bit more control (for example, removing someone else from a stream, which `update_subscriptions` couldn't do because it didn't take in the `principals` argument). So there are definitely some use cases yes, but the changes that were done as part of this shouldn't affect those! :) Thanks!

timabbott:
Yep, makes sense.

PR #297, opened 2018-01-11
zulip: Make a DELETE request in remove_subscriptions.

eeshangarg:
Previously, remove_subscriptions called the
PATCH /api/v1/users/me/subscriptions endpoint, which is more like
an ad-hoc endpoint for bulk adding/removing subscriptions for the
user that makes the request. However, making a DELETE request
allows an admin to pass in the `principals` argument to unsubscribe
other users from streams as well, which is more consistent with how
add_subscriptions works.

@showell: FYI :)

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -810,15 +810,18 @@ class Client(object):
             request=request,
         )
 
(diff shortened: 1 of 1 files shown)


showell:
Merged, thanks @eeshangarg!

PR #241, opened 2017-12-15
Remove markdown includes {!running-bot.md!} from bot docs.

YJDave:
Remove markdown includes from bot docs and inserted a link instead, cause markdown includes doesn't seem work(as shown in image).

runningbot

PR #303, opened 2018-01-17
interactive bots: Create Joinme bot.

itstakenalr:


itstakenalr:
@eeshangarg  @roberthoenig Hey, please can you review this?

tools/test-bots
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
```
eeshangarg: I don't think you should change this to `python`. The `zulipbots` code is now officially python3 only, so I would recommend running `./tools/provision -p /usr/bin/python3` to build a python3 venv if you aren't already. So, you should reverse this! :)

zulip_bots/zulip_bots/bots/joinme/doc.md
```diff
@@ -0,0 +1,30 @@
+# Joinme bot
+
+The Joinme bot posts a link to a Joinme call when being mentioned.
+
+## Setup
+
+1. See [here](https://developer.join.me/docs) for setting up an account and
```
eeshangarg: Change it to `See [Joinme's developer docs] for setting up...`

zulip_bots/zulip_bots/bots/joinme/doc.md
```diff
@@ -0,0 +1,30 @@
...
+   registering the app.
+2. open `bots/joinme/joinme.conf` and change the values of key, secret and the
+   callback url to the ones you created.
```
eeshangarg: Replace with:
```
2. Open `bots/joinme/joinme.conf` and replace the values of `url`, `secret` and
    `callback_url` with your credentials.
```

zulip_bots/zulip_bots/bots/joinme/doc.md
```diff
@@ -0,0 +1,30 @@
...
+3. See [here](https://zulipchat.com/api/running-bots#running-a-bot) for description
+   on how to run a bot.
```
eeshangarg: @roberthoenig: Didn't we have a macro for this?

zulip_bots/zulip_bots/bots/joinme/doc.md
```diff
@@ -0,0 +1,30 @@
...
+4. When using `zulip-run-bot joinme --config-file ~/zuliprc-joinme` to run joinme,
+   don't forget to include `--bot-config-file` and the path to `joinme.conf` at
+   the end. e.g.`zulip-run-bot joinme --config-file ~/zuliprc-joinme --bot-config
+   -file ~/zulip/python-zulip-api/zulip_bots/zulip_bots/bots/joinme/joinme.conf`
+
```
eeshangarg: Okay, so maybe do something like this for step 4:
```
4. You can run the bot by running the following command:

<And then the full command here>
```

zulip_bots/zulip_bots/bots/joinme/doc.md
```diff
@@ -0,0 +1,30 @@
...
+## Usage
+
+1. @-mention it first and it will give out a link which asks for permission
```
eeshangarg: Change it to `... it will respond with a link that asks...`

zulip_bots/zulip_bots/bots/joinme/doc.md
```diff
@@ -0,0 +1,30 @@
...
+   to view the personal url and to start a new meeting. Login is required.
```
eeshangarg: Replace `url` with `URL` :)

zulip_bots/zulip_bots/bots/joinme/doc.md
```diff
@@ -0,0 +1,30 @@
...
+
+2. Then @-mention the bot again with the url shown in the browser after
+   authorising Joinme.
```
eeshangarg: Add a comma after `Then`.

Replace `url` with URL. :)

zulip_bots/zulip_bots/bots/joinme/doc.md
```diff
@@ -0,0 +1,30 @@
...
+
+3. The url will expire in one minute, if it happens repeat step 1.
```
eeshangarg: Change the sentence to `The URL will expire in one minute. If so, just go back to step 1.`

zulip_bots/zulip_bots/bots/joinme/doc.md
```diff
@@ -0,0 +1,30 @@
...
+
+4. Finally the bot will post a link to the new meeting created.
+
```
eeshangarg: Change the sentence to `Finally, the bot will post a link to the newly created meeting.`

zulip_bots/zulip_bots/bots/joinme/joinme.py
```diff
@@ -0,0 +1,56 @@
...
+                           'X-Originating-Ip': s}
+                data = {"startWithPersonalUrl": 'false'}
+
+                req = requests.post('https://api.join.me/v1/meetings/start', headers = headers, data = json.dumps(data))
+                req = req.json()
+                content = "Click this link to join the call:\n {}".format(req['presenterLink'])
+
+        bot_handler.send_reply(message, content)
+
+handler_class = JoinmeHandler
```
eeshangarg: Could you please wrap all of this code at 80 characters? We should try to do that as much as possible unless there is a good reason for not doing so (e.g. long URLs or something).

zulip_bots/zulip_bots/bots/joinme/test_joinme.py
```diff
@@ -0,0 +1,24 @@
...
+        bot_response = "Please click the link below to log in and authorise " \
+                       "Joinme:\n None \nAnd please copy and send the url shown " \
+                       "in browser after clicking accept. Don't forget to @-mention me!"
+        with self.mock_config_info({'key': '1234567890', 'callback_url': 'https://callback.com'}), \
+                self.mock_http_conversation('test_login'):
+            self.verify_reply('', bot_response)
+
+    def test_bot_responds_to_empty_message(self) -> None:
+        # It works the same as test_login
+        pass
```
eeshangarg: If this test case is already covered by another test, you don't need this then! :)

tools/test-bots
roberthoenig: I agree with Eeshan here - the tests are supposed to run on python3, which is why we have this shebang. If you are on Windows and don't have `python3` in your path, just change the shebang for yourself locally.

zulip_bots/zulip_bots/bots/joinme/doc.md
roberthoenig: I think we did, in the Giphy bot. However, I remember someone changing it back to plain text during GCI: https://github.com/zulip/python-zulip-api/pull/241 We'll do a sweep on all bots in a month or so anyway, so I think this should be fine for now.

zulip_bots/zulip_bots/bots/joinme/joinme.py
```diff
@@ -0,0 +1,56 @@
+from typing import Any
+import requests
+import json
+import socket
+from six.moves.configparser import SafeConfigParser
```
roberthoenig: We don't support Python 2, so we don't need to use six. From Python 3.2 on, SafeConfigParser is available as ConfigParser. So you can just use 
```.py
from configparser import ConfigParser
```

zulip_bots/zulip_bots/bots/joinme/joinme.py
```diff
@@ -0,0 +1,56 @@
+from typing import Any
+import requests
+import json
+import socket
```
roberthoenig: Can you order your imports alphabetically?

zulip_bots/zulip_bots/bots/joinme/joinme.py
```diff
@@ -0,0 +1,56 @@
...
+from six.moves.configparser import SafeConfigParser
+
+class JoinmeHandler(object):
+    def usage(self) -> str:
+        return '''
+         A bot that will post a link to joinme call. Mention the bot and it will give out instructions.
```
roberthoenig: "This is bot will ..."

zulip_bots/zulip_bots/bots/joinme/joinme.py
```diff
@@ -0,0 +1,56 @@
...
+        '''
+
+    def handle_message(self, message: Any, bot_handler: Any) -> None:
+        self.config_info = bot_handler.get_config_info('joinme')
+        if 'ZISSHUDBUNIq' not in message['content']:  # The state, a special string to identify which type of url it is.
+            payload = {'client_id': self.config_info['key'], 'scope': 'user_info scheduler start_meeting',
+                       'redirect_uri': self.config_info['callback_url'],
+                       'state': 'ZISSHUDBUNIq', 'response_type': 'code'}
+            r = requests.get('https://secure.join.me/api/public/v1/auth/oauth2', params=payload)
+            if str(r) == '<Response [200]>':
```
roberthoenig: Use `r.status_code` here.

zulip_bots/zulip_bots/bots/joinme/joinme.py
```diff
@@ -0,0 +1,56 @@
...
+            authorization_code = authorization_code.split('=')
+            authorization_code = authorization_code[1].split('&')
+
+            payload = {"client_id": self.config_info['key'], "client_secret": self.config_info['secret'],
+                       "code": authorization_code[0], "redirect_uri": self.config_info['callback_url'],
+                       "grant_type": "authorization_code"}
+            req = requests.post('https://secure.join.me/api/public/v1/auth/token', data = payload)
+
+            token = json.loads(req.text)
+            if 'error' in token:
```
roberthoenig: Can 'error' only occur when the token has expired?

zulip_bots/zulip_bots/bots/joinme/joinme.py
```diff
@@ -0,0 +1,56 @@
...
+            else:
+                token = token['access_token']
+                s = socket.gethostbyname(socket.gethostname())
+
+                headers = {'Authorization': 'Bearer {}'.format(token), 'Content-Type': 'application/json',
+                           'X-Originating-Ip': s}
+                data = {"startWithPersonalUrl": 'false'}
+
+                req = requests.post('https://api.join.me/v1/meetings/start', headers = headers, data = json.dumps(data))
+                req = req.json()
```
roberthoenig: Use a different variable name here, e.g. req_json

zulip_bots/zulip_bots/bots/joinme/joinme.py
itstakenalr: Sorry what does this mean? Do I need to reduce this entire function down to 80 characters?  Because it sounds a bit impossible to me.

zulip_bots/zulip_bots/bots/joinme/test_joinme.py
itstakenalr: If I don't add this line it will then carry out another empty message test which gives strange error.

zulip_bots/zulip_bots/bots/joinme/joinme.py
itstakenalr: I haven't encountered any other error yet if you follow all the steps above.

zulip_bots/zulip_bots/bots/joinme/doc.md
eeshangarg: @roberthoenig: Cool, thanks Robert! :)

zulip_bots/zulip_bots/bots/joinme/joinme.py
eeshangarg: No I want you to take every line and make sure that they are no more than 80 characters long. So instead of doing:
```
some_function_call(argument1, argument2, argument3, argument4, argument5, argument6, argument7, argument8)
```
You wrap the arguments so that there are no more than 80 characters per line, like so:
```
some_function_call(argument1, argument2, argument3, 
                                argument4, argument5, argument6,
                                argument7, argument8)
```
Think if it this way, the above makes it easier to read two files side by side. Does that make sense? :)

zulip_bots/zulip_bots/bots/joinme/joinme.py
roberthoenig: Hmm, it might be best toi check Joinme's API documentation to see their error handling.

zulip_bots/zulip_bots/bots/joinme/joinme.py
itstakenalr: Yes it does! Thank you very much, I thought you meant I need to wrap the entire function into 80 characters.

zulip_bots/zulip_bots/bots/joinme/joinme.py
itstakenalr: But that is not going to work for the bot test, as the response in the test is in form of json, when the actual website response isn't.

zulip_bots/zulip_bots/bots/joinme/joinme.py
itstakenalr: https://developer.join.me/docs/read/faq According to this doc, other errors occur when you refresh the token, which is not needed in this case, or the bearer is in the wrong format. But the bearer's format is fixed in this case as the codes only sub values in the correct format written in the code so there shouldn't be any problem. Therefore I  think expired token is the only one we need to consider.

zulip_bots/zulip_bots/bots/joinme/test_joinme.py
eeshangarg: @itstakenalr: I'm not sure why that is the case! Could you delete it and post the error message here so we can take a look? You shouldn't need this.

itstakenalr:
```
  File "/home/travis/build/zulip/python-zulip-api/zulip_bots/zulip_bots/test_lib.py", line 161, in test_bot_responds_to_empty_message
    response = self.get_response(message)
  File "/home/travis/build/zulip/python-zulip-api/zulip_bots/zulip_bots/test_lib.py", line 106, in get_response
    bot.handle_message(message, bot_handler)
  File "/home/travis/build/zulip/python-zulip-api/zulip_bots/zulip_bots/bots/joinme/joinme.py", line 18, in handle_message
    payload = {'client_id': self.config_info['key'],
KeyError: 'key'
```
@eeshangarg

itstakenalr: It works fine when I test empty message on the server,  but not in the test.

itstakenalr: hey :) @eeshangarg

eeshangarg: @itstakenalr: Hey! Sorry for the late response, had a crazy busy week!

So this is kinda weird, it seems to be running a test that doesn't exist (now that you commented it out).

eeshangarg: And the problem is, it can't seem to find the `key` in your config! Can you try running just `test_login` to see if that works? It should! :)

itstakenalr: `test_login (zulip_bots.bots.joinme.test_joinme.TestHelpBot) ... ok`
@eeshangarg It works!

itstakenalr: @eeshangarg Will it be a good idea to pass the test like I did? I think it's not really affecting anything.

zulip_bots/zulip_bots/bots/joinme/test_joinme.py
```diff
@@ -0,0 +1,24 @@
...
+        bot_response = "Please click the link below to log in and authorise " \
+                       "Joinme:\n None \nAnd please copy and send the url shown " \
+                       "in browser after clicking accept. Don't forget to @-mention me!"
+        with self.mock_config_info({'key': '1234567890', 'callback_url': 'https://callback.com'}), \
+                self.mock_http_conversation('test_login'):
+            self.verify_reply('', bot_response)
+
+    #def test_bot_responds_to_empty_message(self) -> None:
+        # It works the same as test_login
+    #    pass
```
eeshangarg: Probably get rid of these comments now! :)

zulip_bots/zulip_bots/bots/joinme/test_joinme.py
eeshangarg: What do you mean by "like I did"? :)

itstakenalr: I mean just to pass the test like what I did in the first PR:
```
def test_bot_responds_to_empty_message(self) -> None:
         It works the same as test_login
        pass
```
@eeshangarg

eeshangarg: @itstakenalr: But the whole point is that you shouldn't need to do that! I thought you said that it didn't matter if we removed it? So why don't we just remove it?

itstakenalr: Because if I remove it, test_login still works indeed but then test for empty message comes up with that 'key' error. @eeshangarg

itstakenalr: @eeshangarg hey :) sorry for being annoying but can I simply skip the check_for_empty_message_test by using `pass` since it will solve the problem in a easier way?

eeshangarg:
@itstakenalr: Hey! Could you please take a look at the build errors here? We would love to get this merged! :) Thanks!

itstakenalr:
@eeshangarg I've fixed the errors, thank you!

PR #308, opened 2018-01-23
bots: Fix test failure for Baremetrics bot.

ViRu-ThE-ViRuS:
The API has been updated it seems. The message for invalid api key is now includes ' (001)' in the end. This was causing the tests to fail.


roberthoenig:
Thanks @ViRu-ThE-ViRuS! Merging.

PR #320, opened 2018-02-08
Create Dropbox bot to do primary file operations.

amanagr:
A ton of features can still be added to this bot.
I am working on test cases.


amanagr:
@roberthoenig I think it's ready for review.

roberthoenig:
@amanagr before I look at the code, can you run this on playground.zulipdev.org and ping some folks so we can test out the bot?

showell:
@amanagr Thanks for working on this!

I merged this, because the code looks really clean and we can always iterate on it.  We do want a figure out a way to get people using this to help test it out.  I also have concerns about the read command...we probably want to just print the first N bytes or something.

amanagr:
Yeah, since read command can download text upto 150mb, it can be a problem.

PR #321, opened 2018-02-11
bots: Add magic method validate_config().

roberthoenig:
This method allows bots to validate their config info
in a standardized way. Adding the method to a bot is
optional, but recommended for bots with config options
that can be invalid, like api keys. The giphy bot serves
as an example.
The primary reason behind this is to allow the zulip
backend to validate config data for embedded bots. The
backend does not have a permanent bot class instance, so
validate_config() must be a static method.

### DIFF
--- a/zulip_bots/zulip_bots/bots/giphy/giphy.py
+++ b/zulip_bots/zulip_bots/bots/giphy/giphy.py
@@ -7,6 +7,8 @@ import os
 import re
 from requests.exceptions import HTTPError, ConnectionError
 
+from zulip_bots.custom_exceptions import ConfigValidationError
+
 GIPHY_TRANSLATE_API = 'http://api.giphy.com/v1/gifs/translate'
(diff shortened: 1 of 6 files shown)


roberthoenig:
@eeshangarg this PR is mostly done for implementing a new feature in zulip core, so it'd be great if you could make a new release (and update master to use it) once this is merged.

roberthoenig:
Hmm, I now think that using `@staticmethod` for `validate_config` is perhaps unnecessary - I did it originally so we could avoid loading the whole bot model and calling initialize(), but I forgot that initialize is not the same as __init__, and only gets executed when explicitly called.

timabbott:
I checked with Steve and we're both happy with the static method.  Merged, thanks @roberthoenig!  

For #8366, you can just change it to a GitHub URL so we don't need to block on doing a bots release.

PR #325, opened 2018-02-17
Make `game_handler` support starting a game with a range of players.

amanagr:
Allow a game to start even if `max_players` is not reached.
Adding a new command `play game` to start a game if the
number of players is between max and min no. of players.


showell:
@amanagr Merged, thanks!

PR #332, opened 2018-03-02
bot testing : 100% test coverage 

DhruvThakker:
#122 Coverage 100% for link_shortener

### DIFF
--- a/zulip_bots/zulip_bots/bots/baremetrics/baremetrics.py
+++ b/zulip_bots/zulip_bots/bots/baremetrics/baremetrics.py
@@ -205,10 +205,9 @@ class BaremetricsHandler(object):
 
         url = 'https://api.baremetrics.com/v1/{}/plans'.format(parameters[0])
         create_plan_response = requests.post(url, data=data_header, headers=self.auth_header)
-
         if 'error' not in create_plan_response.json():
             return 'Plan Created.'
         else:
-            return 'Some Error Occurred. Maybe the passed arguments are invalid.'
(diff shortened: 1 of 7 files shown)


showell:
This looks good...can you squash into a single commit with a nice commit message?

showell:
@DhruvThakker The code here looks pretty good.  Can you change the commit messages so they have a verb in them?  An example would be:

~~~
github_detail_bot: Add tests for bad requests.

(This module now has 100% coverage.)
~~~

I'm kind of on the fence of having the helper method being added to `test_lib.py`.  I would prefer the tests to just directly import the method from the other module, or, if you really want the helper, split out a separate commit for that.

Thanks for working on this!

showell:
This is getting real close!  Can you put the logging.exception back in the github bot?  You should be able to mock it in the tests.

zulip_bots/zulip_bots/bots/github_detail/github_detail.py
```diff
@@ -52,7 +52,6 @@ def get_details_from_github(self, owner: str, repo: str, number: str) -> Union[N
             r = requests.get(
                 self.GITHUB_ISSUE_URL_TEMPLATE.format(owner=owner, repo=repo, id=number))
         except requests.exceptions.RequestException as e:
-            logging.exception(str(e))
```
DhruvThakker: Add : logging.exception

showell:
Merged!  Thanks @DhruvThakker!

Issue #329, opened 2018-02-27
Make game handler bots take shorthand moves.  (e.g. connect four)


showell:
It's kind of tedious to say "move 7" when you're playing connect four.  We should probably allow just "7" as the command, and the game handler can have an option to assume numbers correspond to moves.

DhruvThakker:
@zulipbot claim

DhruvThakker:
@zulipbot abandon

showell:
I merged some changes for this, so closing.

PR #340, opened 2018-03-09
game_handler : Add shorthand moves

DhruvThakker:
#329 
Please review I did a small change to connect four to allow shorthand moves with regex. Should I do this for other game bots and are tests required for this?


showell:
@DhruvThakker Merged, thanks!  We probably want to add some test coverage here soon.  Feel free to run these bots against chat.zulip.org to have folks try out the new feature.

DhruvThakker:
We tried it out we found an error with move 0 in both games. But it is not related to this update.

showell:
@DhruvThakker Yeah, the individual games will always need to validate moves beyond just the regex.

PR #342, opened 2018-03-11
bots/flock: Add flock bot.

Rishabh570:


zulip_bots/zulip_bots/bots/flock/flock.py
```diff
@@ -0,0 +1,100 @@
...
+
+help_message = '''
+You can send messages to any Flock user associated with your account from Zulip.
+*Syntax*: **@botname to: message** where `to` is **firstName** of recipient.
+'''
+
+# Matches the recipient name provided by user with list of users in his contacts.
+# If matches, returns the matched User's ID
+def find_recipient(res: str, to: str) -> str:
+    for i in range(0, len(res)):
```
showell: You can say this:

~~~ py
for obj in res:
    if to == obj['firstName']
        return res['id']
~~~

It will implicitly return `None`, which you can check in the caller.

zulip_bots/zulip_bots/bots/flock/flock.py
```diff
@@ -0,0 +1,100 @@
...
+    try:
+        res = requests.get(USERS_LIST_URL, params=payload)
+    except ConnectionError as e:
+        logging.exception(str(e))
+        return "Uh-Oh, couldn't process the request \
+right now.\nPlease try again later"
+
+    res = res.json()
+    to = find_recipient(res, to)
+    if to == "notFound":
```
showell: just use `None` as the return value for failed searches

showell:
Looks pretty good!  I made one comment.

Rishabh570:
Updated.

showell:
Merged, thanks @Rishabh570!

PR #346, opened 2018-03-15
game_handler: Use avatars instead of emails in game handler messages.

Rishabh570:
The cases I've seen & replaced the occurrence of user email with user avatar are:

* when a player wins,

win

* when a player quits the game, and

quit

* when a player joins the game.

join

* telling a user about his turn

turn

Please let me know if I've missed any case...


showell:
Merged!  Thanks @Rishabh570!  A good follow up would be to extract a small function to make the avatars, just because the syntax there is subject to change, and that's probably an easy detail to encapsulate.

PR #347, opened 2018-03-15
[WIP] connectfour/controller: Prevent list access through negative indexing.

### LINKED ISSUE #343: Issue with move 0 in connect_four and TicTacToe
```move 0``` input should give invalid move in both these games but this command currently leads to a valid move.

Rishabh570:
Currently, if a user does 'move 0', he moves to column 6(last column)
in connect_four which is unwanted behaviour and happening due to
list getting accessed through negative index (-1).

Added a check for that in controller.py file.

Fixes #343


showell:
Merged, thanks @Rishabh570!

PR #349, opened 2018-03-17
tools/deploy: Annotate tools/deploy using mypy

Shivamgera:
Issue [#338](https://github.com/zulip/python-zulip-api/issues/338).
@aero31aero, please review this.

PR #353, opened 2018-03-23
mypy: Support files without extensions for scripts in /tools.

aero31aero:
This assumes that files passed in force_include without any extension are meant to be run with mypy, and thus assumed to be python scripts.

Also, we ignore the most of the tools dir, except for the tools/deploy script for which #349 adds type annotations. Unless that PR is merged and rebased onto this, this one will fail the Travis build.


showell:
merged, thanks @aero31aero!

PR #355, opened 2018-03-25
[WIP] Add Trello integration script 

Ljz7:


showell:
This LGTM.  Can you add some more info to the commit message (after the first line) describing what's going on here?

Ljz7:
@showell 

I added a small description to the commit message.

showell:
Can you make the commit message a little more detailed (and maybe code comments too)?  Am I reading the code correctly that you're sending a message to the Trello API, but you're setting up a callback so that the response actually goes to the server?  This is a bit of an unusual approach, which isn't to say it's wrong, it's just worth noting how it works.

Ljz7:
@showell 

Thanks for your feedback. 

The script follows the [documentation of Zulip about Trello integration](https://zulipchat.com/integrations/doc/trello). 

It creates the zulip callback and it builds and sends requests to the Trello API to create Webhooks.

Let me know if some part of the code are still too broad after reading the Zulip documentation about Trello Integration.

I'll prepare a new commit with more code comments tomorrow.

showell:
@Ljz7 I went ahead and merged it, but any additional documentation you can add would be great.  cc @eeshangarg

Issue #356, opened 2018-03-26
Automatically add a bot handler's parent dir to the python path


noemartinez:
The existing ```zulip-run-bot``` support for running a python file directly (instead of an existing bot in zulip_bots by name) works pretty well:

``` bash
zulip-run-bot  <path_to_handler_py_file> --config-file <path_to_zuliprc>
```

The only small inconvenience is that by default the bot can't import sibling python packages in the same directory as the bot's handler file that is being run. Could we automatically add the handler's parent dir to python's path as part of the ```zulip-run-bot``` process so the bot automatically has access to modules and packages deployed alongside it?

See [this conversation](https://chat.zulip.org/#narrow/stream/127-integrations/subject/standalone.20bot.20code/near/547688) for more context on this issue.

novokrest:
@timabbott I added another PR #382 to fix this issue. It is ready for review.

PR #361, opened 2018-03-30
Automatically add a bot handler's parent dir to the python path.

Anupam-dagar:
Fixes: #356 

I tested this in two ways:
1. Copied `helloworld` bot outside `python-zulip-api` in my local repo. Created one module in helloworld directory and imported in `helloworld.py`.
ran the bot using `zulip-run-bot path_to_helloworld.py -c path_to_zuliprc`
2. For the helloworld bot residing in zulip repository, created a module in helloworld directory and imported in `helloworld.py`.
ran the bot using `zulip-run-bot helloworld -c path_to_zuliprc`
This also makes it possible to import modules in subdirectories.

### DIFF
--- a/zulip_bots/zulip_bots/run.py
+++ b/zulip_bots/zulip_bots/run.py
@@ -112,10 +112,13 @@ def main():
     args = parse_args()
     if os.path.isfile(args.bot):
         bot_path = os.path.abspath(args.bot)
+        bot_dir = os.path.dirname(bot_path)
         bot_name = os.path.splitext(basename(bot_path))[0]
     else:
         bot_path = os.path.abspath(os.path.join(current_dir, 'bots', args.bot, args.bot+'.py'))
+        bot_dir = os.path.dirname(bot_path)
         bot_name = args.bot
+    sys.path.insert(0, bot_dir)
     if args.provision:
         provision_bot(os.path.dirname(bot_path), args.force)
 
--- a/zulip_bots/zulip_bots/test_run.py
+++ b/zulip_bots/zulip_bots/test_run.py
@@ -14,6 +14,7 @@ from unittest import TestCase
 
 from unittest import mock
 from unittest.mock import patch
+import sys
 
 class TestDefaultArguments(TestCase):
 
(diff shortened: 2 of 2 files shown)


zulip_bots/zulip_bots/run.py
```diff
@@ -112,9 +112,13 @@ def main():
     args = parse_args()
     if os.path.isfile(args.bot):
         bot_path = os.path.abspath(args.bot)
+        bot_dir = os.path.dirname(bot_path)
+        sys.path.insert(0, bot_dir)
         bot_name = os.path.splitext(basename(bot_path))[0]
     else:
         bot_path = os.path.abspath(os.path.join(current_dir, 'bots', args.bot, args.bot+'.py'))
+        bot_dir = os.path.join(current_dir, 'bots', args.bot)
+        sys.path.insert(0, bot_dir)
```
timabbott: Do we need this in both code paths?  

Might be cleaner to do just `sys.path.insert(os.path.dirname(bot_dir))` ourside the if/else block.

timabbott:
I think manual testing from multiple people might be good enough for merging this, but if we can extract that block of code to a function and add some unit tests for it, that'd be definitely better, and probably not hard.

zulip_bots/zulip_bots/run.py
Anupam-dagar: yes, I first tried it by keeping it outside only but that works only for the second case mentioned in the comment. If the user provides path to bot_handler.py then line in if code path is required.

Anupam-dagar: my apologies, I got your point, I misinterpreted it earlier. I'll do it :smile:

timabbott: OK.  We can move the `sys.path.insert` part out, though; just not `bot_dir`.  We might also want to do something with these redundant-sounding names; I think `bot_path` should probably become `bot_module_path` in a follow-up commit.

Anupam-dagar:
@timabbott I did the requested change. I'll take a look at writing tests now for it.

Anupam-dagar:
@timabbott this is ready for a review

zulip_bots/zulip_bots/test_run.py
```diff
@@ -47,6 +48,15 @@ def test_argument_parsing_with_bot_path(self, mock_run_message_handler_for_bot):
             lib_module=mock.ANY,
             quiet=False)
 
+    @patch('sys.argv', ['zulip-run-bot', path_to_bot, '--config-file', '/foo/bar/baz.conf'])
+    @patch('zulip_bots.run.run_message_handler_for_bot')
+    def test_module_import(self, mock_run_message_handler_for_bot):
+        # type: (mock.Mock) -> None
+        path_to_bots_dir = os.path.dirname(os.path.dirname(self.path_to_bot))
+        sys.path.insert(0, path_to_bots_dir)
+        import helloworld
```
timabbott: Why is this `import helloworld`?  And why dod you edit `sys.path` here?  I don't understand how this test validates what we care about, which is the logic in the above code.

neiljp:
@Anupam-dagar Would you like to work more on this?

Anupam-dagar:
@neiljp yes, I'll continue on this.

novokrest:
@Anupam-dagar Would you like to work more on this?

Anupam-dagar:
@novokrest I will continue this after 7th May as my exams are going on currently.

timabbott:
@novokrest if you have time now, I think it'd be pretty reasonable for you to take this over the finish line.

novokrest:
@timabbott I think this PR could be closed because issue was already resolved

timabbott:
Agreed, closing, thanks for your work on this @Anupam-dagar!

PR #362, opened 2018-03-30
API: Use `session.headers.update` method to add user-agent header.

amanagr:
Directly assigning session.headers overrides the default header
paramers and removes `{'Accept-Encoding': 'gzip, deflate'}` from
the headers.


amanagr:
@timabbott FYI

timabbott:
Nice, merged!

PR #363, opened 2018-03-31
Mypy.ini migration & simplification

neiljp:
There have been a few changes In the main Zulip repo, which feel reasonable to replicate here:
* migration to use a `mypy.ini` file, allowing finer-grained control of settings applying to different files;
* removal of the strict-optional setting - it's on by default (and can be controlled per-file in mypy.ini);
* addition of `warn_no_return` and `no_implicit_optional` settings for further strictness.

The latter setting requires some annotation changes and affects situations such as:
```python
def f(x: str=None) -> None:
   ...
```
These pass without this latter setting; now a more explicit `Optional[str]` is necessary, as might be expected.

The mypy.ini commit has a line which shows tracebacks on fatal error, and is included in the main repo; should we include it here?

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -2,6 +2,7 @@
 check_untyped_defs = True
 disallow_any_generics = True
 strict_optional = True
+no_implicit_optional = True
 
 incremental = True
 scripts_are_modules = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,6 +1,7 @@
 [mypy]
 check_untyped_defs = True
 disallow_any_generics = True
+strict_optional = True
 
 incremental = True
 scripts_are_modules = True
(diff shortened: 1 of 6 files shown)


timabbott:
@neiljp can you rebase to address the merge conflicts with your other changes?

neiljp:
@timabbott Rebased and squashed together the two commits in prep for no_implicit_optional, which could be combined with the enforce commit too.

mypy.ini
```diff
@@ -0,0 +1,7 @@
+[mypy]
+check_untyped_defs = True
+disallow_any_generics = True
+
+incremental = True
+scripts_are_modules = True
+#show_traceback = True  # Used in Zulip core; use here?
```
timabbott: We should, I think; it's mostly an option that helps with debugging if mypy crashes.

timabbott:
Posted one comment, otherwise lgtm.

neiljp:
@timabbott: Amended traceback option & squashed those commits together.

showell:
Merged, thanks @neiljp!

Issue #338, opened 2018-03-09
Task list for tools/deploy.


aero31aero:
This issue tracks the initial tasks with the tools/deploy script, of which some would tie in to adding further functionality to the server. `tools/deploy` itself is stable, however the corresponding server is being worked on right now and is pretty buggy. Usage instructions are on #337 right now. 

- [ ] Annotate `tools/deploy` using mypy.
- [ ] Document `tools/deploy` usage. This should be done once the system is stable enough for public usage.
- [ ] Add `tools/deploy delete` to cleanly stop and delete the docker container and image, as well as the source files for the image.
- [ ] Add `tools/deploy log <bot-name>` to show the log of the bot.
- [ ] Add `tools/deploy ls` to show a overall status (names, running/stopped) of a user's bots.

aero31aero:
Better instructions:
https://gist.github.com/aero31aero/4a9e43641f047d983adb1e9a18e12f01

These should be used to prepare the final docs for the tools/deploy script.

PR #337, opened 2018-03-08
[WIP] tools/deploy script for interfacing with the bot server.

aero31aero:
Sample session:

```
$ cd  ~/Documents/gitclones/python-zulip-api
$ source zulip-api-py3-venv/bin/activate
$ tools/deploy prepare c-four -p zulip_bots/zulip_bots/bots/connect_four -c ~/Documents/zulip/tworc -m connect_four.py -s $SERVER -k $KEY
pack: Created zip file at: .bots/c-four.zip.
upload: Uploaded the bot package to botfarm.
clean: Removed .bots/c-four.zip.
process: The bot has been processed by the botfarm.
$ tools/deploy prepare hello -p zulip_bots/zulip_bots/bots/helloworld -c ~/Documents/temp/zulip/onerc -m helloworld.py -s $SERVER -k $KEY
pack: Created zip file at: .bots/hello.zip.
upload: Uploaded the bot package to botfarm.
clean: Removed .bots/hello.zip.
process: The bot has been processed by the botfarm.
$ tools/deploy --server $SERVER --key $KEY stop test
stop: The bot has been stopped by the botfarm.
$ tools/deploy --server $SERVER --key $KEY start hello
start: The bot has been started by the botfarm.
$ tools/deploy --server $SERVER --key $KEY start c-four
start: The bot has been started by the botfarm.
```

PR #366, opened 2018-04-01
tools/deploy : Add logs endpoint to receive logs for the specified bot.

Anupam-dagar:
Fixes one of the tasks listed in [338](https://github.com/zulip/python-zulip-api/issues/338)
Steps to run:
1. Assuming a bot is running on the botfarm with name helloworld. (check [337](https://github.com/zulip/python-zulip-api/pull/337) for more details.)
2. Run the command `tools/deploy -s $SERVER -k $KEY log helloworld`
Sample Traceback:
```
Running Helloworld Bot:

        This is a boilerplate bot that responds to a user query with
        "beep boop", which is robot for "Hello World".

        This bot can be used as a template for other, more
        sophisticated, bots.
        
INFO:root:starting message handling...

```
4. To get top n number of lines as response, run the following command:
`tools/deploy -s $SERVER -k $KEY log helloworld --lines 4`
Sample Traceback:
```
Running Helloworld Bot:

        This is a boilerplate bot that responds to a user query with
        "beep boop", which is robot for "Hello World".

```

### DIFF
--- a/tools/deploy
+++ b/tools/deploy
@@ -146,6 +146,25 @@ def prepare(options: argparse.Namespace) -> None:
     clean(options)
     process(options)
 
+def log(options: argparse.Namespace) -> None:
+    check_common_options(options)
+    headers = {'key': options.key}
+    if options.lines:
+        lines = options.lines
+    else:
+        lines = None
+    payload = {'name': options.botname, 'lines': lines}
+    url = urllib.parse.urljoin(options.server, 'bots/logs/' + options.botname)
+    r = requests.get(url, json=payload, headers=headers)
+    if r.status_code == requests.codes.ok:
+        print(r.text)
+        return
+    if r.status_code == 401:
+        print('log: Authentication error with the server. Aborting.')
+    else:
+        print('log: Error {}: {}. Aborting.'.format(r.status_code, r.text))
+    sys.exit(1)
+
 def main() -> None:
     usage = """tools/deploy <command> <bot-name> [options]
 
@@ -165,6 +184,9 @@ To stop the bot, use:
 
     tools/deploy stop mybot --server=$SERVER --key=$TOKEN
 
+To get logs of the bot, use:
+    tools/deploy log mybot --serverr=$SERVER --key=$TOKEN
+
 """
     parser = argparse.ArgumentParser(usage=usage)
     parser.add_argument('command', help='Command to run.')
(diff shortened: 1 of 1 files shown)


tools/deploy
```diff
@@ -146,6 +146,25 @@ def prepare(options: argparse.Namespace) -> None:
...
+    else:
+        lines = None
+    payload = {'name': options.botname, 'lines': lines}
+    url = urllib.parse.urljoin(options.server, 'bots/logs/' + options.botname)
+    r = requests.get(url, json=payload, headers=headers)
+    if r.status_code == requests.codes.ok:
+        print(r.text)
+        return
+    if r.status_code == 401:
+        print('stop: Authentication error with the server. Aborting.')
```
aero31aero: So far I was following the pattern of having the function name be logged with the messages in this script, so if someone simply gives us the output of this script if it fails, we could tell in which specific function it failed.

Though that'd be useful for chained commands like `prepare`, but not much relevant here except for consistency. Thoughts?

Anupam-dagar: My bad, I forgot to change it. Mentioning function name is good for debugging purpose for users.Since we are not going to put logs in prepare command, we can get away with the function name and simply show the error but it will be helpful if we mention the function name.
I'll change stop to logs.

aero31aero:
Hey! I'm not particularly a fan of too long error messages when a short and direct text would work, though I'll wait for someone else to do that. Still, both of these changes (adding log command and changing print statements) should be done in separate commits.

showell:
@timabbott Can you weigh in on this?

The first commit seems reasonable to me, but I don't have all the context of why we're sending bot logs to the server.  It's something that I've never felt the need for, but that could just be a function of my workflow.

For the second commit, like Rohitt, I'm also a little skeptical that the longer messages are necessary, or we can at least try to handle that in a more centralized fashion.

aero31aero:
@showell we're doing the reverse: getting the logs from the docker botserver running your bots.

showell:
Oh, ok.  For the second commit, we should definitely extract something that wraps `print` and call it something like `error`, and then have the latter prepend any helpful messages about where to find the logs, etc.

aero31aero:
@Anupam-dagar thanks for working on this. I merged your fist commit after fixing a small typo (`s/serverr/server/`). Let's figure out the error situation next.

I believe that we do need a function to print the errors, as @showell also pointed out. That should ideally be the next step, and then we can make a one line change for switching to the language you prefer for the messages.

Closing this PR.

PR #369, opened 2018-04-09
Bots: Support run-time bot identity (update of #47)

neiljp:
This PR is an update of #47.

Main changes:
* refactored against current master & changes since original PR;
* splitting/merging commits to more logical order;
* namedtuple -> class for BotIdentity, as quoted_name/mention is calculable;
* moved to `mention` from `quoted_name`, as former is simpler.

Feedback sought:
* Where should BotIdentity go? The current imports in `simple_lib.py` and `test_lib.py` seem off?;
* Should more bots be updated as part of this PR, or are two sufficient for examples?

### DIFF
--- a/zulip_bots/zulip_bots/bots/wikipedia/test_wikipedia.py
+++ b/zulip_bots/zulip_bots/bots/wikipedia/test_wikipedia.py
@@ -52,7 +52,7 @@ class TestWikipediaBot(BotTestCase):
 
         # Empty query, no request made to the Internet.
         bot_request = ''
-        bot_response = "Please enter your search term after @mention-bot"
+        bot_response = "Please enter your search term after @**test-bot**"
         self.verify_reply(bot_request, bot_response)
 
         # Incorrect status code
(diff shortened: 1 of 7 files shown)


showell:
@neiljp This LGTM, I'm happy to merge it once you fix the conflicts (cc @timabbott).

If you want to update a third place, I'd look at how the connectfour bot notifies people of whose turn it is, who wins, etc.  Ideally all that's in game_handler, so we'd get a couple bots using this in a meaningful way.

neiljp:
@showell Pushed with a few tweaks given the typing is now py3 style.

showell:
It feels a little funny to have to say bot_handler.identity().mention.  I would expect something more like `bot_handler.identity.mention` or `bot_handler.identity().mention()` or even `bot_handler.mention()`, depending on what your philosophy is about what should be a function.  The `identity()` seems kind of funny to me, since `identity` feels like an attribute, not something you compute.

Having said all that, I'm gonna merge this, so people can play around with it, and we can always sweep this later if other folks share my opinion on this.

showell:
merged, thanks @neiljp!

PR #371, opened 2018-04-15
tools: Call pip from a sub-process instead of importing it

### LINKED ISSUE #370: Provisioning doesn't work with the latest version of `pip`
Provisioning fails while setting up the package `zulip_bots` with an `AttributeError` for `pip.main` with the latest version (`10.0.0`) of `pip`.

punchagan:
The pip documentation recommends calling pip using a subprocess, instead of
importing it and using it's internal API. The API of pip==10.0.0 is different
from that of older versions, and provisioning is broken with this version.

https://pip.pypa.io/en/stable/user_guide/#using-pip-from-your-program

Closes #370

### DIFF
--- a/zulip_bots/setup.py
+++ b/zulip_bots/setup.py
@@ -6,7 +6,7 @@ from __future__ import print_function
 import os
 import sys
 import glob
-import pip
+import subprocess
 if False:
     from typing import Any, Dict, Optional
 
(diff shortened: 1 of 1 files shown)


timabbott:
lgtm

@eeshangarg can you verify the fix?  We should do an immediate release and the upgrade the Zulip server project to use the new release of this project.

timabbott:
I chatted with Eeshan and he confirmed the fix, so merged.  Thanks @punchagan!

PR #372, opened 2018-04-16
test_tictactoe.py improvements

kishanlp:
Slightly increased coverage for tictactoe.py from 19% to 20%. Will need further review to test the integration of different functions within the tictactoe bot, as many of the functions seem to not be able to be tested due to bad function calls within the bot code. Additionally, we found that the current tests for the TicTacToeModel object could not be run without explicitly defining current_board outside of the constructor, so we fixed this in order to run our test cases.

### DIFF
--- a/tools/provision
+++ b/tools/provision
@@ -83,7 +83,7 @@ the Python version this command is executed with."""
     def install_dependencies(requirements_filename):
         pip_path = os.path.join(venv_dir, venv_exec_dir, 'pip')
         # We first install a modern version of pip that supports --prefix
-        subprocess.call([pip_path, 'install', 'pip>=9.0'])
+        subprocess.call([pip_path, 'install', 'pip==9.0.3'])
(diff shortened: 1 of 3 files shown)


showell:
Hi @kishanlp are you still planning to finish this up?  If not, that's not a problem, we can have somebody else try to clean up the commits a bit.

showell:
@novokrest Do you have any interest in taking over this PR?  The test code looks pretty reasonable, and it's possible that you can mostly squash the commits together into one big commit.  (For adding tests, we don't need to be super granular.)  There's one change to the actual code that I'm not sure is necessary.

novokrest:
@showell I prepared PR based on current: #443

showell:
Closing this.  @kishanlp Thanks for the work on this--we merged it via #443.  @novokrest Thanks for taking it to the finish line.

PR #374, opened 2018-04-28
#359 Make StateHandler lazy and only do `get` one key at a time

### LINKED ISSUE #359: bots: Make StateHandler lazy and only do `get` one key at a time.
We should not call `self._client.get_storage()` in `StateHandler.__init__`.  Instead, for each call to `StateHandler.get()`, we should call the API for just the key we need.

We should consider eliminating `self._state` altogether, or at least just use it for keys that we have already gotten.

novokrest:
Change StateHandler implementation to the following:
- do not fetch state in `__init__()`;
- fetch every key only once on demand and cache it.

Fixes #359

### DIFF
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -65,11 +65,7 @@ class StateHandler(object):
         self._client = client
         self.marshal = lambda obj: json.dumps(obj)
         self.demarshal = lambda obj: json.loads(obj)
-        response = self._client.get_storage()
-        if response['result'] == 'success':
-            self.state_ = response['storage']
-        else:
-            raise StateHandlerError("Error initializing state: {}".format(str(response)))
+        self.state_ = dict()  # type: Dict[Text, Any]
 
     def put(self, key, value):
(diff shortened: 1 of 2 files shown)


timabbott:
I think the "only fetch on demand" piece is probably going to add extra round trips for anything that does more than one request; is there a good reason to do that change?

timabbott:
@showell I guess we should discuss, since you were the one to suggest doing each key on demand.  My theory is that doing so could carry a significant perf cost, but I guess it depends on how actual bots structure their keys.

neiljp:
@showell Is the motivation for this related to having no guarantee that a bot stays alive between calls?

showell:
I think there a couple factors at play here:
* we definitely don't want to fetch all keys at once (imagine a game bot that is around for a year and makes new keys for each game)
* we want to be sure to handle bots dying and restarting
* we should expect bots to do proper key management, like keep their data kind of coarse per key, since the value for a key could be arbitrarily deep

showell:
Also the code here LGTM so I would want to merge it my prior comments make sense.

PR #375, opened 2018-04-30
Upgrade to mypy 0.590 & distribute py.typed file for API

neiljp:
These are pretty tiny commits, but are separate issues:
* the mypy upgrade had no issues as far as I could tell, but seemed good to isolate;
* the `py.typed` file and installation seems relevant to the API package.

@eeshangarg This seems to be a reasonable minimal packaging change, but I'm not as familiar with the packaging intricacies as you :) Do you have a recommended test method for this?


timabbott:
@neiljp I think one way to test the packaging would be to upgrade zulip.git repo to use it; we have a bit of code that depends on the `python-zulip-api` module, and we could see if it type-checks.

timabbott:
I think it's safe enough to merge this; that'll make it easier to do the testing.

PR #378, opened 2018-05-05
#351 Add file upload access to the bot API

### LINKED ISSUE #351: Add file upload access to the bot API
A full-fledged Client python object has [```upload_file()```](https://github.com/zulip/python-zulip-api/blob/master/zulip/zulip/__init__.py#L665), but [a bot's API exposed via bot_handler](https://zulipchat.com/api/writing-bots#bot-api) doesn't currently include that method. 

While there are non-ideal ways to access that method from a bot's ```handle_message```, it would be nice to have it properly promoted up to the public ```bot_handler``` API. Could we add it?

Uploading files is handy for bots in charge of generating and conveying back imagery or other types of small files to the caller.

novokrest:
Fixes #351.

Added methods ExternalBotHandler.upload_file() and ExternalBotHandler.upload_file_from_path() to upload local files to Zulip server. I decided to allow user to upload any file located on his machine, not only inside bot's local directory. 

P.S. Also I added commit c7acf50 with `file_uploader` bot as an example of using new methods for uploading files. If it is unnecessary I will revert this commit.

image

### DIFF
--- a/zulip_bots/zulip_bots/bots/file_uploader/doc.md
+++ b/zulip_bots/zulip_bots/bots/file_uploader/doc.md
@@ -1,0 +1,21 @@
+# File Uploader Bot
+
+This bot allows the user to upload a file with a given path to the Zulip server.
+
+## Usage
+
+Use this bot with any of the following commands:
+
+- `@uploader <local_file_path>` : Upload a file, where `<local_file_path>` is the path to the file
+- `@uploader help` : Display help message
+
+### Usage examples
+
+The following command will upload the file `/tmp/image.png` to the Zulip server:
+```
+@**uploader** /tmp/image.png
+```
+
+Here's an example response:
+
+> [image.png](https://server.zulipchat.com/user_uploads/3787/RgoZReSsfMjlQSzvVxjIgAQy/image.png)
--- a/zulip_bots/zulip_bots/bots/file_uploader/file_uploader.py
+++ b/zulip_bots/zulip_bots/bots/file_uploader/file_uploader.py
@@ -1,0 +1,41 @@
+from typing import Any, Dict
+
+import os
+from pathlib import Path
+
+class FileUploaderHandler(object):
+    def usage(self) -> str:
+        return (
+            'This interactive bot is used to upload files (such as images) to the Zulip server:'
+            '\n- @uploader <local_file_path> : Upload a file, where <local_file_path> is the path to the file'
+            '\n- @uploader help : Display help message'
+        )
+
+    def handle_message(self, message: Dict[str, str], bot_handler: Any) -> None:
+        HELP_STR = (
+            'Use this bot with any of the following commands:'
+            '\n* `@uploader <local_file_path>` : Upload a file, where `<local_file_path>` is the path to the file'
+            '\n* `@uploader help` : Display help message'
+        )
+
+        content = message['content'].strip()
+        if content == 'help':
+            bot_handler.send_reply(message, HELP_STR)
+            return
+
+        path = Path(os.path.expanduser(content))
+        if not path.is_file():
+            bot_handler.send_reply(message, 'File `{}` not found'.format(content))
+            return
+
+        path = path.resolve()
+        upload = bot_handler.upload_file_from_path(str(path))
+        if upload['result'] != 'success':
+            msg = upload['msg']
+            bot_handler.send_reply(message, 'Failed to upload `{}` file: {}'.format(path, msg))
+            return
+
+        uploaded_file_reply = '[{}]({})'.format(path.name, upload['uri'])
+        bot_handler.send_reply(message, uploaded_file_reply)
+
+handler_class = FileUploaderHandler
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -211,6 +211,15 @@ class ExternalBotHandler(object):
 
         return dict(config_parser.items(bot_name))
 
+    def upload_file_from_path(self, file_path: str) -> Dict[str, Any]:
+        with open(file_path, 'rb') as file:
+            return self.upload_file(file)
+
+    def upload_file(self, file: IO[Any]) -> Dict[str, Any]:
+        if not self._rate_limit.is_legal():
+            self._rate_limit.show_error_and_exit()
+        return self._client.upload_file(file)
+
     def open(self, filepath: str) -> IO[str]:
         filepath = os.path.normpath(filepath)
         abs_filepath = os.path.join(self._root_dir, filepath)
--- a/zulip_bots/zulip_bots/bots/file_uploader/test_file_uploader.py
+++ b/zulip_bots/zulip_bots/bots/file_uploader/test_file_uploader.py
@@ -1,0 +1,39 @@
+from unittest.mock import patch, MagicMock, Mock
+
+from zulip_bots.test_lib import (
+    get_bot_message_handler,
+    StubBotHandler,
+    BotTestCase,
+    DefaultTests,
+)
+
+from pathlib import Path
+
+class TestFileUploaderBot(BotTestCase, DefaultTests):
+    bot_name = "file_uploader"
+
+    @patch('pathlib.Path.is_file', return_value=False)
+    def test_file_not_found(self, is_file: Mock) -> None:
+        self.verify_reply('file.txt', 'File `file.txt` not found')
+
+    @patch('pathlib.Path.resolve', return_value=Path('/file.txt'))
+    @patch('pathlib.Path.is_file', return_value=True)
+    def test_file_upload_failed(self, is_file: Mock, resolve: Mock) -> None:
+        server_reply = dict(result='', msg='error')
+        with patch('zulip_bots.test_lib.StubBotHandler.upload_file_from_path',
+                   return_value=server_reply) as m:
+            self.verify_reply('file.txt', 'Failed to upload `/file.txt` file: error')
+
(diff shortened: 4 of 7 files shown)


zulip_bots/zulip_bots/lib_tests.py
```diff
@@ -138,3 +143,36 @@ def test_message(message, flags):
                                         config_file=None,
                                         bot_config_file=None,
                                         bot_name='testbot')
+
```
neiljp: I would consider refactoring the following two test functions to use a helper, given the duplication.

zulip_bots/zulip_bots/bots/file_uploader/file_uploader.py
```diff
@@ -0,0 +1,27 @@
+from typing import Any
+
+from pathlib import Path
+
+class FileUploaderHanlder(object):
```
neiljp: Is `FileUploaderHandler` the intended name?

zulip_bots/zulip_bots/bots/file_uploader/file_uploader.py
```diff
@@ -0,0 +1,27 @@
...
+    def usage(self) -> str:
+        return (
+            'This interactive bot used to upload files, images, etc to Zulip server.\n'
+            'Specify the file path you want to upload and file will be upload to Zulip server'
+        )
+
+    def handle_message(self, message: Any, bot_handler: Any) -> None:
```
neiljp: The type of `message` can be more specific than `Any`, though I'm aware that the typing of many bots in the repo is quite loose in this way; that said, I'm not convinced the bot will be type-checked automatically at the moment, since you don't have a diff to `tools/run-mypy` to explicitly check it?

zulip_bots/zulip_bots/bots/file_uploader/file_uploader.py
```diff
@@ -0,0 +1,27 @@
...
+
+        upload = bot_handler.upload_file_from_path(str(path.resolve()))
+        if upload['result'] != 'success':
+            msg = upload['msg']
+            bot_handler.send_reply(message, 'Failed to upload file: {}'.format(msg))
+            return
+
+        uploaded_file_reply = '[{}]({})'.format(path.name, upload['uri'])
+        bot_handler.send_reply(message, uploaded_file_reply)
+
```
neiljp: There is no `help` command; while this is a very minimal bot, this might be useful to include :)

zulip_bots/zulip_bots/bots/file_uploader/doc.md
```diff
@@ -0,0 +1,3 @@
+# File Uploader Bot
+
+This bot allows user to upload file on given path to Zulip server.
```
neiljp: This is a simple bot, but this seems a little short; perhaps include usage instructions here, as per other `doc.md` files, and maybe discuss limitations of uploading, if there are any?

zulip_bots/zulip_bots/bots/file_uploader/file_uploader.py
```diff
@@ -0,0 +1,27 @@
...
+    def usage(self) -> str:
+        return (
+            'This interactive bot used to upload files, images, etc to Zulip server.\n'
+            'Specify the file path you want to upload and file will be upload to Zulip server'
+        )
+
+    def handle_message(self, message: Any, bot_handler: Any) -> None:
+        path = Path(message['content'].strip())
+        if not path.is_file():
+            bot_handler.send_reply(message, 'File not found')
```
neiljp: Consider adding the filename here, as in the failed-upload case, so the user can be sure of what path failed? Also perhaps in the case of multiple upload commands overlapping.

zulip_bots/zulip_bots/lib_tests.py
```diff
@@ -138,3 +143,36 @@ def test_message(message, flags):
...
+        client.upload_file = MagicMock()
+
+        handler = ExternalBotHandler(
+            client=client,
+            root_dir=None,
+            bot_details=None,
+            bot_config_file=None
+        )
+
+        file = io.StringIO('text')
```
neiljp: Should this be `BytesIO`? (see my other query re text vs binary)

zulip_bots/zulip_bots/lib.py
```diff
@@ -158,6 +158,18 @@ def update_message(self, message):
         else:
             self._rate_limit.show_error_and_exit()
 
+    def upload_file_from_path(self, file_path):
+        # type: (str) -> Dict[str, Any]
+        with open(file_path, 'rb') as file:
+            return self.upload_file(file)
+
+    def upload_file(self, file):
+        # type: (IO[Any]) -> Dict[str, Any]
```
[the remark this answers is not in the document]
novokrest: I think we can require (from user of this method) to open the passed file in binary mode, but it looks like an artificial restriction. Eventually, the opened files are sent by `requests.Session.request`, and, according to the [documentation](http://docs.python-requests.org/en/master/user/quickstart/#post-a-multipart-encoded-file), the `files` parameter can represented by files openned in both binary mode or text mode. It is worth noting that there is a [recommendation](http://docs.python-requests.org/en/master/user/quickstart/#post-a-multipart-encoded-file) to open files in binary mode. As an another possible solution we can add two functions `upload_binary_file()` and `upload_text_file()` to use more precise types in `# type` comments but dynamic nature of python still allows to pass both 'kind' of files to each function. But in this case we also need to implement bot 'upload_binary_file_from_path' and 'upload_text_file_from_path' methods. 
In my opinion, we can restrict user to pass only binary-open files (despite this solution is less flexible). @neiljp How do you think?

zulip_bots/zulip_bots/bots/file_uploader/file_uploader.py
novokrest: Yes, inspired by `LinkShortenerHandler` :)

zulip_bots/zulip_bots/lib_tests.py
novokrest: Add `_create_client_and_handler_for_file_upload()` helper method

zulip_bots/zulip_bots/bots/file_uploader/file_uploader.py
novokrest: It seems to be a my mistake. Changed type to more precise `Dict[str, str]` as in another bots.

zulip_bots/zulip_bots/bots/file_uploader/file_uploader.py
novokrest: Added `help` command

zulip_bots/zulip_bots/bots/file_uploader/doc.md
novokrest: Added an example of using the File Uploader Bot in `doc.md`

zulip_bots/zulip_bots/bots/file_uploader/file_uploader.py
neiljp: Right - I was meaning whether you noticed the typo :)

zulip_bots/zulip_bots/bots/file_uploader/doc.md
neiljp: That's an improvement; the `doc.md` files for other bots show the typical structure in use - eg. subheadings and other information they include. As with coding, if you can follow the current (majority) style that is good practice :)

neiljp:
Just minor tweaks left; other than that LGTM @aero31aero.

neiljp:
Re the binary issue, I'd suggest leaving as-is unless there's further feedback.

zulip_bots/zulip_bots/bots/file_uploader/file_uploader.py
novokrest: I noticed now :) Corrected

zulip_bots/zulip_bots/bots/file_uploader/doc.md
novokrest: I updated `doc.md` (used `dropbox` bot's documentation as an example)

novokrest:
@neiljp @timabbott Could we merge this PR ?

timabbott:
@YagoGG @eeshangarg can one of you review this?

zulip_bots/zulip_bots/bots/file_uploader/file_uploader.py
novokrest: @neiljp changed message to `File {} not found`

zulip_bots/zulip_bots/lib_tests.py
novokrest: @neiljp sure! changed to `BytesIO`

zulip_bots/zulip_bots/bots/file_uploader/file_uploader.py
```diff
@@ -0,0 +1,38 @@
...
+            'Specify the file path you want to upload and file will be upload to Zulip server'
+        )
+
+    def handle_message(self, message: Dict[str, str], bot_handler: Any) -> None:
+        HELP_STR = (
+            'Mention the file uploader bot in a conversation and '
+            'then enter path of any file you want to upload in the body of '
+            'the message.'
+        )
+
```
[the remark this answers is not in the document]
novokrest: @eeshangarg Also I added resolving `~` in given file's path

eeshangarg:
@novokrest: Did you update this? :)

showell:
Merged, thanks @novokrest!

PR #380, opened 2018-05-07
[WIP] Upgrade mypy to 0.600 & add dmypy support

neiljp:
Upgrading to mypy 0.600 does not present any particular challenges in terms of updating annotations, etc.

However, it does incorporate support for running the checks persistently in a daemon (`dmypy`). Upon initial loading, the daemon can load from a cache generated by the regular process (`.mypy_cache`), or else starts from nothing. However, after that point, the 'incremental' checks are much faster. For comparison on my local machine:
* regular check, no file cache: 40-50s
* regular check, using file cache: 7-15s (ie. incremental mode)
* dmypy check, initial run: 23-67s
* dmypy check, subsequent: 1-3s
Of course, these are already relatively short in this small repo, so it's unclear how these might scale upwards.

The current commits integrate controlling dmypy into the `tools/run-mypy` script, since so much of the logic is already there.

One factor to consider is that from what I can tell `dmypy` runs with the options as specified, and each subsequent run with the same running daemon uses the same options - consequently since I don't believe this information is able to be interrogated from the dmypy process, I ensure that the specified options are synchronised using a simple json-formatted `dmypy.options`, written when (re)starting `dmypy`. This should probably be added to `.gitignore`, as well as the process detail files (dmypy.json, dmypy.sock).

In terms of speeding up the initial run of dmypy, the wide range of times listed above is since the options used in dmypy and the cache seem to also heavily influence whether the cache is used. Resolving this, and so reliably increasing any initial load time, is not something I've looked into at this point. The mypy docs refer to the idea of maintaining a repo of cache files for each commit/branch, which could tie into this, if we go in this direction.

A second potential outstanding issue, is that `dmypy` currently requires `--follow-imports=skip` or `=error`, while the zulip option is normally `=silent`. I've selected `=error` for the current implementation using `dmypy`, but to use the regular cache when running `dmypy` for the first time, the regular run would also need to switch to this, as per the WIP `sync run-mypy` commit. The user of `=error` has picked up a few issues, eg. annotations in `zulip/integrations`, but while the import style is a little non-standard I'd appreciate feedback on the salesforce change (other WIP). The `=error` option only appears to be an issue with showing too many 'notes' re imports, which don't appear on subsequent runs.

Whether to leave the daemon running indefinitely is a separate issue, but that will not be difficult to change later.

This should not be difficult to port across to the main repo at a later stage, but it seemed easier to prototype on the smaller (and local, to me) API repo.

Seeking feedback from anyone who's interested, but I chatted to @timabbott about this already, and @gnprice may have thoughts on the infrastructure too?


timabbott:
Quick note: I merged the mypy 0.600 changes; for the mypy daemon piece, we should wait until 0.610 comes out, which I think will make this a lot simpler.  If there are prep pieces we should merge, though, can you pull off the mypy daemon commit?

PR #382, opened 2018-05-09
#356 Automatically add a bot handler's parent dir to the python path 

### LINKED ISSUE #356: Automatically add a bot handler's parent dir to the python path

novokrest:
Fixes #356.

### DIFF
--- a/zulip_bots/zulip_bots/run.py
+++ b/zulip_bots/zulip_bots/run.py
@@ -116,6 +116,7 @@ def main():
     else:
         bot_path = os.path.abspath(os.path.join(current_dir, 'bots', args.bot, args.bot+'.py'))
         bot_name = args.bot
+    sys.path.insert(0, os.path.dirname(bot_path))
     if args.provision:
         provision_bot(os.path.dirname(bot_path), args.force)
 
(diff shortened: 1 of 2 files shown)


timabbott:
Looks great, merged, thanks @novokrest!  I think a follow-up to this would be to go through the flow of writing an out-of-tree bot (ideally, all the way through having a package with its own `Git` repository that one can `python setup.py install` or upload to `pypi`), taking careful notes on what one needs to do.  We'll use those notes to (1) add to our docs for writing bots and (2) use as a TODO list for things that aren't as simple as they should be, and we should fix.

PR #125, opened 2017-09-19
[WIP] bots: Improve input handling

neiljp:
This is intended to resolve #6, but also a few other case(s) that may exists with input handling.

So far:
* Wikipedia: URL handling
* Define: initially just avoiding symbols, but this is complex in python3; the API is unclear, but it likely only accepts ascii, so that's the limit I've put in now.

PR #384, opened 2018-05-12
#6 bots: Encode URLs properly to outbound sites.

### LINKED ISSUE #6: bots: Encode URLs properly to outbound sites.
See https://github.com/zulip/zulip/issues/5152.

novokrest:
Fixes #6.

The PR turned out rather big but all changes are very simple.

### How to encode
I made some research and it turned out that `requests` module already encodes query parameters correctly when query parameters are passed as dictionary `params` parameter in `requests.get|post|...` methods (the result of encoding path parts is same as `urllib.parse.quote` and the result of encoding query parameters is same as `urllib.parse.quote_plus`). 
So, when you call:
```python
requests.get('http://server.domain/api/foo?query=some#query')
```
then actual target of outgoing request will be: `http://server.domain/api/foo?query=some` (the fragment parts was truncated)
But when you call:
```python
requests.get('http://server.domain/api/foo', params=dict(query='some#query'))
```
then actual target of outgoing request will be: `http://server.domain/api/foo?query=some%23query` (this is what we want).
That's why bug in `wikipedia` bot could be avoided if query parameters were passed as `params` parameter in `requests.get()` method.

Also the following should be noted: if there is URL with format 'http://server.com/{part_id}/api' and `part_id` contains _sequence of characters_ matched with some encoded non-safe character e.g. `abc%25def` (very unusual case and, fortunately, our bots have not such problem) then `request` will not encode `%` repeatedly in this `part_id`, so we should manually encode it with `urllib.parse.quote` to make call to `http://server.com/abc%2525def/api`. Also if valid `part_id` contains character like `#` we also should manually encode it with `urllib.parse.quote` too (or part after `#` will be considered as fragment and will be truncated).

### Refactoring Strategy
I found all bots that use the `requests` module and change them with following rules:
- all query parameters should be passed as `params` parameter in `requests.get|post|...` methods;
- all path parts with possible non-safe characters should be encoded with `urllib.parse.quote` method. 
 
### Summary
There is summary reports with description why some changes were or were not made:

#### baremetrics bot
  - not use `urllib.parse.quote` for encoding `source_id` because it consists of `[a-zA-Z0-9]`

#### beeminder bot
  - not use `urllib.parse.quote` for encoding `username` and `goalname` appeared in URL path parts
    because username consists of only letters and number and goalname consists of `[a-z0-9-]`

#### define bot
  - was already updated in https://github.com/zulip/python-zulip-api/pull/125 (only ascii character are allowable)
  - note: currently `owlbot.info` supports only English dictionary but it can be expand (as noted in [README.md](https://github.com/payamnj/owlbot))

#### flock bot
  - all query params were already passed as `params` parameter to `requests.get()` (nothing was changed)

#### front bot
  - not use `urllib.parse.quote` for encoding URL path parts because `conversation_id` does not contain non-safe characters

#### giphy bot
  - all query params were already passed as `params` parameter to `requests.get()` (nothing was changed)

#### github_detail bot
  - not use `urllib.parse.quote` for encoding path parts because `owner`, `repo` contains only alphanumeric with hyphens and issue's `id` is numeric value

#### google_search bot
  - all query params were already passed as `params` parameter to `requests.get()` (nothing was changed)

#### google_translate bot
  - all query params were already passed as `params` parameter to `requests.get()` (nothing was changed)

#### idonethis bot
  - query params should not be encoded (because only `team_id` does not contain non-safe character) but I decide to pass query params as `params` to requests as a canonical way to pass parameter

#### jira bot
  - not use `urllib.parse.quote` for encoding path parts 

#### link_shortener bot
  - all query params were already passed as `params` parameter to `requests.get()` (nothing was changed)

#### mention bot
  - not use `urllib.parse.quote` for encoding path parts (only IDs are used)

#### monkeytestit bot (internal `extract` module)
  - all query params were already passed as `params` parameter to `requests.get()` (nothing was changed)

#### stack_overflow bot
  - use `params` parameter in `requests.get()` method for passing query parameters

#### susi bot
  - use `params` parameter in `requests.get()` method for passing query parameters

#### trello bot
  - not use `urllib.parse.quote` for encoding path parts (only IDs are used)
  - all query params were already passed as `params` parameter to `requests.get()` (nothing was changed)

#### weather bot
  - use `params` parameter in `requests.get()` method for passing query parameters

#### wikipedia bot
  - use `params` parameter in `requests.get()` method for passing query parameters

#### xkcd bot
  - not use `urllib.parse.quote` for encoding path parts (only IDs are used)

#### yoda bot
  - use `params` parameter in `requests.get()` method for passing query parameters

#### youtube bot
  - all query params were already passed as `params` parameter to `requests.get()` (nothing was changed)

### DIFF
--- a/zulip_bots/zulip_bots/bots/idonethis/fixtures/api_noop.json
+++ b/zulip_bots/zulip_bots/bots/idonethis/fixtures/api_noop.json
@@ -3,7 +3,8 @@
     "api_url": "https://beta.idonethis.com/api/v2/noop",
     "headers": {
         "Authorization": "Token 12345678"
-    }
+    },
+    "params": null
   },
   "response": {"doesnt": "matter"},
   "response-headers": {
--- a/zulip_bots/zulip_bots/bots/idonethis/fixtures/team_list.json
+++ b/zulip_bots/zulip_bots/bots/idonethis/fixtures/team_list.json
@@ -3,7 +3,8 @@
     "api_url": "https://beta.idonethis.com/api/v2/teams",
     "headers": {
         "Authorization": "Token 12345678"
-    }
+    },
+    "params": null
   },
   "response": [
     {
--- a/zulip_bots/zulip_bots/bots/idonethis/fixtures/test_401.json
+++ b/zulip_bots/zulip_bots/bots/idonethis/fixtures/test_401.json
@@ -3,7 +3,8 @@
     "api_url": "https://beta.idonethis.com/api/v2/teams",
     "headers": {
         "Authorization": "Token 87654321"
-    }
+    },
+    "params": null
   },
   "response": {
     "error": "Invalid API Authentication"
--- a/zulip_bots/zulip_bots/bots/idonethis/fixtures/test_create_entry.json
+++ b/zulip_bots/zulip_bots/bots/idonethis/fixtures/test_create_entry.json
@@ -8,7 +8,8 @@
     },
     "headers": {
         "Authorization": "Token 12345678"
-    }
+    },
+    "params": null
   },
   "response": {
     "body": "something and something else",
--- a/zulip_bots/zulip_bots/bots/idonethis/fixtures/test_create_entry_team_2.json
+++ b/zulip_bots/zulip_bots/bots/idonethis/fixtures/test_create_entry_team_2.json
@@ -8,7 +8,8 @@
     },
     "headers": {
         "Authorization": "Token 12345678"
-    }
+    },
+    "params": null
   },
   "response": {
     "body": "something and something else",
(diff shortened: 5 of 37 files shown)


timabbott:
So, the main concern I have about this is that generally for HTTP GET requests, one is expected to encode parameters in the URL, not as post-style parameters; I'm not confident that we should expect the APIs for every third-party service to parse `params` (as opposed to paremeters encoded for the URL).  We could test, but it might be better to just do the normal thing for what we pass into `requests.get`?

My other thought is that it's better to just defensively do the encoding step whether or not there are any special characters used; that way, if the third-party service changes the rules about their usernames (or whatever), the bot code is still correct.

novokrest:
@timabbott For the third-party service it is transparent how we call its API. For example, invoking `requests.get('http://server.com?query=value')` and `requests.get('http://server.com', params=dict(query=value))` produce the same outgoing HTTP-request to URL `http://server.com?query=value`. The main difference in these method calls is that the last call (with `params` parameter) automatically correctly encodes (if necessary) the query parameters according to the HTTP-standard.

timabbott:
OK, merged, thanks @novokrest!  I think for things like front and github detail, I would consider just using the `params` encoding model just for consistency (so that when people crib off an existing bot, they don't do it wrong)

PR #387, opened 2018-05-14
api: Implement /users/me/<stream_id>/topics.

YagoGG:
**Testing plan:**

With the Zulip development server running in the local machine:

```bash
$ curl -X POST -d "username=iago@zulip.com" http://localhost:9991/api/v1/dev_fetch_api_key
{"result":"success","msg":"","email":"iago@zulip.com","api_key":"5z1AlKmwAXPHCn3lsr4bo8IdxidUDtot"}
$ python zulip/zulip/examples/get-stream-topics --stream-id=15 --user=iago@zulip.com --api-key=5z1AlKmwAXPHCn3lsr4bo8IdxidUDtot --site=localhost:9991
{'result': 'success', 'msg': '', 'topics': [{'name': 'Denmark1', 'max_id': 128}, {'name': 'Denmark3', 'max_id': 124}, {'name': 'Denmark2', 'max_id': 117}]}
```

Plus all lint checks passed.


zulip/zulip/examples/get-stream-topics
```diff
@@ -0,0 +1,19 @@
...
+
+usage = """get-stream-topics --stream-id=<stream-id> [options]
+
+Get all the topics for a specific stream.
+"""
+
+import zulip
+
+parser = zulip.add_default_arguments(argparse.ArgumentParser(usage=usage))
+parser.add_argument('--stream-id')
```
timabbott: This should have been tagged as `required=True`.  

I also suspect that from a user perspective, we should probably have this example do two calls: First one to get the stream from name->ID, then another to do the actual thing (since on the command line, you probably have a name).

timabbott:
I fixed the `required=True` thing and merged; we should do an audit for similar issues where `required` is missing in the examples...

timabbott:
@YagoGG I think it's worth creating an issue (can you deal?  Sprints are hectic), but I also don't think there's a reason not to do it right now; it's I think pretty obvious in each example script which arguments are required or not, and I'm a fan of just cleaning out problems the moment one notices them where possible if it's easy :).

YagoGG:
Added the required fields for the rest of the examples in #388.

PR #388, opened 2018-05-14
api: Set required fields in the examples.

YagoGG:
Now all the API examples have their corresponding arguments set as required when they have to be present for the example to run.

**Testing plan:** Lint checks passed.


timabbott:
Nice, merged, thanks @YagoGG!  

I think for the testing plan, you should have instead said that you ran these manually (since argument parsing is basically always tested by hand).

PR #390, opened 2018-05-15
[WIP] New dependency-alerter bot & support for async bot-handlers

neiljp:
Motivation:
* Bot: It's challenging to keep up with latest updates of eg. mypy, so a bot seemed a reasonable idea :)
* Async: Grabbing the information via lots of http-requests takes time, as have other bots on occasion, so allowing for an async handler would enable the bots to be more responsive, to one user or between users.

Dependency-alerter bot:
* Looks at a requested URL for a `requirements.txt`-formatted file, then uses PyPI to check whether those dependencies are up to date.
* A dict of sample URLs are listed (zulip server repo!); open to including others as default :) 
* Marked [WIP], since tests and docs are relatively limited so far.

Async support in lib.py:
* Appears to work well :)
* If the bot has an async handler, then it will try and run that instead of the regular one
* The lib runs the async event-loop in a new thread, and runs the bot async message-handler in that loop when a new message is received
* It shuts down the loop and thread upon SIGINT cleanly
* This will require python 3.5+, which I'm not sure if we specify explicitly (3.4 is maybe implied?)
* We could work with pre-3.5, but the syntax is somewhat less clean
* Python 3.6 would make things more concise again, but that may be going too far for now, at least for the lib
* [WIP] since there may be lib tests needing to be added?

Async support in dependency-alert bot:
* Initially retained use of sync `requests`, but with `async def` and `await` as appropriate
* Expansion of dict comprehension to for loop is required currently, assuming we stick with only python 3.5
* Last commit migrates to the async `aiohttp` from `requests`; this is faster due to the async nature

All commits should pass `lint` and `run-mypy`.

This is also WIP since I've not gone through the idea of async support on czo, but hopefully it seems reasonable, at least as a first step :)

Potential future bot developments:
* cache the requirements / pypi downloads on a 1h-1d basis
* allow the repo dict to be amended at run-time (or use storage)
* improve output of the current/available; maybe a table (though this can look a little ugly?)
* if a download isn't cached, post the message as eg. "processing..." and then `update` it later, given it can take a while (then the user gets feedback, and it places it after the user message, in time)

@showell - since async is bot library related
@timabbott - discussed the bot with you on czo

### DIFF
--- a/tools/run-mypy
+++ b/tools/run-mypy
@@ -95,6 +95,7 @@ force_include = [
     "zulip_bots/zulip_bots/bots/susi/test_susi.py",
     "zulip_bots/zulip_bots/bots/front/front.py",
     "zulip_bots/zulip_bots/bots/front/test_front.py",
+    "zulip_bots/zulip_bots/bots/dependency_alerter/dependency_alerter.py",
     "tools/deploy"
 ]
 
--- a/zulip_bots/zulip_bots/bots/dependency_alerter/dependency_alerter.py
+++ b/zulip_bots/zulip_bots/bots/dependency_alerter/dependency_alerter.py
@@ -20,7 +20,7 @@ class DependencyAlerterHandler:
     def usage(self) -> str:
         return ''' '''
 
-    def handle_message(self, message: Dict[str, str], bot_handler: Any) -> None:
+    async def handle_message_async(self, message: Dict[str, str], bot_handler: Any) -> None:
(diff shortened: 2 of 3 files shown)


neiljp:
Perhaps unsurprisingly, this doesn't run pass the tests on python 3.4, though this seems as much about the presence of `Awaitable` in the annotations as the use of `async`.

PR #394, opened 2018-05-19
zulip-terminal: Add bot's directory to sys.path.

SivagiriVisakan:
This change was added to run.py in the commit 5992050d8276b523e099b76e55266912020342aa.
The same change is done to terminal.py so that the modules in the bot's directory can be imported when testing using `zulip-terminal` command.


showell:
Merged, thanks @SivagiriVisakan!

PR #397, opened 2018-05-22
zulip_botserver: Handle all requests from the root / endpoint.

roberthoenig:
Previously, the Botserver determined which bot to run by
dispatching on a different URL endpoint /bots/<botname> for each bot.
Now, instead, the Botserver determines which bot to run by the section
header of the bot in the flaskbotrc.
This implementation simplifies the setup of the Botserver. However,
without the dispatch on the URL endpoint, the Botserver cannot tell
anymore which bot to run for an incoming message. Therefore, it now
inspects each message for the bot name that is @mentioned in it. It
then looks for a section in the flaskbotrc that has this name as a
property and runs the associated bot. The name property is newly
introduced with this commit.

This depends on https://github.com/zulip/zulip/pull/9493.

### DIFF
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -2,6 +2,7 @@ import configparser
 import json
 import logging
 import os
+import re
 import signal
 import sys
 import time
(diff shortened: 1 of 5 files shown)


roberthoenig:
Hmm, this implementation will not work with private messages that do not @-mention the bot. So I guess we should first always try to find a bot email in the recipients, and only try to match the name if there is no email match.

roberthoenig:
Superseded by #398.

PR #398, opened 2018-05-22
zulip_botserver: Handle all requests from the root / endpoint.

roberthoenig:
Previously, the Botserver determined which bot to run by
dispatching on a different URL endpoint /bots/<botname> for each bot.
Now, instead, the Botserver determines which bot to run by the section
header of the bot in the flaskbotrc.
This implementation simplifies the setup of the Botserver. However,
without the dispatch on the URL endpoint, the Botserver cannot tell
anymore which bot to run for an incoming message. Therefore, it now
inspects each message for the bot name that is @mentioned in it. It
then looks for a section in the flaskbotrc that has this name as a
property and runs the associated bot. The name property is newly
introduced with this commit.

This implementation will not work with private messages that do not @-mention the bot. So I guess we should first always try to find a bot email in the recipients, and only try to match the name if there is no email match?


timabbott:
I think we should provide an explicit field in the outgoing webhook request as to what bot user is involved (E.g. the email for the bot user).  It'd be easy to do, and then we wouldn't need to do any parsing.

Since the botserver has basically never been widely used in production, we can just make it depend on a new feature we're adding for outgoing webhooks.

roberthoenig:
Yeah, makes sense. We can include the email along with the message flags, which are also needed by the botserver.

Do I assume correctly that when two outgoign webhook bots have the same name and are @mentioned, two messages will be dispatched?

timabbott:
Yeah, that's correct behavior -- those bots could be on totally different servers and do totally different things.

roberthoenig:
Updated.

roberthoenig:
@timabbott bump on this.

roberthoenig:
I added a series of commits that do cleanups and fix the fact that the botserver didn't strip messages with @mentions. For this, I also expanded the test system to accept real message tests. It's pretty basic and limited, but it serves the purpose for the botserver.

roberthoenig:
Updated to only remove the second check. Also, added a new commit where I really just wanted to use a Python structure I recently learned about :)

timabbott:
Merged, thanks @roberthoenig!  

Do you want to take the little `token` project as well while you have this in your head?

timabbott:
I was thinking of the value that gets stored in `Service.token`, but I suspect they're the same.

PR #401, opened 2018-05-25
Create finder.py

SivagiriVisakan:
This PR splits the path resolution logic and import_module_from_source into  a seperate file.

### DIFF
--- a/zulip_bots/zulip_bots/finder.py
+++ b/zulip_bots/zulip_bots/finder.py
@@ -1,0 +1,23 @@
+import sys
+from os.path import basename, splitext
+from typing import Any, Optional, Text
+
+def import_module_from_source(path: Text, name: Optional[Text]=None) -> Any:
+    if not name:
+        name = splitext(basename(path))[0]
+
+    # importlib.util.module_from_spec is supported from Python3.5
+    py_version = sys.version_info
+    if py_version.major < 3 or (py_version.major == 3 and py_version.minor < 5):
+        import imp
+        module = imp.load_source(name, path)
+    else:
+        import importlib.util
+        spec = importlib.util.spec_from_file_location(name, path)
+        module = importlib.util.module_from_spec(spec)
+        loader = spec.loader
+        if loader is None:
+            return None
+        loader.exec_module(module)
+
+    return module
--- a/zulip_bots/zulip_bots/finder.py
+++ b/zulip_bots/zulip_bots/finder.py
@@ -5,9 +5,7 @@ from typing import Any, Optional, Text, Tuple
 
 current_dir = os.path.dirname(os.path.abspath(__file__))
 
-def import_module_from_source(path: Text, name: Optional[Text]=None) -> Any:
-    if not name:
-        name = splitext(basename(path))[0]
+def import_module_from_source(path: Text, name: Text) -> Any:
 
     # importlib.util.module_from_spec is supported from Python3.5
     py_version = sys.version_info
--- a/zulip_bots/zulip_bots/finder.py
+++ b/zulip_bots/zulip_bots/finder.py
@@ -1,6 +1,9 @@
 import sys
+import os
 from os.path import basename, splitext
-from typing import Any, Optional, Text
+from typing import Any, Optional, Text, Tuple
+
+current_dir = os.path.dirname(os.path.abspath(__file__))
 
 def import_module_from_source(path: Text, name: Optional[Text]=None) -> Any:
     if not name:
@@ -21,3 +24,13 @@ def import_module_from_source(path: Text, name: Optional[Text]=None) -> Any:
         loader.exec_module(module)
 
     return module
+
+def resolve_bot_path(name: Text) -> Tuple[Text, Text]:
+    if os.path.isfile(name):
+        bot_path = os.path.abspath(name)
+        bot_name = splitext(basename(bot_path))[0]
+    else:
+        bot_path = os.path.abspath(os.path.join(current_dir, 'bots', name, name + '.py'))
+        bot_name = name
+
+    return (bot_path, bot_name)
--- a/zulip_bots/zulip_bots/run.py
+++ b/zulip_bots/zulip_bots/run.py
@@ -6,38 +6,14 @@ import sys
 import os
 from os.path import basename, splitext
 
-from typing import Any, Optional, Text
-
 from zulip_bots.lib import (
     run_message_handler_for_bot,
     NoBotConfigException,
 )
-
+from zulip_bots import finder
 from zulip_bots.provision import provision_bot
 
 current_dir = os.path.dirname(os.path.abspath(__file__))
-
-
-def import_module_from_source(path: Text, name: Optional[Text]=None) -> Any:
-    if not name:
-        name = splitext(basename(path))[0]
-
-    # importlib.util.module_from_spec is supported from Python3.5
-    py_version = sys.version_info
-    if py_version.major < 3 or (py_version.major == 3 and py_version.minor < 5):
-        import imp
-        module = imp.load_source(name, path)
-    else:
-        import importlib.util
-        spec = importlib.util.spec_from_file_location(name, path)
-        module = importlib.util.module_from_spec(spec)
-        loader = spec.loader
-        if loader is None:
-            return None
-        loader.exec_module(module)
-
-    return module
-
 
 def parse_args() -> argparse.Namespace:
     usage = '''
@@ -115,7 +91,7 @@ def main() -> None:
         provision_bot(os.path.dirname(bot_path), args.force)
 
     try:
-        lib_module = import_module_from_source(bot_path, bot_name)
+        lib_module = finder.import_module_from_source(bot_path, bot_name)
     except ImportError as e:
         req_path = os.path.join(os.path.dirname(bot_path), "requirements.txt")
         with open(req_path) as fp:
--- a/zulip_bots/zulip_bots/run.py
+++ b/zulip_bots/zulip_bots/run.py
@@ -80,13 +80,10 @@ def exit_gracefully_if_bot_config_file_does_not_exist(bot_config_file: str) -> N
 
 def main() -> None:
     args = parse_args()
-    if os.path.isfile(args.bot):
-        bot_path = os.path.abspath(args.bot)
-        bot_name = os.path.splitext(basename(bot_path))[0]
-    else:
-        bot_path = os.path.abspath(os.path.join(current_dir, 'bots', args.bot, args.bot+'.py'))
-        bot_name = args.bot
+
+    bot_path, bot_name = finder.resolve_bot_path(args.bot)
     sys.path.insert(0, os.path.dirname(bot_path))
+
     if args.provision:
         provision_bot(os.path.dirname(bot_path), args.force)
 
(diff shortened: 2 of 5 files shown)


showell:
This is awesome!  The code seems way more organized this way.

zulip_bots/zulip_bots/run.py
```diff
@@ -80,13 +80,9 @@ def exit_gracefully_if_bot_config_file_does_not_exist(bot_config_file: str) -> N
 
 def main() -> None:
     args = parse_args()
-    if os.path.isfile(args.bot):
-        bot_path = os.path.abspath(args.bot)
-        bot_name = os.path.splitext(basename(bot_path))[0]
-    else:
-        bot_path = os.path.abspath(os.path.join(current_dir, 'bots', args.bot, args.bot+'.py'))
-        bot_name = args.bot
-    sys.path.insert(0, os.path.dirname(bot_path))
```
roberthoenig: The code removed here seems not semantically identical to the new code in `resolve_bot_path`. Can you split this commit into two commits: One where you just move the code as it is, and another where you make the logical changes?

zulip_bots/zulip_bots/finder.py
```diff
@@ -21,3 +24,14 @@ def import_module_from_source(path: Text, name: Optional[Text]=None) -> Any:
...
     return module
+
+def resolve_bot_path(name: Text) -> Tuple[Text, Text]:
+    if os.path.isfile(name) or os.path.isfile(name + '.py'):
+        bot_path = os.path.abspath(name) if os.path.isfile(name) else os.path.abspath(name + '.py')
+        bot_name = splitext(basename(bot_path))[0]
+    else:
+        bot_path = os.path.abspath(os.path.join(current_dir, 'bots', name, name + '.py'))
+        bot_name = name
+    sys.path.insert(0, os.path.dirname(bot_path))
```
roberthoenig: I'm not sure if we want this function to also insert the bot path into the environment PATH, it's not something I'd expect a function named `resolve_bot_path` to do. Maybe we can find a more suitable name for it. @showell do you have an opinion on this?

zulip_bots/zulip_bots/terminal.py
```diff
@@ -3,7 +3,7 @@
 import sys
 import argparse
 
-from zulip_bots.run import import_module_from_source
+from zulip_bots.finder import import_module_from_source
```
roberthoenig: This should happend in the first commit where you move the function to `import_module_from_source`. The idea is to keep commits as atomical elements that work on their own. If, for example, we would now commit the first commit only, the tests would fail for zulip-terminal.

zulip_bots/zulip_bots/test_finder.py
```diff
@@ -0,0 +1,20 @@
...
+    def test_resolve_bot_path_for_custom_bot(self):
+
+        with patch('os.path.isfile', return_value=True), \
+                patch('os.path.abspath', return_value='path/to/custom_handler.py'):
+
+            custom_bot_handlers = ['custom_handler', 'custom_handler.py']
+            for custom_bot_handler in custom_bot_handlers:
+                expected_bot_path_and_name = ('path/to/custom_handler.py', 'custom_handler')
+                bot_path_and_name = finder.resolve_bot_path(custom_bot_handler)
+                self.assertEqual(bot_path_and_name, expected_bot_path_and_name)
```
roberthoenig: These tests should be commited in the commit that creates `resolve_bot_path`.

zulip_bots/zulip_bots/test_finder.py
```diff
@@ -0,0 +1,20 @@
+from zulip_bots import finder
+import unittest
+from unittest import TestCase
+
```
roberthoenig: Redundant newline.

zulip_bots/zulip_bots/test_finder.py
```diff
@@ -0,0 +1,20 @@
...
+from unittest import mock
+from unittest.mock import patch
+
+
```
roberthoenig: Redundant newline.

zulip_bots/zulip_bots/test_finder.py
roberthoenig: This test seems a bit too dependent on the inner workings of `resolve_bot_path`. Instead of using a fake path and mocking the `os.path` calls, it's probably more robust and closer to real use cases if this tests would try to resolve the paths of actual existing bots, like `helloworld`.

zulip_bots/zulip_bots/test_finder.py
```diff
@@ -1,14 +1,17 @@
...
-
+from typing import Text
 from unittest import mock
 from unittest.mock import patch
 
 
 class FinderTestCase(TestCase):
+    def setUp(self) -> None:
+        self.original_current_dir = finder.current_dir
+        finder.current_dir = 'api-repo/zulip_bots/zulip_bots'
```
roberthoenig: Why are we changing `finder.current_dir` here?

zulip_bots/zulip_bots/test_finder.py
```diff
@@ -18,3 +21,20 @@ def test_resolve_bot_path_for_custom_bot(self):
...
+            return 'absolute/path/to/'+path
+
+        with patch('os.path.isfile', return_value=False), \
+                patch('os.path.dirname'), \
+                patch('os.path.abspath', side_effect=abspath_side_effect):
+
+            expected_existing_bot_path_and_name = ('absolute/path/to/api-repo/zulip_bots/zulip_bots/bots/helloworld/helloworld.py',
+                                                   'helloworld')
+            existing_bot_path_and_name = finder.resolve_bot_path('helloworld')
+            self.assertEqual(existing_bot_path_and_name, expected_existing_bot_path_and_name)
```
roberthoenig: From how I see it, this function doesn't really test resolve_bot_path for an existing bot, since the paths used are still nonexistent. I'd just make a single function `test_resolve_bot_path` that tests to resolve the actual path for an actual bot.

SivagiriVisakan:
@timabbott `zulip_botserver` is being distributed as a seperate pip package . I think it would be better if it didn't have to depend on the code from `zulip_bots`, no ?

timabbott:
@SivagiriVisakan no, it's a server for running things inside `zulip_bots`, so that doesn't make any sense.

zulip_bots/zulip_bots/finder.py
```diff
@@ -0,0 +1,36 @@
+import sys
+import os
+from os.path import basename, splitext
+from typing import Any, Optional, Text, Tuple
+
+current_dir = os.path.dirname(os.path.abspath(__file__))
+
+def import_module_from_source(path: Text, name: Optional[Text]=None) -> Any:
```
novokrest: I think the second argument `name` should be of type `Text` and must be required. Now this method is always used with both passed arguments.

zulip_bots/zulip_bots/run.py
```diff
@@ -104,17 +80,14 @@ def exit_gracefully_if_bot_config_file_does_not_exist(bot_config_file: str) -> N
...
     args = parse_args()
-    if os.path.isfile(args.bot):
-        bot_path = os.path.abspath(args.bot)
-        bot_name = os.path.splitext(basename(bot_path))[0]
-    else:
-        bot_path = os.path.abspath(os.path.join(current_dir, 'bots', args.bot, args.bot+'.py'))
-        bot_name = args.bot
+
+    bot_path, bot_name = finder.resolve_bot_path(args.bot)
     sys.path.insert(0, os.path.dirname(bot_path))
```
novokrest: @SivagiriVisakan maybe it is worth to move this logic to `import_module_from_source` method ? When we will implement importing of bot's module by module name (such as `some.bot.module.name`) this line will be used only for case when bot's module are imported by module's file path

zulip_bots/zulip_bots/finder.py
SivagiriVisakan: Is there any reason that it must always be called with `name` argument ? 
If the `name`  is not specified, it is assigned [here](https://github.com/zulip/python-zulip-api/pull/401/files/7b7e10afdcc6b2822bb175e492ab88a99036625f#diff-7d6ba3f111cf7a3c9528de51bd6a1e1eR9).

zulip_bots/zulip_bots/run.py
SivagiriVisakan: I am not sure if we want to move this logic under `import_module_from_source`. I think it is more organized this way( for which this PR was originally opened, to better organize code).
I think the name `resolve_bot_path` is clear enough to denote that we are importing the through file path.
@showell @roberthoenig Should this be moved as per @novokrest is suggesting ?

roberthoenig: I don't have strong opinions on that, I'm fine with either option. For now, since this
> When we will implement importing of bot's module by module name (such as some.bot.module.name)
is done yet (e.g. `provision_bot` still required the bot_path as an argument), I think we should keep this as it is right now. Updating it later is easy.

zulip_bots/zulip_bots/finder.py
novokrest: @SivagiriVisakan I meant that the current method's calls don't use the ability to pass only one argument. This method is always called with both arguments in production code. That's why I proposed to change type `name` to just `Text` because `if not name` is always `False` in production code. I agree that your current implementation of method is more flexible, but, unfortunately, we don't need this flexibility now.

zulip_bots/zulip_bots/run.py
novokrest: I will not argue because I agree completely that we can update it easy if we will need it. Current implementation is fine for me too.

showell:
@novokrest @roberthoenig Do either of you have time to take this to the finish line?  I think if one of you can give the final LGTM, we'll merge it, otherwise it might make sense for one of you to just fix any remaining nits.

zulip_bots/zulip_bots/tests/test_finder.py
```diff
@@ -0,0 +1,19 @@
...
+from zulip_bots import finder
+
+from unittest import TestCase
+from unittest.mock import patch
+
+
+class FinderTestCase(TestCase):
+
+    def test_resolve_bot_path(self) -> None:
+        with patch('os.path.isfile', return_value=False):
```
roberthoenig: Why do we need to patch os.path.isfile?

SivagiriVisakan: Because if a file exists at the given `name`, then the absolute path to it is returned by [`resolve_bot_path`](https://github.com/zulip/python-zulip-api/pull/401/commits/e41af88c30cb6db9acdfac5200558761c5e632e7#diff-7d6ba3f111cf7a3c9528de51bd6a1e1eR28). (We have it like this to support running custom bots i.e bots which are not included in python-zulip-api.)

So I thought we would be explicit here and patch isfile to be false to run the test for the path of an existing bot.

roberthoenig: In that case I think we should just remove this line. We expect `resolve_bot_path` to do the right thing on its own, and it should just see that there is no `helloworld` file in the `zulip_bots/zulip_bots` directory.

SivagiriVisakan: @roberthoenig I have removed the patch for os.path.isfile .

roberthoenig:
@showell this is ready to merge from my side.

showell:
merged, thanks @SivagiriVisakan!  thanks @roberthoenig @novokrest for help here too

Issue #402, opened 2018-05-26
Bot tests are not being fully executed


neiljp:
Noticed first when exploring the use of `pytest`, which should work with `unittest.TestCase` - but it couldn't find a bot dependency (`tweepy` for twitpost). Then I noticed that this bot was not even being tested!

Cases so far:
* tools/test-bots: **passes OK** with **270 tests** <-- what is seen in CI, presumably?
* tools/test-bots * each bot (scripted): **fails 5 times** with **283 tests**
* Some bots are not being tested at all, and no error/failure is indicated.

neiljp:
@zulipbot claim

neiljp:
With some simplification and common methods for test discovery in `test-bots`, both `test-bots [bot-names-in-bots-dir]` and `test-bots` currently find **295** tests, with **11** failures.

This is after addition of many `__init__.py` files to bot directories, which the current discovery misses.

WIP PR incoming for this, likely tomorrow, though more work will be needed to fix the tests.

timabbott:
Awesome, thanks for discovering this!  I wonder what the right tooling method to prevent folks from adding bots without `__init__.py` files is...

neiljp:
I'm currently working on fixing the bot tests.

I have an alternative solution for the test discovery which seems simpler in terms of identifying tests, though with the other issues found I will explore whether it is necessary.

Agreed that including detection of cases with no `__init__.py` files is definitely important to remedy!

neiljp:
So master has the following, with different test discovery for each:
* `test-bots` -> 270 tests pass
* `test-bots ``ls zulip_bots/zulip_bots/bots|grep -v __|grep -v bridge|grep -v fifte`-> 287 run, 5x E

So far my branch:
* fixes fifteen naming (for consistency, in case we want names to match, and to add `_`s)
* adds 3 missing `__init__.py` files for bots with good tests -> 288 tests from `test-bots`
* adds `__init__.py` for twitpost & adds the dependency to requirements.txt -> 292 tests from `test-bots`
* adds `__init__.py` for witai and fixes test issues -> 295 tests from `test-bots`
* amends test discovery -> only finds 293 tests (3 failing chess ones:)
* renames `chess` to `chessbot` (provisional name), to avoid it conflicting with the pypi module name -> 293 passing tests from `test-bots`

There is now "just":
* the issue of which 2 tests *aren't* being run in my update `test-bots`
* ensuring bots have a `__init__.py` file during tests
[...]

PR #403, opened 2018-05-27
Ensure additional bot testing & general test-bots improvements

neiljp:
See #402 for reference, but:
* Consistently name gameoffifteen as game_of_fifteen (based on existing bot testing requirements)
* Enable testing of 3 bots by simply adding `__init__.py` files
* Slightly modify 3 bots (twitpost, witai, chess) and add `__init__.py` to get them testing
* Incrementally improve test discovery and style in `test-bots`

This should be easy to review, as most commits are small and many are independent.

The penultimate commit enables the updated `test-bots` to run cleanly for bots with no `__init__.py`, as currently (mistakenly!). This behavior can be modified later (and controlled at run-time with `--error-on-no-init`), but maintains current status until the last bot with tests but no `__init__.py` has their tests fully fixed - which the last commit starts to do.

NOTE: From "Use unused available_bots to discover tests" to "Detect absent `__init__.py` & optionally exit", merels will currently fail; this can easily be minimised with a rebase, but to avoid it entirely would require a little more work.

### DIFF
--- a/zulip_bots/zulip_bots/bots/merels/test/test_constants.py
+++ b/zulip_bots/zulip_bots/bots/merels/test/test_constants.py
@@ -1,6 +1,6 @@
 import unittest
 
-from libraries import constants
+from ..libraries import constants
 
 
 class CheckIntegrity(unittest.TestCase):
--- a/zulip_bots/zulip_bots/bots/merels/test/test_database.py
+++ b/zulip_bots/zulip_bots/bots/merels/test/test_database.py
@@ -1,7 +1,7 @@
 import unittest
 
-from libraries import database
-from libraries import game_data
+from ..libraries import database
+from ..libraries import game_data
 from zulip_bots.simple_lib import SimpleStorage
 
 
--- a/zulip_bots/zulip_bots/bots/merels/test/test_game.py
+++ b/zulip_bots/zulip_bots/bots/merels/test/test_game.py
@@ -1,7 +1,7 @@
 import unittest
 
-from libraries import game
-from libraries import database
+from ..libraries import game
+from ..libraries import database
 
 from zulip_bots.simple_lib import SimpleStorage
 
--- a/zulip_bots/zulip_bots/bots/merels/test/test_interface.py
+++ b/zulip_bots/zulip_bots/bots/merels/test/test_interface.py
@@ -1,6 +1,6 @@
 import unittest
 
-from libraries import interface
+from ..libraries import interface
 
 
 class BoardLayoutTest(unittest.TestCase):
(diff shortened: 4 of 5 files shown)


timabbott:
Merged all but the last commit.

roberthoenig:
@neiljp in some commits paths hardcoded paths like
```
top_level = "zulip_bots/zulip_bots/bots/"
```
were used. I'm fairly confident they'll lead to tests breaking on Windows. While you're at it, can you normalize the paths. You could e.g. use
```
top_level = os.path.normpath("zulip_bots/zulip_bots/bots/")
```

neiljp:
@roberthoenig We use `normpath` in one place in the api repo, and run no automated tests on windows. Other than bug reports, how do we know what works? Is there some guide we can follow for this, that you can recommend?

Maybe we need automated testing on Windows, if we intend to support that?

roberthoenig:
@neiljp Where we don't use normpath we probably use `os.sep`. Afaik, Travis doesn't support Windows builds. One CI service that does support Windows is AppVeyor. I don't know if it's worth setting it up, but opening an issue can't hurt.

I've been doing a lot of work on this repo from a Windows machine, so fortunately there shouldn't be many issues. But other than that, yes, only reports and careful staring can help, and linter rules.

showell:
@neiljp This LGTM but we seem to have a build failure.  Can you investigate that and rebase as needed?

neiljp:
@showell The build failure is related to the merels bot tests actually running with that commit, which they may have been at some point (somehow), but are some ways from doing right now from what I can tell.

I can certainly investigate further but was unsure whether to delegate to the bot or tests author (@amanagr ? @Privisus ?).

amanagr:
An`__init__` file was missing the merels bot directory and hence the tests were not running until now. I send a PR to fix it and then @neiljp can rebase.

amanagr:
The current test for merels is according to the previous implementation. The new implementation with the game handler will require a lot of changes in the tests. I will just make merels pass tests by removing the cases which don't work and require a complete workaround.
PS: I feel sorry that I overlooked this earlier.

neiljp:
@amanagr Could you add an issue such as "Rework and enable Merels tests" with your understand of the situation? `test-bots` as it stands will just warn about a missing `__init__.py` file, or if run with `-p` (just merged) then excludes it explicitly in the script as it stands.

My feeling would be that if we have a separate issue to track this, we should be able to close this PR.

amanagr:
@neiljp created #433 for it.

neiljp:
Closed as per above, ie. pending #433.

PR #406, opened 2018-05-28
Bots: Improve typing & code in various bots

neiljp:
This PR started off intending to clarify some typing, and ended up in some tidying of the code and tests as I got carried away; these primarily increase legibility and removal of repeated code.

I also found a suspected bug in the baremetrics output, as per the added comment line, namely a switch of name and display_name, which was revealed by use of the `dict` style formatting.


neiljp:
Just pushed again after missing an adjustment to allow acceptance in python3.4 which should have already been included in the branch.

showell:
Merged, thanks @neiljp!

PR #410, opened 2018-05-30
bot lib tests: Move tests of bot library to tests/.

neiljp:
I found it unclear which files were for bots to use in writing tests, and which were tests of the bot infrastructure; this is intended to make it clearer by having a separate directory for the latter.

The `lib_tests.py` has been re-named to `tests/test_lib.py`; note that this is the same filename as `test_lib.py`, which contains the lib code for testing bots, but is in the parent directory. 

I also found that the tests in `test_run.py` were not being run, so this now re-enables those extra 5 tests.


timabbott:
Merged, thanks for cleaning this up @neiljp!

PR #411, opened 2018-06-01
Make the Botserver CLI easier to use.

roberthoenig:


timabbott:
@roberthoenig let me know when you have the `zulip-botserver` rename done :)

roberthoenig:
@timabbott updated.

timabbott:
@roberthoenig can you investigate the CI failure?

zulip_botserver/tests/test_server.py
```diff
@@ -108,19 +107,19 @@ def test_wrong_bot_credentials(self, mock_StateHandler: mock.Mock, mock_LoggingE
...
-                                  available_bots=available_bots,
-                                  event=dict(message={'content': "@**test** test message"},
-                                             bot_email='helloworld-bot@zulip.com',
-                                             trigger='mention',
-                                             token='abcd1234'),
-                                  bots_config=bots_config))
+        # This works, but mypy still complains:
+        # error: No overload variant of "assertRaisesRegexp" of "TestCase" matches argument types
+        # [def (*args: builtins.object, **kwargs: builtins.object) -> builtins.SystemExit, builtins.str]
+        with self.assertRaisesRegexp(SystemExit,  # type: ignore
```
timabbott: I think you want assertRaisesRegex?

timabbott: I guess that isn't the problem, though; the problem is probably that `SystemExit` isn't an `Exception`.

timabbott:
Merged, thanks @roberthoenig!  

I guess we should finish converting the way config is provided, and then do a release, soon...

roberthoenig:
:/ okay, although I'm afraid of that task...

timabbott:
Given that, maybe it's reasonable to hold that for the next release; I think it's a less disruptive change than renaming the tool :).

@eeshangarg can you prepare a release?  I think we might want to do 0.5.0 as the version.  The key changelog item here is that we've more or less rewritten the botserver to be easier to use (and renamed its command).

PR #412, opened 2018-06-01
#357 Add a better way to run standalone bots 

### LINKED ISSUE #357: Add a better way to run standalone bots
```zulip-run-bot``` currently supports running a bot in two ways:

1. By name, assuming it exists in the ```zulip_bots``` directory in a local clone of ```python-zulip-api```
2. By an explicit path to the bot's handler .py file, regardless of where it lives

Neither of those two ways feels 100% well suited for running a standalone bot (a bot not meant to be contributed back or deployed as part of ```python-zulip-api```) whose code and handler class is already deployed in the current python environment, for example as a standalone pip-installed package.

Can we add a better way to run this kind of bot? Perhaps by passing to ```zulip-run-bot``` an explicit module path to the python module that holds the bot's handler class, or something similar. Or perhaps by providing and advertising an easy to customize bit of code similar to what powers ```zulip-run-bot``` that can be pointed to the bot handler class, which I hear [might already be in the works](https://chat.zulip.org/#narrow/stream/127-integrations/subject/standalone.20bot.20code/near/547843) for docker-deployed bots, etc.

See [this conversation](https://chat.zulip.org/#narrow/stream/127-integrations/subject/standalone.20bot.20code/near/547688) for more context on this issue.

### PR DESCRIPTION
**Pull request by novokrest**
Fixes #357.

Change logic to run bot by `zulip-run-bot` to the following:
- If `zulip-run-bot` was run with _name_ as path to bot's file, then import bot from file and add bot's parent dir to `sys.path`;
- If `zulip-run-bot` was run with some _name_ (not pointing to file) then:
  - try to import bot from file located at `zulip_bots/zulip_bots/bots/{name}/{name}.py`. If import was completed succesfully, then add add bot's parent dir (`zulip_bots/zulip_bots/bots/{name}`) to `sys.path`;
  - try to import bot by specified module _name_. If import was completed successfully, then also check that `args.provision` was *not* passed

### DIFF
--- a/zulip_bots/zulip_bots/run.py
+++ b/zulip_bots/zulip_bots/run.py
@@ -81,25 +81,34 @@ def exit_gracefully_if_bot_config_file_does_not_exist(bot_config_file: str) -> N
 def main() -> None:
     args = parse_args()
 
-    bot_path, bot_name = finder.resolve_bot_path(args.bot)
-    sys.path.insert(0, os.path.dirname(bot_path))
+    result = finder.resolve_bot_path(args.bot)
+    if result:
+        bot_path, bot_name = result
+        sys.path.insert(0, os.path.dirname(bot_path))
 
-    if args.provision:
-        provision_bot(os.path.dirname(bot_path), args.force)
+        if args.provision:
+            provision_bot(os.path.dirname(bot_path), args.force)
 
-    try:
-        lib_module = finder.import_module_from_source(bot_path, bot_name)
-    except ImportError as e:
-        req_path = os.path.join(os.path.dirname(bot_path), "requirements.txt")
-        with open(req_path) as fp:
-            deps_list = fp.read()
+        try:
+            lib_module = finder.import_module_from_source(bot_path, bot_name)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: zulip_bots/zulip_bots/run.py [hunk 1]
```diff
@@ -18,10 +18,7 @@
 current_dir = os.path.dirname(os.path.abspath(__file__))
 
 
-def import_module_from_source(path: Text, name: Optional[Text]=None) -> Any:
-    if not name:
-        name = splitext(basename(path))[0]
-
+def import_module_from_source(path: Text, name: Text) -> Any:
```
timabbott: I think the commit message for this commit has the name out of order :)

timabbott:
Can you compare this to e.g. https://github.com/zulip/python-zulip-api/pull/401 ?

novokrest:
@timabbott @aero31aero I updated PR. I think it is ready for review now

aero31aero:
There are some linting errors that need fixing.

showell:
Merged, thanks @novokrest!

PR #413, opened 2018-06-02
Bring bridges together in zulip & improve+publish matrix bridge

neiljp:
There are 3 main thrusts to this PR:
* Bring the 3 existing bridges together into `zulip/integrations` and rename them in a cohesive way (including moving the "interrealm bot" for zulip<->zulip from `zulip_bots`)
* Make the matrix bridge easier to use generally;
* Include (install) the matrix bridge by default in the `zulip` package.

Potential issues:
* the zulip-zulip bridge (`bridge_between_zulips`) uses python3 type annotations, whereas the rest of the API uses python2 code; this could be reverted, assuming the code is compatible, or we could add a run-time check for python3?
* the matrix bridge adds a dependency; this is just one, but should we require it just for the API? This suggests that maybe the integrations/bridges belong in yet another package?
* the error-handling commit modifies the caught exception in `matrix_to_zulip` to just be `Exception`, since I highly doubt that the Zulip API raises a `MatrixRequestError`; however, are the exceptions potentially thrown by the (python) Zulip API clearly defined?
* tests for the script would be useful; I've previously achieved this using `subprocess` (`Popen`) calls, but this can be slow to run many tests, so I've held off on writing tests for this at this point.

This does not include setting up the existing IRC or inter-realm bot at this stage, based on the recent work on this by @rheaparekh and the push to get it out and used/tested, though the latter seems like a good next step in this series.

Potential reviewers:
@rheaparekh - you've worked on this most-recently :)
@showell @timabbott - this includes the renames and script install we discussed in #integrations.


zulip/integrations/bridge_with_matrix/matrix_bridge.py
```diff
@@ -192,11 +198,49 @@ def read_configuration(config_file: str) -> Dict[str, Dict[str, str]]:
...
+
+    if options.sample_config:
+        try:
+            write_sample_config(options.sample_config)
+        except Bridge_Exception as exception:
+            sys.exit(exception)
+        print("Wrote sample configuration to '{}'".format(options.sample_config))
+        sys.exit(0)
+    elif not options.config:
+        print("Options required: -c or --config to run, OR --write-sample-config.")
```
rheaparekh: I think we should update the docs with this method. The README.md, [Matrix integration doc](https://github.com/zulip/zulip/blob/master/templates/zerver/integrations/matrix.md) and [IRC integration doc](https://github.com/zulip/zulip/blob/master/templates/zerver/integrations/irc.md)

timabbott:
I merged one small refactor commit when I had a quick look.  I would have much preferred doing the renames at the end of the branch (actually, this is a general rule, of putting potentially controversial things at the end), since I'm not sure I'm happy with the naming scheme and want to think about it more, whereas many of the other cleanups are good to merge immediately.  If it's not hard, it'd be great if you can reorder things.  A few notes on the code changes:

* I like the configparser migration (agree we need to update the docs when we make that change) and the shift to writing a sample config, though for that latter bit, I think it'd be a good idea to have it take a `zuliprc` file as an argument, so that we can provide good defaults for email/apikey/site.
* Adding `matrix_client` as a dependency seems not totally ideal, given how minimal we've been with dependencies for the core `zulip` package, but might be worth it.  I want to think about that.

zulip/integrations/bridge_with_matrix/matrix_bridge.py
```diff
@@ -212,7 +218,14 @@ def main() -> None:
...
             print("Starting message handler on Zulip client")
             zulip_client.call_on_each_message(zulip_to_matrix(zulip_config, room))
-        except Exception:
+
+        except Bridge_MatrixException as exception:
+            sys.exit("Matrix error: {}".format(exception))
+        except Bridge_ZulipException as exception:
+            sys.exit("Zulip error: {}".format(exception))
+        except zulip.ZulipError as exception:
+            sys.exit("Zulip error: {}".format(exception))
```
timabbott: I think we should have names like `Bridge_ZulipFatalException` if these are going to be fatal, and only use them for things that are structurally fatal, like invalid API key.  Otherwise, I worry this bot will kill itself by accident.

zulip/integrations/bridge_with_matrix/matrix_bridge.py
```diff
@@ -76,12 +82,12 @@ def _matrix_to_zulip(room: Any, event: Dict[str, Any]) -> None:
...
                 })
-            except MatrixRequestError as e:
+            except Exception as e:  # XXX This should be more specific
                 # Generally raised when user is forbidden
-                raise Exception(e)
+                raise Bridge_ZulipException(e)
             if result['result'] != 'success':
                 # Generally raised when API key is invalid
-                raise Exception(result['msg'])
+                raise Bridge_ZulipException(result['msg'])
```
timabbott: We should check the `result['code']` to see whether it's actually an invalid API key, here.  

(If it there is no `code`, I'll add one to the server).

zulip/integrations/bridge_with_matrix/matrix_bridge.py
```diff
@@ -26,27 +26,33 @@
...
     try:
         room = matrix_client.join_room(matrix_config["room_id"])
         return room
     except MatrixRequestError as e:
         if e.code == 403:
-            sys.exit("Room ID/Alias in the wrong format")
+            raise Bridge_MatrixException("Room ID/Alias in the wrong format")
         else:
-            sys.exit("Couldn't find room.")
+            raise Bridge_MatrixException("Couldn't find room.")
```
timabbott: This isn't new to your code, but I'm not a fan of these else clauses for non-403 errors assuming what the error is; can't we just pass through whatever the text of the `MatrixRequestError` was?

zulip/integrations/bridge_between_zulips/run-interrealm-bridge
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
neiljp: I've amended the commit message, but I refer to my initial comment re whether we should return type annotations to python2 style in this and the matrix bridge.

In the next set of commits, I've also included a commit which moves back to python2 type annotations for the matrix bridge and an extra dependency to ensure the script is compatible with python2.

zulip/integrations/bridge_with_matrix/matrix_bridge.py
neiljp: I have renamed the internal exceptions to include `Fatal`. Generally, however, I'm not sure which exceptions are *not* fatal, except maybe the very general `except` clause.

From what I can tell, `ZulipError` is not recoverable?

zulip/integrations/bridge_with_matrix/matrix_bridge.py
neiljp: I'm not sure where to look for this - @rheaparekh can you comment on this, since you added this exception section in `93627550`?

zulip/integrations/bridge_with_matrix/matrix_bridge.py
neiljp: This seems reasonable to some degree, though if we have a concrete idea that the error is something specific, I'm wondering if it makes sense to interpret it before wrapping it in the different `Exception`? Or maybe chaining with something like:
`Bridge_MatrixException("Couldn't find room. "+ str(exception))` ?

@rheaparekh You added this error handling; could you comment?

neiljp:
I've made some improvements, but will hold off pushing the rebased version until a little more feedback, to avoid losing the discussion points.

zulip/integrations/bridge_with_matrix/matrix_bridge.py
rheaparekh: There is no `code` here. We get the result as `{'msg': 'Invalid API key', 'result': 'error'}`. I just tested and confirmed this again.

zulip/integrations/bridge_with_matrix/matrix_bridge.py
rheaparekh: I had been following the sample codes for the official `matrix_client` for this, but the sample codes aren't consistent which each other as well. It will be better to pass just the text of the `Exception` without chaining it.

rheaparekh: Sample code: https://github.com/matrix-org/matrix-python-sdk/tree/master/samples

zulip/integrations/bridge_with_matrix/matrix_bridge.py
timabbott: OK.  I'll integrate adding one; basically, other errors that might occur here feel more transient.

zulip/integrations/bridge_with_matrix/matrix_bridge.py
timabbott: I think some transient network errors might come back as `ZulipError`?

zulip/integrations/bridge_with_matrix/matrix_bridge.py
timabbott: OK, sounds like the library is a bit sloppy.  Let's hold off on trying to clean this up until we've done the rest.

zulip/integrations/bridge_with_matrix/matrix_bridge.py
neiljp: @timabbott If you mean from the Zulip python API (ie. `__init__.py`), there's:
- `Client.__init__` -> `ZulipError`, `ConfigNotFoundError`, `MissingURLError`
- `Client.do_api_query` -> `ZulipError`
Of those, the latter appears to occur only from a `requests` `ConnectionError` if there has been no server connection (~line 582), which I would take as non-transient?

neiljp:
Here's an updated set of commits:
* hopefully in an order which is easier to merge incrementally
* with some `README.md` amendments
* with renames/moves after most of the matrix changes
* with matrix tests at the end

Note that the order of commits is not as shown above, as is typical for github after re-pushing.

The tests aren't in a test script at the moment, since I've been developing using `pytest` and they run fine with that :) I can place those in another PR along with a runner if that makes it easier.

timabbott:
Cool, this is great; I merged the first 8 commits (which is enough that we should do a PR to the zulip.git docs ASAP with the new config format).

The `--from-zuliprc` option doesn't work for me:
```
  File "./zulip/integrations/matrix/matrix_bridge.py", line 315, in <module>
    main()
  File "./zulip/integrations/matrix/matrix_bridge.py", line 257, in main
    write_sample_config(options.sample_config, options.zuliprc)
  File "./zulip/integrations/matrix/matrix_bridge.py", line 238, in write_sample_config
    sample_dict['zulip']['email'] = zuliprc_config['api']['email']
  File "/usr/lib/python3.5/configparser.py", line 956, in __getitem__
    raise KeyError(key)
KeyError: 'api'
```

We should also be sure to use `six.moves.configparser`; not sure if that's related to this issue.

neiljp:
@timabbott My guess is that your zuliprc file did not exist? I've updated that commit with a check for this condition. I'm running it as something similar to the following:
```
zulip-matrix-bridge  --from-zuliprc ~/zuliprc --write-sample-config /tmp/foo`
```

I also added a commit to add a separate test for this case, and another to separate some of the `sys.exit` calls from outputting the error via `print`, to enable that test (and potentially others).

The rebased version should be an improvement :)

neiljp:
Well, as above, I'm running the command using `~/zuliprc`, so that seems unlikely to be the issue?

neiljp:
@timabbott I appreciate you're travelling, but since the sample config code is not checked yet, I wondered if I reorder these tests then we could commit those? They run with pytest, and I can add a runner pretty easily.

@showell FYI

I'd like to add some extra tests on the core functionality too, after the bug report on czo#general.

showell:
@neiljp I'm not really sure what you're asking about here, since I don't have a ton of context on the bridges changes, but if you have some commits that you are eager to get merged to un-block other work, I can try to help.  This seems like a long PR, though, so I'm a bit wary.

neiljp:
@showell I'm mainly wanting to get the tests merged, since I'd like to get more coverage there. The tests from here (and a few more) are now in #437, though currently the integrated runner doesn't run them (`test-zulip`), though it does with the changes from #431.

rheaparekh:
I just tested this PR by creating a bridge and sending messages over it. I used all the config parsers (` --from-zuliprc`, ` --write-sample-config`, `-c`) and they seem to work fine. 

For the first commit `bridges: Bring inter-realm (zulip) bridge into zulip/ & rename others.`, I had a suggestion. Would it be a good idea to create a new directory `bridges` and have the folder `irc`, `matrix`, `zulip-interrealm`? I think this would be more clear than having `bridge-with-*`.

For the third commit `zulip: Update setup.py to install matrix bridge & add dependencies.`, as @timabbott mentioned, there is still a pending decision whether `matrix_client` should be added as a dependancy.

Other than this, the second, fourth and fifth commits look fine to me. I'll look at the tests in the other PR. I think as a priority the fourth commit `matrix bridge: Add optional from-zuliprc to use with write-sample-config.` should be merged.

neiljp:
* Good to hear your experience with the parsers/config-files was positive :)
* This was part of the discussion on czo, and we discussed various options; this can certainly be adjusted if that is preferable
* I agree that the dependency on matrix is pending; all of the bridges have dependency files (I actually missed updating the matrix one, it seems), but if we want the matrix (or others) to be installed as scripts then we need dependency installation. Perhaps one approach could be a sort of common 'provision' option that each bridge/integration can use to optionally install the dependencies required? Then the core `zulip` package doesn't depend on eg. `matrix_client`, but `zulip-matrix-bridge --provision` would install it?
* I just reread and I missed @timabbott wanting `six` for `configparser` in the current code; it's probably best to get the simpler and less 'controversial' #437 merged first, then I'll rebase this and make that change then.

neiljp:
@timabbott Rebased and amended as per your requests.

I'm unsure if we need configparser as a dependency now?

There are not currently any extra docs PRs, but they will need changing slightly in line with the change to the config-file generation.

timabbott:
I removed the explicit `configparser` and merged this.  Please do the documentation update soon.  I imagine we'll want to advertise the IRC/Matrix integration in our Zulip 1.9 release notes, so we should be sure the docs work :).  Thanks @neiljp!

PR #415, opened 2018-06-02
#414 Bot config file is not parsed 

### LINKED ISSUE #414: Bot config file is not parsed
There is bug in parsing bot's config file in `lib.py` on 202th line:
```
config_parser.read(conf)
```
Must be changed to `config_parser.read_file(conf)`

novokrest:
Fixes #414.

### DIFF
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -199,7 +199,7 @@ class ExternalBotHandler(object):
             config_parser = configparser.ConfigParser()
             with open(self.bot_config_file) as conf:
                 try:
-                    config_parser.read(conf)
+                    config_parser.read_file(conf)
                 except configparser.Error as e:
                     display_config_file_errors(str(e), self.bot_config_file)
                     sys.exit(1)


timabbott:
Do you understand when this was introduced?  This code path definitely used to work...

Also, your commit message looks unfinished?

novokrest:
@timabbott I updated commit's message. As it seems from git history, the bug was introduced after some refactoring. Previously, config file was read by `ConfigParser.readfp` method. After refactoring, used method was changed to `ConfigParser.read` method. But this method should be used with file's path passed as argument, not open file handle. Now if you try to run bot with custom config (e.g. beeminder), the bot's section would not be read from the config file. This PR fixes current issue.

timabbott:
Merged as 229f62a483501572884aeb8a39e8992f80758c34, after cleaning up the commit message.  Please write your commit messages with this sort of explanation and research done; it saves a ton of time in review, and can often save even more time when future developers try to understand why we did something.  Check out the Zulip commit message guidelines for more details on how you should be writing these: https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages

Thanks @novokrest!

PR #416, opened 2018-06-02
#122 Get `beeminder` bot to 100% test coverage

novokrest:
Related to #122.


timabbott:
@novokrest your commit messages should explain in the body of the commit message why the change is the right thing to do.  So " Remove duplicated try-except in `initialize` method. " is great, but it's totally unclear to me why we should prefer the `0` value over `5`?

novokrest:
@timabbott Added another PR #418 with modified logic for checking `auth_token`

PR #418, opened 2018-06-02
beeminder: Check auth_token in `initialize` by idempotent request.

novokrest:
Check `auth_token` without changing goal's datapoints.


timabbott:
@novokrest can you debug the test failure?

novokrest:
@timabbott I updated PR. Unfortunately, method `assert_called_once` was introduced in `unittest` only from Python 3.6. So, I modified assert condition to be compatible with Python 3.5 and 3.6.

PR #420, opened 2018-06-04
[WIP] Enable pytest as test-runner in API repo (test-bots only so far)

neiljp:
This is a first step to explore how well pytest might work for the API repo in general; currently it only updates `test-bots`, though it is by far the more complex case of the test scripts.

In addition to using the test scripts explicitly, with pytest installed one can also run `pytest` or `pytest <path-to-tests>`, so `test-bots -p` is broadly equivalent to `pytest zulip_bots/zulip_bots/bots` except for some default settings. These default settings could be extracted into `pytest.ini` or similar files in an updated version of this commit or later. However, having the scripts too may remain useful if we adopt pytest.

The first commit uses test skipping to avoid testing the `BotTestCase` versions of the test functions; I've not found a way around this for pytest, but wanted a level playing field between unittest and pytest for the migration, so this commit also removes the filtering in `test-bots` which otherwise removed these tests from the test-suite.

The pytest case does not require the presence of `__init__.py` files unlike the unittest discovery process, so `test-bots` currently excludes the `merels` bot tests explicitly.

**Potential remaining issues**:
a) DIffering test discovery
* On `master`, `test-bots` claims 302 tests of which 4 are skipped (via unittest), those 4 being `monkeytestit`, leaving 298 passing tests.
* On this branch, `test-bots` claims 382 tests of which 84 are skipped (via unittest), leaving 298 passing tests.
* On this branch, `test-bots -p` claims 370 items of which 82 are skipped (via pytest), and 288 pass.

b) pytest appears to be slower in discovery ('collection'), at least currently, leading to longer (*reported*) run-times to run all tests:
* `test-bots`: 4.25, 4.43, 4.46, 4.58, 4.82, 4.93, 6.70
* `test-bots -p`: 8.49, 9.84, 10.92, 11.36, 11.40, 11.40, 13.98

**Overview**
Remaining issues notwithstanding, personally the experience with pytest is markedly better, including:
* easier style of pytest test writing (reduced verbosity, fixtures)
* test output clarity (eg. obvious that monkeytestit is where the multiple skips are)
* features (eg. only re-test last failure)

Of course, currently this just uses pytest as a test-runner, so we don't benefit from the first point.

I would love to hear what other people think!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,7 @@ twine
 coverage>=4.4.1
 pycodestyle==2.3.1
 mock
+pytest
 -e ./zulip
 -e ./zulip_bots
 -e ./zulip_botserver
--- a/tools/test-bots
+++ b/tools/test-bots
@@ -9,6 +9,7 @@ import argparse
 import glob
 import unittest
 from unittest import TestCase, TestSuite
+import pytest
 
 def parse_args():
     description = """
@@ -50,6 +51,14 @@ the tests for xkcd and wikipedia bots):
                         default=False,
                         action="store_true",
                         help="whether to exit if a bot has tests which won't run due to no __init__.py")
+    parser.add_argument('--pytest', '-p',
+                        default=False,
+                        action='store_true',
+                        help="run tests with pytest")
+    parser.add_argument('--verbose', '-v',
+                        default=False,
+                        action='store_true',
+                        help='show verbose output (with pytest)')
     return parser.parse_args()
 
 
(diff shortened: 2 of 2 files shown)


timabbott:
What's the story with the extra 84 skipped tests on this branch?

neiljp:
This actually relates to the 'base class' tests, as discussed recently on czo!

In `master` they are expressly filtered out, but I've not determined if that's possible with `pytest`, so had to add a 'skip' for them. To compare equally, I removed the filter from the `unittest` case too.

timabbott:
Hmm.  Maybe I'm missing something, but could we make the base class an abstract class (without direct tests on it) and then move the tests for the base class to a subclass that isn't the one every bot is subclassing?

neiljp:
To clarify, the current structure can be reduced to: (identifiers adjusted)
```
class BaseTest(unittest.TestCase):
    def util_func():  # some multiple of this
        ...
    def test_func_every_bot_should_test_for():  # some multiple of this
        ...
class TestSomeBot(BaseTest):
    ...
```
The base `test_*` methods are currently to ensure that
a) the tested class has usage() (not so helpful IMO, but could be extended to check more generally)
b) the tested_class responds to an empty message
These can be overridden/extended in `TestSomeBot`.

As per the discussion on czo, we can have:
```
class BaseTest:
    ...  # as before
class TestSomeBot(unittest.TestCase, BaseTest):
    ... # as before
```
This seems reasonable, though a little more tricky due to the multiple inheritance; it seems to work, at least, ie. avoiding the skips :)

My exploration referenced on czo was more detailed than that, wondering if the util functions could be isolated from the inherited tests as a helper class that isn't inherited. I think you're saying something similar but with multiple inheritance again? So something like:
```
class BaseTest(unittest.TestCase):
    def util_func():  # some multiple of this
        ...
class DefaultTests:
    def test_func_every_bot_should_test_for():  # some multiple of this
        ...
class TestSomeBot(BaseTest, DefaultTests):
    ...
```
We could maybe add a lint rule to ensure that the DefaultTests are included, and specified correctly in the Test class?

PR #422, opened 2018-06-04
bots: Improve response at 3 wrong answers in trivia bot.

rheaparekh:
I have introduced two variables `answered_options` and `ended` in the bot storage. Initially `ended` is set to False. 
When we have start answering, the options we answer are stored in `answered_options`. After three wrong failure we get the message `Three failed attempts at <quiz_id>:disappointed:\nThe correct answer is <answer>." at three wrong answers.`. After that, `ended` is being set to `True`. 
Now if the user continues playing with the same question, they won't get the three failed attempts message, and they can continue answering the question.

Hence we would receive the 3-failed attempts message from the bot only for the first time.

### DIFF
--- a/zulip_bots/zulip_bots/bots/trivia_quiz/test_trivia_quiz.py
+++ b/zulip_bots/zulip_bots/bots/trivia_quiz/test_trivia_quiz.py
@@ -2,12 +2,13 @@ import json
 import html
 
 from unittest.mock import patch
-from typing import Optional
+from typing import Optional, Tuple, Any, Dict
 
 from zulip_bots.test_lib import (
     BotTestCase,
     DefaultTests,
     read_bot_fixture_data,
+    StubBotHandler,
 )
 
 from zulip_bots.request_test_lib import (
@@ -17,6 +18,9 @@ from zulip_bots.request_test_lib import (
 from zulip_bots.bots.trivia_quiz.trivia_quiz import (
     get_quiz_from_payload,
     fix_quotes,
+    get_quiz_from_id,
+    update_quiz,
+    handle_answer,
 )
 
 class TestTriviaQuizBot(BotTestCase, DefaultTests):
@@ -28,6 +32,13 @@ class TestTriviaQuizBot(BotTestCase, DefaultTests):
         '* **C** Reptiles\n' + \
         '* **D** Mammals\n' + \
         '**reply**: answer Q001 <letter>'
+
+    def get_test_quiz(self) -> Tuple[Dict[str, Any], Any]:
+        bot_handler = StubBotHandler()
+        quiz_payload = read_bot_fixture_data('trivia_quiz', 'test_new_question')['response']
+        with patch('random.shuffle'):
+            quiz = get_quiz_from_payload(quiz_payload)
+        return quiz, bot_handler
 
     def _test(self, message: str, response: str, fixture: Optional[str]=None) -> None:
         if fixture:
@@ -74,9 +85,12 @@ class TestTriviaQuizBot(BotTestCase, DefaultTests):
         with patch('random.shuffle'):
             quiz = get_quiz_from_payload(quiz_payload)
 
+        # Test initial storage
         self.assertEqual(quiz['question'], 'Which class of animals are newts members of?')
         self.assertEqual(quiz['correct_letter'], 'A')
         self.assertEqual(quiz['answers']['D'], 'Mammals')
+        self.assertEqual(quiz['answered_options'], [])
+        self.assertEqual(quiz['pending'], True)
 
         # test incorrect answer
         with patch('zulip_bots.bots.trivia_quiz.trivia_quiz.get_quiz_from_id',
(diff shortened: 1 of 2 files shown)


zulip_bots/zulip_bots/bots/trivia_quiz/trivia_quiz.py
```diff
@@ -198,14 +199,28 @@ def format_quiz_for_markdown(quiz_id, quiz):
...
 
-def grade_question(quiz, answer):
+def grade_question(quiz, answer, quiz_id, bot_handler):
     correct = (answer == quiz['correct_letter'])
+    long_answer = quiz['answers'][quiz['correct_letter']]
 
-    if correct:
-        long_answer = quiz['answers'][answer]
+    if correct and len(quiz['answered_options']) <= 2:
         response = '**CORRECT!** {long_answer} :tada:'.format(long_answer=long_answer)
```
showell: I think I would simplify this logic.  If you get a correct answer, don't worry about whether we already revealed the correct answer.  It's possible that somebody answered before seeing the correct answers.

zulip_bots/zulip_bots/bots/trivia_quiz/test_trivia_quiz.py
```diff
@@ -58,14 +60,55 @@ def test_answers(self) -> None:
...
             with patch('zulip_bots.bots.trivia_quiz.trivia_quiz.start_new_quiz') as mock_new_quiz:
                 self._test('answer Q001 A', '**CORRECT!** Amphibian :tada:')
+
+    def test_grade_questions(self) -> None:
+        bot_handler = StubBotHandler()
+        quiz_payload = read_bot_fixture_data('trivia_quiz', 'test_new_question')['response']
+        with patch('random.shuffle'):
+            quiz = get_quiz_from_payload(quiz_payload)
+
+        bot_handler.storage.put('Q001', json.dumps(quiz))
```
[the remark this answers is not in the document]
showell: For this kind of test, where the sequence of operations is pretty important, I think a long test is fine, so I'd lean toward Neil's first advice, which is to use comments to highlight the steps of the test.  Sometimes a a really good helper function inside the test can also help, so that the end of the test reads a little quicker.

zulip_bots/zulip_bots/bots/trivia_quiz/trivia_quiz.py
```diff
@@ -198,14 +199,28 @@ def format_quiz_for_markdown(quiz_id, quiz):
     )
     return content
 
-def grade_question(quiz, answer):
+def grade_question(quiz, answer, quiz_id, bot_handler):
```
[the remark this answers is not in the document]
showell: The original name here was my choosing, and I wasn't very happy with it even before the new features.  I think `handle_answer` would be better at this point.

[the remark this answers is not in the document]
neiljp: That would be an improvement :+1: If it stays in one function, what do you think about clarifying the control flow to return from one place? `correct` is still accurate for the early variable, but each branch of logic could set a value for `start_new_quiz` or similar, before returning it and `response` near the end?

zulip_bots/zulip_bots/bots/trivia_quiz/test_trivia_quiz.py
```diff
@@ -69,3 +74,49 @@ def test_answers(self) -> None:
...
+
+        # test for an incorrect answer
+        start_new_question, response = handle_answer(quiz, 'D', 'Q001', bot_handler)
+        self.assertFalse(start_new_question)
+        self.assertEqual(response, '**WRONG!** D is not correct :disappointed:')
+
+        # create test storage for a question which has been incorrectly answered twice
+        quiz['answered_options'].append('C')
+        quiz['answered_options'].append('B')
+        bot_handler.storage.put('Q001', json.dumps(quiz))
```
neiljp: Now you've added `update_quiz`, maybe test that first, and then use it in this test method? Is that reasonable, @showell ?

zulip_bots/zulip_bots/bots/trivia_quiz/test_trivia_quiz.py
```diff
@@ -69,3 +74,49 @@ def test_answers(self) -> None:
...
+        # test response after question has ended
+        start_new_question, response = handle_answer(quiz, 'D', 'Q001', bot_handler)
+        self.assertEqual(response, '**WRONG!** D is not correct :disappointed:')
+        self.assertFalse(start_new_question)
+        start_new_question, response = handle_answer(quiz, 'A', 'Q001', bot_handler)
+        self.assertEqual(response, '**CORRECT!** Amphibian :tada:')
+        self.assertFalse(start_new_question)
+        start_new_question, response = handle_answer(quiz, 'B', 'Q001', bot_handler)
+        self.assertEqual(response, '**WRONG!** B is not correct :disappointed:')
+
```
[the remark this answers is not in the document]
rheaparekh: I didn't add it because it's already been tested in the above line. But now I feel it'll be better to add it.

showell:
For the testing stuff, I think my main message is that it's ok for a test to be kind of long if it's testing a sequence of operations that are cohesive, but I also agree that splitting out tests for completely different scenarios is good, and if there's a lot of common setup, you can usually address that with a helper.

rheaparekh:
@showell I have updated this

rheaparekh:
Added #436 for the bot cleanups.

zulip_bots/zulip_bots/bots/trivia_quiz/trivia_quiz.py
```diff
@@ -198,16 +200,41 @@ def format_quiz_for_markdown(quiz_id, quiz):
...
+            # we reset this so that the users can keep answering a question after it has ended
+            quiz['answered_options'] = []
+            quiz['ended'] = True
+            update_quiz(quiz, quiz_id, bot_handler)
+        else:
+            response = '**WRONG!** {answer} is not correct :disappointed:'
 
-    response = '**WRONG!** {answer} is not correct :disappointed:'.format(answer=answer)
-    return correct, response
+    response = response.format(answer=answer, long_answer=long_answer, quiz_id=quiz_id)
```
neiljp: MINOR: I appreciate this is long and maybe it wouldn't fit on one line, but I would combine these two lines either:
* splitting the two tuple components, if the line is overly long
* shorten the format keys or values slightly

zulip_bots/zulip_bots/bots/trivia_quiz/trivia_quiz.py
```diff
@@ -198,16 +200,41 @@ def format_quiz_for_markdown(quiz_id, quiz):
...
+    if answer == quiz['correct_letter']:
+        response = '**CORRECT!** {long_answer} :tada:'
+        # We don't start a new question if the question has already been
+        # answered
+        start_new_question = not quiz['ended']
+    else:
+        if len(quiz['answered_options']) == 3:
+            # We reveal the answer and start a new question upon having three failed
+            # attempts at answering a question
+            response = '''Three failed attempts at {quiz_id}:disappointed:
```
neiljp: MINOR: This is personal style (I prefer to avoid triple-quotes, at least when wrapping onto the next line), so does not require changing, but you could consider exploring other styles such as:
* multiple single-quoted strings, grouped with parentheses into one value and enabling splitting over lines;
* If the `"\n"` become too much noise, then I favour using `.join` on a list of strings, each to be a separate line.

neiljp:
OK, so since @showell merged the other, feel free to explore the above here, or in another PR if it seems like a definite change, after rebasing.

zulip_bots/zulip_bots/bots/trivia_quiz/trivia_quiz.py
rheaparekh: I followed the triple quotes this has been followed throughout the script and I wanted to be consistent. Here is the [example](https://github.com/zulip/python-zulip-api/blob/master/zulip_bots/zulip_bots/bots/trivia_quiz/trivia_quiz.py#L188-L196).

rheaparekh:
@showell @neiljp Updated this again :)

zulip_bots/zulip_bots/bots/trivia_quiz/trivia_quiz.py
```diff
@@ -43,7 +43,7 @@ def handle_message(self, message: Dict[str, Any], bot_handler: Any) -> None:
                 bot_handler.send_reply(message, bot_response)
                 return
             quiz = json.loads(quiz_payload)
-            correct, bot_response = grade_question(quiz, answer)
+            correct, bot_response = handle_answer(quiz, answer, quiz_id, bot_handler)
```
showell: We should rename `correct` to `start_new_question` here, since the semantics of the variable have changed to that.

zulip_bots/zulip_bots/bots/trivia_quiz/trivia_quiz.py
```diff
@@ -196,16 +198,40 @@ def format_quiz_for_markdown(quiz_id: str, quiz: Dict[str, Any]) -> str:
...
+        # answered
+        start_new_question = not quiz['ended']
+    else:
+        if len(quiz['answered_options']) == 3:
+            # We reveal the answer and start a new question upon having three failed
+            # attempts at answering a question
+            response = '''Three failed attempts at {id}:disappointed:
+The correct answer is {answer}.'''
+            start_new_question = True
+            # we reset this so that the users can keep answering a question after it has ended
```
showell: I think there's a simpler approach than clearing `answered_options` to allowing people to answer.  You don't really need to clear the data structure.  Instead, just accept answers forever, and separate your concerns:

* correct/disappointed -- just see if they answered the question right
* start new question? -- (answer_correct or >= 3 answers) and (still_active)
* update `answered_options` - just always do it unless answer has already been given

showell: Here's my suggestion of how to write it, and note that I make a little helper function that just takes care of message formatting (you'll have to flesh it out a bit, of course);

~~~ py
def build_response(is_correct, start_new_question, num_answers):
    if is_correct:
        response = '**CORRECT** ...'
    else:
        if num_answers >= 3:
            response = 'disappointed...correct answers is X'
        else:
            response = 'disappointed'

def handle_answer(...):
    is_new_answer = (option not in quiz['answered_options']
    if is_new_answer:
        quiz['answered_options'].append(new_answer)

    num_answers = len(quiz('answered_options'])

    is_correct = (option == quiz['correct_letter'])

    start_new_question = quiz['pending'] and (is_correct or num_answers >= 3)
    if start_new_question:
        quiz['pending'] = False

    if is_new_answer or start_new_question:
        update_quiz(quiz, quiz_id, bot_handler)

    response = build_response(is_correct, start_new_question):

    return start_new_question, response

rheaparekh: Right, I was clearing `answered_options` up everytime because I wanted two different error messages. I wanted the three failed attempt warning to appear only once, and all attempts after that all the warnings were same as before. But here, since we are changing the warning and just revealing the correct answer, so it would work.

rheaparekh:
@showell I have updated this according to recommended logic!

showell:
Merged, thanks @rheaparekh!  Can you run the latest code pointing at czo?  If we see any quirks, we can fix forward.

rheaparekh:
@neiljp @showell Thank you for the reviews :)

PR #423, opened 2018-06-04
bot tests: Allow raw responses in fixtures.

roberthoenig:
Previously, the responses set in bot test fixtures
where handled as JSON objects. This works fine for
most bot tests, because ost of the APIs that bots
are calling return a JSON-formatted response object.
However, some, like Trello, do return raw data.
This hasn't been noticed so far, because the respective
Trello test needed internet access. Tests shouldn't
need internet access.
This commit makes that Trello test use a fixture. To
work properly, it also adds a way to make http_mock_config
parse the response object as raw data.
This can now be done by modifying the "is_raw_response"
property in a newly introduced "meta" object that can
be used to specify how a fixture should be handled.

@timabbott this should fix #411.

### DIFF
--- a/zulip_bots/zulip_bots/bots/trello/fixtures/invalid_key.json
+++ b/zulip_bots/zulip_bots/bots/trello/fixtures/invalid_key.json
@@ -1,0 +1,15 @@
+{
+  "meta": {
+    "is_raw_response": true
+  },
+  "request": {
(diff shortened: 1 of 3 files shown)


timabbott:
Merged after fixing "ost" in the commit message, thanks @roberthoenig!

PR #424, opened 2018-06-04
zulip_bots: Clean up the bots dependency installation process.

eeshangarg:
@timabbott: FYI

### DIFF
--- a/tools/provision
+++ b/tools/provision
@@ -4,6 +4,7 @@ import os
 import sys
 import argparse
 import subprocess
+import glob
 
 from importlib import import_module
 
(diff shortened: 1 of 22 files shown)


timabbott:
Merged, thanks @eeshangarg!  @roberthoenig FYI, since this was something you worked on.

roberthoenig:
Yeah, I wonder about the motivation behind removing all these requirements.txt files? I didn't feel they were cluttering bots, and always having a requirements.txt, even if it's empty, feels clearer and more explicit than only having a requirements.txt if the bot has third party dependencies but not for dep x because x gets installed with the `zulip` package and not for y  and z because y and z already come with the zulip_bots package.

timabbott:
Well, for `requests`, they were just redundant, and my perspective was that if only 20% of our bots need a `requirements.txt`, it's probably better to not add an extra step in writing a bot (of writing an empty `requirements.txt` file for it).  We have enough trouble with folks forgetting to commit their `__init__.py` file.

PR #425, opened 2018-06-04
#122 Get `dropbox_share` bot to 100% test coverage 

novokrest:
Related to #122.


zulip_bots/zulip_bots/bots/dropbox_share/test_dropbox_share.py
```diff
@@ -36,6 +51,9 @@ def search_files(*args, **kwargs):
         )
     ])
 
+def empty_search(*args, **kwargs):
```
aero31aero: Minor nitpicking, it should probably be `get_empty_search_result` looking at how other functions are named.

novokrest: @aero31aero Renamed function to `get_empty_search_result`

PR #430, opened 2018-06-09
#122 Get `front` bot to 100% test coverage

novokrest:
Related to #122.

### DIFF
--- a/zulip_bots/zulip_bots/bots/front/test_front.py
+++ b/zulip_bots/zulip_bots/bots/front/test_front.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 from zulip_bots.test_lib import BotTestCase, DefaultTests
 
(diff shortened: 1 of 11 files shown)


zulip_bots/zulip_bots/bots/front/test_front.py
```diff
@@ -28,27 +34,53 @@ def test_archive(self) -> None:
             with self.mock_http_conversation('archive'):
                 self.verify_reply('archive', "Conversation was archived.")
 
+    def test_archive_error(self) -> None:
```
neiljp: This increase in coverage is good to see, but a lot of these tests seem very similar; perhaps consider a helper function? Each test is very short, so is legible as it is, but the patterns suggest refactoring, particularly with all the error replies being identical.

showell: I agree that a helper function is worthwhile here.  It's fine to just do this an additional commit that sweeps all the functions that follow the same pattern and does something more like `assert_bot_error(fixture='archive_error', message='Something went wrong')`

novokrest: @neiljp I introduced helper function `_test_command_error` to test command's error.

showell:
Merged, thanks @novokrest!

PR #431, opened 2018-06-09
Tests: Simplify & enable use of pytest to run all tests

neiljp:
This is a follow-up to #420, which completes the migration started there and tidies up the tests slightly.

Each test script run from `test-main`, and that script too, now support passing in `-p` to run via pytest, and also with `-v` to show each test.

I'm unsure if we'd like to see `-v` mode as the default; its output is similar to that from `unittest`, but can be verbose if the tests pass :)

`test-lib` is simplified a lot by using the common `test_handler` as in the other tests, with an amendment to that function, also comprising a header to nicely divide the test output.

If this all looks good, the pytest options can be factored out into a configuration file, but they're only in two locations right now.

### DIFF
--- a/tools/server_lib/test_handler.py
+++ b/tools/server_lib/test_handler.py
@@ -9,8 +9,8 @@ import pytest
 TOOLS_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 os.chdir(os.path.dirname(TOOLS_DIR))
 
-def handle_input_and_run_tests_for_package(package_name):
-    parser = argparse.ArgumentParser(description="Run tests for the {} package.".format(package_name))
+def handle_input_and_run_tests_for_package(package_name, path_list):
+    parser = argparse.ArgumentParser(description="Run tests for {}.".format(package_name))
     parser.add_argument('--coverage',
                         nargs='?',
                         const=True,
@@ -34,7 +34,7 @@ def handle_input_and_run_tests_for_package(package_name):
         cov.start()
 
     if options.pytest:
-        location_to_run_in = os.path.join(TOOLS_DIR, '..', package_name)
+        location_to_run_in = os.path.join(TOOLS_DIR, '..', *path_list)
         paths_to_test = ['.']
         pytest_options = [
             '-s',    # show output from tests; this hides the progress bar though
@@ -50,7 +50,7 @@ def handle_input_and_run_tests_for_package(package_name):
     else:
         # Codecov seems to work only when using loader.discover. It failed to capture line executions
         # for functions like loader.loadTestFromModule or loader.loadTestFromNames.
-        test_suites = unittest.defaultTestLoader.discover(package_name)
+        test_suites = unittest.defaultTestLoader.discover(os.path.join(*path_list))
         suite = unittest.TestSuite(test_suites)
         runner = unittest.TextTestRunner(verbosity=2)
         result = runner.run(suite)
--- a/tools/server_lib/test_handler.py
+++ b/tools/server_lib/test_handler.py
@@ -4,6 +4,7 @@ import os
 import sys
 import argparse
 import unittest
+import pytest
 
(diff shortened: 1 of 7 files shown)


neiljp:
For reference to the issue raised on #420, namely the difference in the number of tests run:
* With the merger of the pytest no-skipping version, I currently have 337 (unittest) and 314 (pytest)
* As discussed on czo in #integrations, the issue appears to be 23 duplicated tests for `dropbox_share`
* This occurs due to the globbing picking up multiple `test_*.py` files in the bot folder, which are then translated into multiple directory entries.
I've added a commit which takes the bot names into a `set`; this could have been applied around the previous code, but seems neater as a set comprehension.

tools/server_lib/test_handler.py
```diff
@@ -4,36 +4,66 @@
...
+                        help="run tests with pytest")
+    parser.add_argument('--verbose', '-v',
+                        default=False,
+                        action='store_true',
+                        help='show verbose output (with pytest)')
     options = parser.parse_args()
 
+    test_session_title = ' Running tests for {} '.format(package_name)
+    header = test_session_title.center(shutil.get_terminal_size().columns, '#')
+    print(header)
```
eeshangarg: What is this `print(` call for?

tools/test-lib
```diff
@@ -1,51 +1,6 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
```
eeshangarg: Any particular reason, why we should not be using python3 here?

tools/server_lib/test_handler.py
neiljp: This outputs a clean header (separator) between each set of tests, using the previous two lines, of the form
```
####...#### Running tests for API ##########...##########
```
centered across the terminal.

tools/test-lib
neiljp: It's been a while since I looked into this, but my recollection was we have a lint rule against it.

For reference, currently we get:
```
$ git grep /env|grep tools|grep -v bash
tools/custom_check.py:        {'pattern': '#!/usr/bin/env python$',
tools/custom_check.py:         'description': 'Fix shebang line with proper call to /usr/bin/env for Bash path, change -x|-e switches'
tools/deploy:#!/usr/bin/env python
tools/lint:#! /usr/bin/env python
tools/provision:#!/usr/bin/env python3
tools/release-packages:#!/usr/bin/env python3
tools/review:#!/usr/bin/env python
tools/run-mypy:#!/usr/bin/env python
tools/server_lib/lister.py:#!/usr/bin/env python
tools/test-bots:#!/usr/bin/env python3
tools/test-botserver:#!/usr/bin/env python
tools/test-lib:#!/usr/bin/env python3
tools/test-zulip:#!/usr/bin/env python
```

tools/server_lib/test_handler.py
eeshangarg: Oh niice thanks for explaining! :)

tools/test-lib
eeshangarg: Not sure what you mean? Do we have a rule against using python3? It I feel like since `tools/provision` uses python3, all of the other scripts should use python3 as well, since the virtualenv that `tools/provision` creates is a python3 env. What do you think?

neiljp: This does make sense, particularly given that all but the `zulip/` component should be python3 now? That said, there may be a separate recommendation in terms of this use.

showell:
We definitely want to use version 3 of Python everywhere, and then the question is what it's named as (python vs. python3) in various environments we use.  I'd err on the side of explicit `python3`, and if we find out it breaks something, at least we'll know. 😄

showell:
merged, thanks @neiljp!

We'll want to be pretty vigilant about any bug reports with this new system.

neiljp:
@showell I would like to hear about any issues we find! Though as per on CZO, this isn't enabled unless using `-p`.

PR #432, opened 2018-06-10
Bot tests: Various tidying & minor improvements

neiljp:
* Improve typing in a few cases
* Remove quite a few unnecessary imports
* Use consistent imports within test files and across bots; the latter isn't vital, but the code is simpler
* Resolve the issue with `monkeytestit` not being tested since it fails for python 3.4
* Name the merels test better and improve it's style (though it is not currently being run)


showell:
Merged, thanks @neiljp!

PR #435, opened 2018-06-11
#122 Get `jira` bot to 100% test coverage

novokrest:
Related to #122.

Also made some refactoring for old tests.

### DIFF
--- a/zulip_bots/zulip_bots/bots/jira/test_jira.py
+++ b/zulip_bots/zulip_bots/bots/jira/test_jira.py
@@ -10,18 +10,6 @@ class TestJiraBot(BotTestCase, DefaultTests):
         'domain': 'example.atlassian.net'
     }
 
-    MOCK_GET_JSON = {
-        'fields': {
-            'creator': {'name': 'admin'},
-            'description': 'description',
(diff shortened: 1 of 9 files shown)


zulip_bots/zulip_bots/bots/jira/test_jira.py
```diff
@@ -46,6 +46,112 @@ class TestJiraBot(BotTestCase, DefaultTests):
 
     MOCK_NOTHING_RESPONSE = 'Sorry, I don\'t understand that! Send me `help` for instructions.'
 
+    MOCK_HELP_RESPONSE = '''
```
showell: I generally like the technique of just copying the help text into the test to validate the help command, but this help text is pretty enormous.  I'm not sure how I feel about it.  Part of the problem here is that the help text is probably way too long for human users, too, but that's not something we need to fix in this PR.  Maybe @neiljp or somebody can give a second opinion here.

novokrest: @neiljp Could you review this PR ?

showell:
Merged, thanks @novokrest!

PR #437, opened 2018-06-17
Matrix tests: Initial tests for the matrix bridge, from #413 & beyond

neiljp:
The test files are in the `zulip` tree, but are not discovered by the `unittest` discover function at present.

The test runner in #431 with pytest will discover and run these tests, or simply using `pytest` as a command (or `pytest <dir>`) will work too.

### DIFF
--- a/zulip/integrations/matrix/test_matrix.py
+++ b/zulip/integrations/matrix/test_matrix.py
@@ -1,0 +1,35 @@
+from unittest import TestCase
+from subprocess import Popen, PIPE
+import os
+
+script_file = "matrix_bridge.py"
+script_dir = os.path.dirname(__file__)
+script = os.path.join(script_dir, script_file)
+
+
+class MatrixBridgeTests(TestCase):
+    def test_no_args(self):
+        # type: () -> None
+        popen = Popen(["python", script], stdin=PIPE, stdout=PIPE, universal_newlines=True)
+        output_lines = popen.communicate()[0].strip().split("\n")
+        expected_lines = [
+            "Options required: -c or --config to run, OR --write-sample-config.",
+            "usage: {} [-h]".format(script_file)
+        ]
+        for expected, output in zip(expected_lines, output_lines):
+            self.assertIn(expected, output)
+
+    def test_help_usage_and_description(self):
+        # type: () -> None
+        popen = Popen(["python", script] + ["-h"], stdin=PIPE, stdout=PIPE, universal_newlines=True)
+        output_lines = popen.communicate()[0].strip().split("\n")
+        usage = "usage: {} [-h]".format(script_file)
+        description = "Script to bridge"
+        self.assertIn(usage, output_lines[0])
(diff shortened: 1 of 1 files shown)


zulip/integrations/matrix/test_matrix.py
```diff
@@ -0,0 +1,32 @@
...
+            self.assertIn(expected, output)
+
+    def test_help_usage_and_description(self):
+        # type: () -> None
+        popen = Popen(["python", script] + ["-h"], stdin=PIPE, stdout=PIPE, universal_newlines=True)
+        output_lines = popen.communicate()[0].strip().split("\n")
+        usage = "usage: {} [-h]".format(script_file)
+        description = "Script to bridge"
+        self.assertIn(usage, output_lines[0])
+        blank_lines = [num for num, line in enumerate(output_lines) if line == '']
```
rheaparekh: I don't understand the need to test `blank_lines.`

zulip/integrations/matrix/test_matrix.py
```diff
@@ -73,3 +75,19 @@ def test_write_sample_config(self):
...
+        zulip_config = dict(stream="some stream",
+                            topic="some topic",
+                            email="some@email")
+        msg = dict(
+            sender_email="John@Smith.smith",  # must not be equal to config:email
+            type="stream",  # Can only mirror Zulip streams
+            display_recipient=zulip_config['stream'],
+            subject=zulip_config['topic']
+        )
+        assert msg['sender_email'] != zulip_config['email']
```
rheaparekh: I don't think this is needed, as the `msg` and `zulip_config` are just defined above.

zulip/integrations/matrix/test_matrix.py
```diff
@@ -78,40 +78,37 @@ def test_write_sample_config(self):
...
+        stream="some stream",
+        topic="some topic",
+        email="some@email"
+    )
+    valid_msg = dict(
+        sender_email="John@Smith.smith",  # must not be equal to config:email
+        type="stream",  # Can only mirror Zulip streams
+        display_recipient=valid_zulip_config['stream'],
+        subject=valid_zulip_config['topic']
+    )
```
rheaparekh: I feel that the previous commit and this commit should be squashed as this change is pretty minor.

zulip/integrations/matrix/test_matrix.py
neiljp: This is to ensure that they are different.

zulip/integrations/matrix/test_matrix.py
neiljp: The previous commit adds tests; this commit is a refactor. I would be happy to discuss, but I feel that these are separate.

neiljp:
Pushed updated commits:
* Added comments
* Amended code in last commit slightly
* Use unittest asserts more consistently

@timabbott I separated out the tests to make them easier to apply independently, given the slowness in applying the #413.

timabbott:
Merged, thanks @neiljp!

PR #438, opened 2018-06-21
api: Implement GET /messages.

YagoGG:
**Testing plan:** lint and mypy checks passed.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -686,6 +686,17 @@ class Client(object):
                 callback(event['message'])
         self.call_on_each_event(event_callback, ['message'])
 
+    def get_messages(self, message_filters):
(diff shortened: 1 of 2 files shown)


zulip/zulip/__init__.py
```diff
@@ -686,6 +686,17 @@ def event_callback(event):
                 callback(event['message'])
         self.call_on_each_event(event_callback, ['message'])
 
+    def get_messages(self, messages_filter):
```
eeshangarg: Should we call this `message_filters`?

eeshangarg:
Merged, thanks @YagoGG! :)

timabbott:
@YagoGG we should probably remove the now-obsolete `zulip/zulip/examples/recent-messages`, which this replaces...

PR #439, opened 2018-06-23
api: Followup for the get-messages example.

YagoGG:
The GET /messages action has been recently implemented into our bindings. Therefore this example has been superseded by get-messages, which is capable of fetch messages in a more fine-grained way.

Followup for #438.


eeshangarg:
@YagoGG: Merged, thanks! See my comment on #441 :)

PR #441, opened 2018-06-26
api: Implement DELETE /messages/<message_id>.

YagoGG:

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -728,6 +728,16 @@ class Client(object):
             request=message_data,
         )
 
+    def delete_message(self, message_id):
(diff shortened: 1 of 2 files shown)


eeshangarg:
@YagoGG: It might be time to think about documenting a couple of these in `templates/zerver/api/*` before we add any more bindings. We want to be sure that tests in the main repo pass before these bindings can be considered "correct". Does that make sense?

Apart from that, happy to merge! Thanks for working on this! :)

YagoGG:
Hey @eeshangarg! Thanks a lot for reviewing and merging :)

All of these are already documented, we just have to start merging them on the server repo. There is a small merging backlog that hopefully will be cleared over the next few days, but all of the endpoint implementations I have been adding here have their docs ready and will be available soon.

PR #442, opened 2018-06-26
api: Implement PATCH /streams/<stream_id>.

YagoGG:

About the quoting in the example, apparently the endpoint requires the strings to be quoted to be considered "valid JSON".


eeshangarg:
@YagoGG: Merged, thanks for working on this! :)

PR #444, opened 2018-06-27
 api: Implement GET /messages/<message_id>/history.

YagoGG:


eeshangarg:
@YagoGG: Merged, thanks! :) Also, let's have a conversation on `chat.zulip.org` about when we should make a new PyPI release. I think it might be time for another one, there's been a lot of activity in this area! :) Thanks for working on this!

PR #447, opened 2018-06-28
[wip] Begin zgram support.

showell:
This still needs significant cleanup and, **more importantly**, it shouldn't be merged till we merge some of the server pieces.

But the first raw commit demonstrates the basic protocol where bots can both receive and send zgrams.


showell:
See also: https://github.com/zulip/zulip/pull/9729

The next step here will be some explanation on how to test this out in dev branches.

showell:
The zgram effort has been abandoned.

PR #449, opened 2018-07-03
api: Implement POST /messages/flags.

YagoGG:


zulip/zulip/examples/update-flags
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python
```
timabbott: After merging this, I kinda think this should be `update-message-flags`, for clarity.

PR #454, opened 2018-07-11
Fix post requests for JIRA bot

jpellman:
JSON post requests aren't supposed to be form-encoded (see http://docs.python-requests.org/en/master/user/quickstart/#more-complicated-post-requests)

### DIFF
--- a/zulip_bots/zulip_bots/bots/jira/fixtures/test_create.json
+++ b/zulip_bots/zulip_bots/bots/jira/fixtures/test_create.json
@@ -3,10 +3,9 @@
     "api_url": "https://example.atlassian.net/rest/api/2/issue",
     "method": "POST",
     "headers": {
-      "Authorization": "Basic ZXhhbXBsZUBleGFtcGxlLmNvbTpxd2VydHkhMTIz",
-      "Content-type": "application/json"
+      "Authorization": "Basic ZXhhbXBsZUBleGFtcGxlLmNvbTpxd2VydHkhMTIz"
     },
-    "data": {
+    "json": {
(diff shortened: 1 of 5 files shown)


jpellman:
@timabbott I tested it against a non-cloud JIRA instance running JIRA 7.6.3.  I was using the version distributed alongside v 0.5.1 of the PyPI package (which would have used the a17bd77 version of the bot I believe).  

Prior to the changes the bot was returning the string "Issue ** is up! https://jirainstance.example.com/browse/" when typing in "create" followed by the prescribed usage, most likely because the 'key' variable wasn't being populated / there was no key in the JIRA response.

Having `'Content-type': 'application/json'` in the headers might be redundant too, since I thank that `json=` automatically adds that to the headers, but I haven't tried that.

jpellman:
Jirabot text fixtures have been updated.

jpellman:
Sure, I also verified that `'Content-type': 'application/json'` in the headers is, in fact, redundant w/ `json=` (i.e., the `requests` docs don't lie) so I'll go ahead and remove that too while I'm at it (and update fixtures if necessary), testing against a JIRA instance.

skunkmb:
@jpellman Thanks for changing that! I think that you now have to update the fixtures again, this time for the headers, now that you've removed the `'Content-type': 'application/json'`.

jpellman:
There we go- just 23 minutes shy of sundown!  If Travis obliges, I plan to celebrate with pancakes and sleep.

timabbott:
Merged, thanks @jpellman and @skunkmb!

PR #455, opened 2018-07-12
api: Implement POST /users/me/subscriptions/properties.

YagoGG:
**Testing plan**: lint and mypy checks passed locally, tried the example in the function's comment.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -950,6 +950,29 @@ class Client(object):
             request=request
         )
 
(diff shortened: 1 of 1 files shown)


timabbott:
Merged, thanks @YagoGG!  

Regarding example scripts, it's mostly a matter of expediency; for lower-importance endpoints, it's probably more important to have them in the Python API and documented than to have an example script.  And regardless, I'd prefer to not write 50 example scripts until we're confident exactly what naming conventions we want to use for installing them into users' PATHs.

PR #458, opened 2018-07-17
[WIP] Add giphy zgram handler.

shubhamdhama:
**[DON'T MERGE]**
Just created for testing and collaboration purpose

This would work with https://github.com/zulip/zulip/pull/9962 only

@rheaparekh FYI.

### DIFF
--- a/zulip_bots/zulip_bots/bots/giphy/giphy.py
+++ b/zulip_bots/zulip_bots/bots/giphy/giphy.py
@@ -55,6 +55,30 @@ class GiphyHandler(object):
 class GiphyNoResultException(Exception):
     pass
 
+def request_giphy(url: str, query: str) -> str:
+    try:
+        data = requests.get(url, params=query)
+    except requests.exceptions.ConnectionError as e:  # Usually triggered by bad connection.
+        logging.exception('Bad connection')
+        raise
+    data.raise_for_status()
+    return data
+
+def get_many_giphy_gifs(keyword: str, api_key: str, count: int) -> Union[int, str]:
+    url = GIPHY_SEARCH
+    query = dict(
+        api_key=api_key,
+        q=keyword,
+        limit=count,
+    )
+
+    data = request_giphy(url, query)
+
+    try:
+        gifs = [item['images']['original']['url'] for item in data.json()['data']]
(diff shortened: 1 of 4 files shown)


zulip_bots/zulip_bots/bots/giphy/giphy.py
```diff
@@ -75,8 +112,11 @@ def get_url_gif_giphy(keyword: str, api_key: str) -> Union[int, str]:
     data.raise_for_status()
 
     try:
-        gif_url = data.json()['data']['images']['original']['url']
-    except (TypeError, KeyError):  # Usually triggered by no result in Giphy.
+        if count == 1:
```
showell: I wasn't reviewing this, was just curious about progress.  This isn't directly related to your fix, but it would be nice if we could clean up this function so it doesn't have all the conditional branching on count-equals-one-vs.-otherwise.  It would be nice to avoid having two different return types.  I know we have to a conditional **somewhere** to handle the only-one and only-zero cases, but if we could organize the code to have that happen in as few places as possible, that would be nice.

zulip_bots/zulip_bots/bots/giphy/giphy.conf
```diff
@@ -1,2 +1,2 @@
 [giphy]
-key=12345678
+key=9Cg4s7nRdHNTCK9cBHQhx3PTBDTCl3YL
```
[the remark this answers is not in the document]
shubhamdhama: Not adding this to commit was in my brain till last moment... (sigh) I think my brain stopped functioning after 03:30 am :laughing: Just revoked this key... thanks for pointing this out :smile_cat:

zulip_bots/zulip_bots/bots/giphy/giphy.py
shubhamdhama: Okay, I'll clean this function now... this was the result of "hack mode" advise... though it was super hacky I think :sweat_smile:

showell:
Closing because we have abandoned the zgram effort.  Thanks for working on this @shubhamdhama!  I think we may eventually try to revive this concept, but for now we just don't have the interest/bandwidth.

PR #468, opened 2018-08-11
api: Implement GET /server_settings.

YagoGG:
**Testing plan:** lint and mypy checks passed locally. Tested the example in Python's REPL interpreter.


timabbott:
@amanagr BTW, we've just added a lot of endpoints to the bindings; might be worth looking at whether this lets you remove some code from the terminal app.

PR #474, opened 2018-08-18
api: Implement GET /user_groups.

zy964c:
Added binding for `/user_groups` endpoint

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -1208,6 +1208,18 @@ class Client(object):
             method='GET'
         )
 
+    def get_user_groups(self):
+        # type: () -> Dict[str, Any]
+        '''
+            Example usage:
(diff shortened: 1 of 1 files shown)


zulip/zulip/__init__.py
```diff
@@ -1208,6 +1208,19 @@ def get_stream_topics(self, stream_id):
             method='GET'
         )
 
+    def get_user_groups(self, request=None):
```
YagoGG: What is `request` for? The original endpoint doesn't accept any parameters, right?

zy964c: @YagoGG ,you are right, removed `request`.

eeshangarg:
@zy964c: Thanks for working on this! Merged! :)

@YagoGG: Thanks for reviewing! :)

PR #475, opened 2018-08-24
api: Rename "get_presence" to "get_user_presence".

YagoGG:
Given that this method fetches the presence status for a single user, the "get_presence" name should be reserved for the endpoint that gets the presence for all users.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -989,12 +989,12 @@ class Client(object):
             request=request,
         )
 
-    def get_presence(self, email):
+    def get_user_presence(self, email):
         # type: (Dict[str, Any]) -> Dict[str, Any]
         '''
             Example usage:
 
-            >>> client.get_presence()
+            >>> client.get_user_presence('iago@zulip.com')
             {'presence': {'website': {'timestamp': 1486799122, 'status': 'active'}}, 'result': 'success', 'msg': ''}
         '''
         return self.call_endpoint(
--- a/zulip/zulip/examples/get-user-presence
+++ b/zulip/zulip/examples/get-user-presence
@@ -24,7 +24,7 @@
 from __future__ import print_function
 import argparse
 
-usage = """get-presence --email=<email address> [options]
+usage = """get-user-presence --email=<email address> [options]
 
 Get presence data for another user.
 """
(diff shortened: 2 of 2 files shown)


eeshangarg:
@timabbott: This looks good to me, but should we be renaming an existing endpoint? Is it okay to expect users to migrate their old code to the new name after they upgrade to a new release of the bindings?

@YagoGG: Thanks for working on this! Waiting for a thumbs-up from Tim before I go ahead and merge this one! :)

YagoGG:
@eeshangarg No problem, sounds like something reasonable to take into account.

My two cents: maybe we can make `get_presence` a wrapper for `get_user_presence` that displays a deprecation notice, and then get rid of `get_presence` after some transition time.

zulip/zulip/__init__.py
```diff
@@ -989,12 +989,12 @@ def get_profile(self, request=None):
...
         )
 
-    def get_presence(self, email):
+    def get_user_presence(self, email):
         # type: (Dict[str, Any]) -> Dict[str, Any]
         '''
             Example usage:
 
-            >>> client.get_presence()
+            >>> client.get_user_presence()
```
timabbott: doesn't this need an argument in the example?

timabbott:
@eeshangarg it's fine to rename this; it's a function in the API bindings we just added recently and that has never been documented.

timabbott:
lgtm aside from the example comment above.

YagoGG:
@timabbott Just for the record, this function isn't one from my recent additions. It has been in our bindings since February 2017 (see a87ae4b1c4401cea45e0d2729f6208271a98f708), hence Eeshan's concerns.

zulip/zulip/__init__.py
YagoGG: Added the argument, and also renamed the example file.

timabbott:
Ahh, OK, I'd misread the commit history.  I'd be surprised if there are many sites using it, so we should just be sure to include mention of the rename in the changelog for this project.

timabbott:
@eeshangarg can you make sure this gets mentioned in the project's changelog?

eeshangarg:
@timabbott: Done! :)

Issue #445, opened 2018-06-27
Renaming bots can break them.


showell:
I recently renamed one of my bots from "Steve's Trivia Bot" to "Steve's Giphy Bot", and the bot program stopped working (it was running giphy, I just hadn't bothered to rename it till later).  Once I re-started the program, it worked again.

I suspect this is a problem that affects many bots, and I just happened to see it with giphy.

I didn't investigate, but I think there's something going on on the bot side where the bot library is still trying to strip off the old name of the bot in the mentions or is looking for the mention.  If so, we need to either a) use a regex to strip off mentions (and assume the mention is for us) or b) register for name change events so we can update our data structures.  The (b) option seems like slight overkill, but name changes are rare, so it's maybe the proper solution and probably not that hard.

As I said, I haven't investigated this, so a lot of the effort here will be just running a bot, renaming the bot user, and getting to the bottom of what's going on.  If you have multiple ways to solve it, you can discuss on the integrations stream.

shubhamdhama:
```diff
diff --git a/zulip_bots/zulip_bots/lib.py b/zulip_bots/zulip_bots/lib.py
index 2281d66..4ca9bae 100644
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -1,6 +1,7 @@
 import configparser
 import json
 import logging
+import re
 import os
 import signal
 import sys
@@ -229,7 +230,8 @@ def extract_query_without_mention(message: Dict[str, Any], client: ExternalBotHa
     If the bot is the first @mention in the message, then this function returns
     the stripped message with the bot's @mention removed.  Otherwise, it returns None.
     """
-    mention = '@**' + client.full_name + '**'
+    find_mentions = r'(?<![^\s\'\"\(,:<])@(\*\*[^\*]+\*\*)'
+    mention = re.search(find_mentions, message['content']).group()
     if not message['content'].startswith(mention):
         return None
     return message['content'][len(mention):].lstrip()

```
@showell I've tried the option `a)` and it's working properly, is there any possible case where the first mention don't belong to the bot.
Here is the image of this working:
![working](https://user-images.githubusercontent.com/22238472/42018981-618bf15e-7ad1-11e8-9446-b1d0f4e9179d.png)

showell:
It's plausible that somebody could mention two different users in a message to the bot, but I don't think it's actually going to happen in practice.

I'm not sure the regex needs to be quite as complicated here.

[...]

PR #480, opened 2018-10-02
lib: Add support for extended mention syntax.

aero31aero:
Fixes zulip/zulip#10572

The issue never was in the server repository, it was merely the bot lib rejecting the messages of the new format.

Works towards fixing #445.

### DIFF
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -5,6 +5,7 @@ import os
 import signal
 import sys
 import time
+import re
 
 
 from typing import Any, Optional, List, Dict, IO, Text
@@ -238,10 +239,18 @@ def extract_query_without_mention(message: Dict[str, Any], client: ExternalBotHa
     If the bot is the first @mention in the message, then this function returns
     the stripped message with the bot's @mention removed.  Otherwise, it returns None.
     """
+    content = message['content']
     mention = '@**' + client.full_name + '**'
-    if not message['content'].startswith(mention):
-        return None
-    return message['content'][len(mention):].lstrip()
+    extended_mention_regex = re.compile(r'^@\*\*.*\|' + str(client.user_id) + r'\*\*')
+    extended_mention_match = extended_mention_regex.match(content)
+
+    if extended_mention_match:
+        return content[extended_mention_match.end():].lstrip()
+
+    if content.startswith(mention):
+        return content[len(mention):].lstrip()
+
+    return None
 
 
 def is_private_message_from_another_user(message_dict: Dict[str, Any], current_user_id: int) -> bool:
--- a/zulip_bots/zulip_bots/tests/test_lib.py
+++ b/zulip_bots/zulip_bots/tests/test_lib.py
@@ -4,6 +4,7 @@ from zulip_bots.lib import (
     ExternalBotHandler,
     StateHandler,
     run_message_handler_for_bot,
+    extract_query_without_mention
 )
 
 import io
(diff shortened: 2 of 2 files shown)


zulip_bots/zulip_bots/lib.py
```diff
@@ -239,9 +239,12 @@ def extract_query_without_mention(message: Dict[str, Any], client: ExternalBotHa
...
     mention = '@**' + client.full_name + '**'
-    if not message['content'].startswith(mention):
-        return None
-    return message['content'][len(mention):].lstrip()
+    if message['content'].startswith(mention):
+        return message['content'][len(mention):].lstrip()
+    extended_mention = '@**' + client.full_name + '|' + str(client.user_id) + '**'
+    if message['content'].startswith(extended_mention):
+        return message['content'][len(extended_mention):].lstrip()
+    return None
```
timabbott: I think it'd be better to do an `if/elif` check, starting with checking for `extended_mention`, so that it's clearer that this won't remove 2 mentions from a message (logically).

I think there might also be a copy of similar logic in the `outgoing_webhook` part of the server repo.

aero31aero: I figured using an early return should make that clear, but I'll make that change.

zulip_bots/zulip_bots/lib.py
```diff
@@ -239,9 +240,13 @@ def extract_query_without_mention(message: Dict[str, Any], client: ExternalBotHa
     the stripped message with the bot's @mention removed.  Otherwise, it returns None.
     """
     mention = '@**' + client.full_name + '**'
-    if not message['content'].startswith(mention):
-        return None
-    return message['content'][len(mention):].lstrip()
+    extended_mention_regex = re.compile('^@\\*\\*.*\\|' + str(client.user_id) + '\\*\\*')
```
timabbott: I think this might be better spelled using `r'` (aka raw strings); you need less backslash-escaping that way.

timabbott:
Agreed with @showell that this should not depend on the name, which it looks like you've done.

aero31aero:
Pushed with changes.

zulip_bots/zulip_bots/lib.py
```diff
@@ -239,9 +240,14 @@ def extract_query_without_mention(message: Dict[str, Any], client: ExternalBotHa
     the stripped message with the bot's @mention removed.  Otherwise, it returns None.
     """
     mention = '@**' + client.full_name + '**'
-    if not message['content'].startswith(mention):
-        return None
-    return message['content'][len(mention):].lstrip()
+    extended_mention_regex = re.compile(r'^@\*\*.*\|' + str(client.user_id) + r'\*\*')
+    print(extended_mention_regex)
```
showell: There's a stray `print` here.  Also, I think it's worth saying `content = message['content']` so you don't have to repeat `message['content']` in 4 places.

zulip_bots/zulip_bots/lib.py
```diff
@@ -239,9 +240,14 @@ def extract_query_without_mention(message: Dict[str, Any], client: ExternalBotHa
...
+    extended_mention_match = extended_mention_regex.match(message['content'])
+    if extended_mention_match:
+        return message['content'][extended_mention_match.end():].lstrip()
+    elif message['content'].startswith(mention):
```
showell: Stylistically, if an `if` block ends in a `return`, I usually put a blank line after it, and then do an `if`, not an `elif`.  It's not a big deal, I just find it a bit easier to read:

~~~ py
if case_one:
    # bla bla bla
    return foo

if case_two:
    return bar

# We got this far, so we must not have matched on anything
return None
~~~

aero31aero: Formatted it accordingly now. I think I initially changed it from that because Tim suggested that he preferred that design pattern. I agree that the early return pattern looks neater, though.

timabbott: I think it previously didn't obviously return in the first code path?  Anyway, looks good now.

timabbott:
Merged, thanks @aero31aero!

PR #484, opened 2018-10-28
provision: Refactor version tuple generation logic.

### LINKED ISSUE #482: tools/provision error with python3.6.7rc1
Script is unable to convert this version to int.

```
  File "tools/provision", line 119, in <module>
    main()
  File "tools/provision", line 39, in main
    py_version = tuple(int(num) for num in py_version_list)
  File "tools/provision", line 39, in <genexpr>
    py_version = tuple(int(num) for num in py_version_list)
ValueError: invalid literal for int() with base 10: '7rc1'
```

ujjwalrox:
Previous tuple generation logic was throwing an error parsing version like 3.6.7rc1
because micro-versions like '7rc1' cannot be converted into integer.

Fix #482

### DIFF
--- a/tools/provision
+++ b/tools/provision
@@ -36,7 +36,7 @@ the Python version this command is executed with."""
                                                 stderr=subprocess.STDOUT, universal_newlines=True)
     # The output has the format "Python 1.2.3"
     py_version_list = py_version_output.split()[1].split('.')
-    py_version = tuple(int(num) for num in py_version_list)
+    py_version = tuple(int(num) for num in py_version_list[0:2])
     venv_name = 'zulip-api-py{}-venv'.format(py_version[0])
 
     if py_version <= (3, 1) and (not options.force):


ujjwalrox:
@timabbott   @rishig , I would really appreciate it if you could please review this one.

tools/provision
```diff
@@ -32,17 +32,14 @@ the Python version this command is executed with."""
     options = parser.parse_args()
 
     base_dir = os.path.abspath(os.path.join(__file__, '..', '..'))
-    py_version_output = subprocess.check_output([options.python_interpreter, '--version'],
-                                                stderr=subprocess.STDOUT, universal_newlines=True)
-    # The output has the format "Python 1.2.3"
-    py_version_list = py_version_output.split()[1].split('.')
-    py_version = tuple(int(num) for num in py_version_list)
+
+    py_version = tuple(sys.version_info)  # (3, 6, 7, 'candidate', 1) for Python 3.6.7rc1
```
timabbott: I don't think this change is correct, since you're now ignoring the `-p` option.  We want 
`./tools/provision -p /usr/bin/python2.7` to throw this error.  

Instead, we should just fix the parsing logic to handle `rc1` type values.

timabbott: (I think a reasonable option would be to just put `[0:2]` at the end of the old `py_version_list` definition, since we don't need the minor version at all in the following code).

timabbott:
Nice, merged, thanks @ujjwalrox!

Issue #486, opened 2018-11-28
twitter integration: Add support for filtering messages matching a given account/term


timabbott:
Our deployment of the Twitter integration (`zulip/integrations/twitter/twitter-bot`; docs are on zulipchat.com/integrations) has started showing all the tweets from a user who picked the username `@zulip_engg` (and there's a few other things we'd want to exclude too).  It seems worth adding options for it to exclude messages from certain accounts and/or matching certain terms (probably one option for each).

Hyaxia:
@zulipbot claim

zulipbot:
Welcome to Zulip, @Hyaxia! We just sent you an invite to collaborate on this repository at https://github.com/zulip/python-zulip-api/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

Hyaxia:
Hey, was wondering about the "certain terms" you thought about.
Did you mean that like the `--search` that receives a list of arguments to check, whether

timabbott:
Yeah, that's what I had in mind.

zulipbot:
Hello @Hyaxia, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

Hyaxia:
waiting for review :relaxed:

timabbott:
Fixed via #487.  @Hyaxia next time, you can put `Fixes: #487` in the commit message and it'll auto-close the issue when the commit is merged.  

[...]

PR #487, opened 2018-11-29
Added --excluded-users and --excluded-terms options

Hyaxia:
Relevant issue - https://github.com/zulip/python-zulip-api/issues/486

Any thing that will be entered in the --exclude string separated by
a ',' will then be checked if it appears in the tweet.
If it does then the tweet wont be sent.

### DIFF
--- a/zulip/integrations/twitter/twitter-bot
+++ b/zulip/integrations/twitter/twitter-bot
@@ -51,7 +51,13 @@ twitter-bot --config-file=<path/to/.zuliprc> --search="<search-query>"
 or
 twitter-bot --config-file=<path/to/.zuliprc> --twitter-name="<your-twitter-handle>"
 
-4. Configure a crontab entry for this script. A sample crontab entry
+- optional - Exclude any terms or users by using the flags `--exluded-terms` or `--excluded-users`:
+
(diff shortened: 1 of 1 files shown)


zulip/integrations/twitter/twitter-bot
```diff
@@ -198,6 +206,16 @@ for status in statuses[::-1][:opts.limit_tweets]:
     text_to_check = [status.text, status.user.screen_name]
     text_to_check.extend(url.expanded_url for url in status.urls)
 
+    # Check that the tweet doesn't contain any terms that
+    # are supposed to be excluded
+    exclude = False
+    for term in excluded_terms:
+        if any(term.lower() in text.lower() for text in text_to_check):
```
showell: A minor optimization here would be to lowercase text_to_check strings outside the loop.

Hyaxia: Will be done, thanks for reviewing

timabbott:
(I did some basic testing on chat.zulip.org, and this does work for filtering worlds from the content of messages; but the main use case we have in mind is filtering messages sent by certain accounts, so we should add that.)

timabbott:
Sorry for the slow review!  This looks great, merged.  Thanks for fixing!

PR #492, opened 2018-12-10
bots: Fix a Chessbot dependency error.

skunkmb:
This PR specifies a version for Chessbot so that it will continue to function properly if the library is updated. This also updates `test-bots` to remove a temporary fix for Chessbot.


showell:
Merged, thanks @skunkmb!

PR #494, opened 2018-12-17
api: Implement endpoint to upload custom emoji.

rogersouza:
**Testing plan**: lint and mypy checks passed locally.

This PR is based on the PRs made by @YagoGG, so please let me know if there's is anything left to do here.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -864,6 +864,20 @@ class Client(object):
             method='GET'
         )
 
+    def upload_custom_emoji(self, emoji_name, file_obj):
+        # type: (str, IO[Any]) -> Dict[str, Any]
+        '''
+            Example usage:
+
+            >>> client.upload_custom_emoji(emoji_name, file_obj)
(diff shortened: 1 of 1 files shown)


zulip/zulip/__init__.py
```diff
@@ -864,6 +864,20 @@ def get_realm_emoji(self):
             method='GET'
         )
 
+    def upload_custom_emoji(self, emoji_name, file):
+        # type: (str, IO[Any]) -> Dict[str, Any]
+        '''
+            Example usage:
+
+            >>> client.upload_custom_emoji(emoji_name, file)
```
eeshangarg: I think the example call to the function here could be more explicit. We should just make it clear that `file` is a file object and not just a path.

Also, it is probably better to use another name such as `file_obj`, the name `file` is a reserved keyword in Python! :)

rogersouza: I changed it, as you suggest :)

eeshangarg:
@rogersouza: This looks good to me now! Thanks for working on this! Merged! :)

PR #495, opened 2019-01-14
API: Add configurable request timeout internally & to deregister

neiljp:
This is motivated by a desire to configure the timeout of the request, which is currently set to be ~15s. If the client is initially connected, loses connection and then makes a call (such as to `deregister`), then the API call blocks for this time - which is perhaps more noticeable with `deregister` since it may be called at the end of an application.

The PR is split into one commit to pass timeout parameters inside the regular API, and another to allow deregister to explicitly set a timeout. The latter could be excluded, but I think it could be a common enough use case to include, for the above reason.

I did consider making the timeout-setting line be something like the following, but I wasn't sure what a good value for `SOME_MINIMUM` might be.
```python
   488             request_timeout = 15 if not timeout else max(timeout, SOME_MINIMUM)
```

The timeouts are currently set as `int`, but perhaps these should be `float`?


neiljp:
@timabbott Updated to be `float`s.

PR #503, opened 2019-01-30
bots: Use bold formatting judiciously in trivia bot.

aero31aero:


showell:
Merged, @aero31aero! Thanks for the quick follow up.

PR #504, opened 2019-02-08
interrealm: Fix the formatting of messages with special prefixes.

rht:


zulip/integrations/bridge_between_zulips/run-interrealm-bridge
```diff
@@ -34,6 +34,12 @@ def create_pipe_event(to_client: zulip.Client, from_bot: Dict[str, Any],
                 # comes from
                 msg["content"] = msg["content"].replace("/user_uploads/",
                                                         from_bot["site"] + "/user_uploads/")
+            for prefix in ["```quote", "- ", "* ", "> ", "1. "]:
+                if msg["content"].startswith(prefix):
```
punchagan: Instead of iterating over the list of prefixes, you could just [pass a prefix](https://docs.python.org/3/library/stdtypes.html?highlight=string%20startswith#str.startswith) tuple to `startswith`.

rht: That's neat, TIL.

rht: It looks like https://github.com/zulip/zulip/blob/e4832cc9da2ea38825817a7ed1ec9305abc3ca22/tools/linter_lib/custom_check.py#L59-L62, https://github.com/zulip/zulip/blob/e4832cc9da2ea38825817a7ed1ec9305abc3ca22/zerver/lib/storage.py#L70 can be simplified this way as well.

timabbott:
Merged as 6368b63e6cecfee8e5e72718780e023342a892ed, after changing the rule to handle code blocks as well.

I'm sure this will not be the final version of that code, but it'll do for now.

PR #505, opened 2019-02-08
bridge_with_irc: Check if bot is subscribed to stream at startup.

### LINKED ISSUE #500: bridge_with_irc: Check if the bot is subscribed to the relevant stream on startup.
This way, the important precaution to make sure that the bot is subscribed is no longer necessary, since the bot will print out the problem itself.

rht:
This fixes #500.
- [x] Test this

### DIFF
--- a/zulip/integrations/bridge_with_irc/irc_mirror_backend.py
+++ b/zulip/integrations/bridge_with_irc/irc_mirror_backend.py
@@ -19,6 +19,8 @@ class IRCBot(irc.bot.SingleServerIRCBot):
         self.topic = topic
         self.IRC_DOMAIN = server
         self.nickserv_password = nickserv_password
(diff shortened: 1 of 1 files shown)


orientor:
Workiing properly. Tested.
Screenshot from 2020-03-01 13-31-15

rht:
@orientor appreciate the test!

showell:
@rht Thanks for doing this!  I am sure this will save some troubleshooting headaches.

If either you or @orientor has time, we could probably make this a bit more friendly by reporting all the missing streams by name, and maybe just exit with a non-zero return code rather than having the traceback.

Merging.

rht:
I will let @orientor do the reporting of the stream and the `exit(1)`.

PR #506, opened 2019-02-16
API: Fix add_reaction method to pass reaction_data.

neiljp:
Context: I'm in the process of migrating ZT to use 'full' API calls, rather than `do_api_query` or `call_endpoint`, which were necessary to call endpoints directly when the python API didn't have the functionality.

I wondered why the `add_reaction` method wasn't working; apparently the method doesn't pass the actual data to `call_endpoint`, so it's not really so surprising!

I also corrected the example usage text, which had an extra `emoji` in the method name.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -824,7 +824,7 @@ class Client(object):
         '''
             Example usage:
 
-            >>> client.add_emoji_reaction({
+            >>> client.add_reaction({
                 'message_id': '100',
                 'emoji_name': 'joy',
                 'emoji_code': '1f602',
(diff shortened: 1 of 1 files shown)


timabbott:
@eeshangarg should this have been caught by our API tests?

eeshangarg:
@timabbott: Yes, it should have been! But as it turns out, we don't have tests on the server side for this function in `zerver/lib/api_test_helpers.py`. I'm on it, will get this done asap and make a release tonight! :) Thanks for letting me know!

neiljp:
Thanks for the merge; I look forward to the release to finish my ZT migration :+1:

It's good to know about the test helpers; I've not dug super-deep into them, but some of it feels a little like  using pytest! I was surprised this had slipped through too - should the doc difference have been caught too?

PR #311, opened 2018-01-27
Update __init__.py

Shivamgera:
I did [Issue#164](https://github.com/zulip/python-zulip-api/issues/164). Removed RuntimeError and created a custom error called ZulipError. Also, used some more descriptive errors.

PR #165, opened 2017-11-21
Fix zuliprc handling in zulip-run-bot.

showell:

PR #508, opened 2019-03-24
api: Add NetworkError Exception & use twice in do_api_query.

neiljp:
This replaces one case where the base ZulipError was raised, and another
where the original exception was re-raised.

This method, and so all API calls, should now only fail via raising this
Exception or by a non-'success' result in the return value dict; this
should avoid leaking any implementation detail and clarify expectations
in user code.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -294,6 +294,9 @@ class ConfigNotFoundError(ZulipError):
 class MissingURLError(ZulipError):
     pass
 
+class UnrecoverableNetworkError(ZulipError):
+    pass
+
 class Client(object):
     def __init__(self, email=None, api_key=None, config_file=None,
                  verbose=False, retry_on_errors=True,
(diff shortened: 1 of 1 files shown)


neiljp:
Tagging @showell for initial review since `ZulipError` was added in #165 (and extended in #311).

timabbott:
I think we should call this `IrrecoverableNetworkError` if we're going to use it -- the point is that these are issues like invalid SSL certificate that retrying within the next 30 seconds isn't going to fix, and thus these escape from our automatic retry logic for transient errors.

neiljp:
@timabbott Renamed `NetworkException` to `UnrecoverableNetworkException`, which seemed like the word I would personally expect (I also did a quick check for which term was more popular based on https://trends.google.com/trends/explore?date=all&q=unrecoverable,irrecoverable)

timabbott:
OK, makes sense as a way to pick which of the two to use to me.  Thanks @neiljp!

Issue #433, opened 2018-06-10
Rework and enable tests for Merels.


amanagr:
Merels was ported to `game_handler` recently, but it's tests were not revised. They need to be reworked. Tests for `connect_four` bot can serve as good examples for how to modify the tests for Merels.

[merels.py](https://github.com/zulip/python-zulip-api/blob/master/zulip_bots/zulip_bots/bots/merels/merels.py) was totally changed so write new set of tests for it. Files in [libraries](https://github.com/zulip/python-zulip-api/tree/master/zulip_bots/zulip_bots/bots/merels/libraries) were not much changed so they need to be tweaked to accomodate for the [changes](https://github.com/zulip/python-zulip-api/commit/243f7bbe5ddc76ed8f69573f6b6c1b6022880932).

Enable testing for merels by adding `__init__.py` file to the directory.

amanagr:
@zulipbot label "help wanted"

amanagr:
@zulipbot label "good first issue"

neiljp:
See the following for a good first commit to work from: https://github.com/zulip/python-zulip-api/pull/403/commits/bd945c3d5b7558864ecd2ab681a404cbaa63cb07

JordanTroutman:
@zulipbot claim

zulipbot:
Welcome to Zulip, @JordanTroutman! We just sent you an invite to collaborate on this repository at https://github.com/zulip/python-zulip-api/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

JordanTroutman:
@zulipbot claim

zulipbot:
Hello @JordanTroutman, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

[...]

PR #511, opened 2019-04-22
bot tests: Created tests for Merels

NealMatta:
This PR improves test coverage from 0% to 98%. 

This PR directly tackles issue #433

### DIFF
--- a/zulip_bots/zulip_bots/bots/merels/test/test_constants.py
+++ b/zulip_bots/zulip_bots/bots/merels/test/test_constants.py
@@ -1,6 +1,6 @@
 import unittest
 
-from libraries import constants
+from .. libraries import constants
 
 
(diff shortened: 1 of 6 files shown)


NealMatta:
Just in case anyone want's to continue to work on this error, the tests which I wrote work locally and provide a line coverage of 98%. Travis CI is failing on merels/libraries/game_data.py and merels/libraries/mechanics.py because these files have circular imports with one another. I tried doing a bit of research and implementing changes to no avail.

eeshangarg:
@NealMatta: I've run into this before, here's a quick suggested fix (which should work):

I would take a look at the function or block that is causing the error and move the offending import statement there. Usually, circular import errors can be fixed that way, but it is only a quick fix. A better fix would be to remove the circular dependency altogether! :) Thanks!

Issue #409, opened 2018-05-30
Remove or streamline the converter bot.


showell:
@roberthoenig We may want to remove the converter bot, as per our chat conversation.  Or maybe just streamline a bit, so it's optimized for somebody just wanting to learn how to write a really basic bot.

PR #512, opened 2019-04-22
bots: Refactor and add conversions to converter bot

jchinnam:
This commit refactors unit handling into a class and encapsulate similar portions into functions. Also, added conversions for time, pressure, and angles.

This commit is related to #409


zulip_bots/zulip_bots/bots/converter/converter.py
```diff
@@ -50,6 +53,61 @@ def handle_message(self, message: Dict[str, str], bot_handler: Any) -> None:
...
+
+    @property
+    def fr(self) -> str:
+        '''Getter for the input unit'''
+        return self.unit_from
+
+    @property
+    def to(self) -> str:
+        '''Getter for the output unit'''
+        return self.unit_to
```
eeshangarg: We don't really need these getters here, it would just be simpler to access the underlying attributes `unit_from` and `unit_to` directly.

PR #514, opened 2019-05-22
Add Confluence Integration.

mayankmadan:
Fixes: 5142

Made some changes to the original PR by @paras112
Also made the script executable since its a cronjob.

@eeshangarg Can you have a look at it?


zulip/integrations/confluence/confluence
```diff
@@ -0,0 +1,115 @@
...
+To use this script:
+1. Set up Confluence authentication, as described below
+2. Set up a Zulip bot user and download its `.zuliprc`
+   config file to e.g. `~/.zuliprc`
+3. Subscribe the bot to the stream that will receive Confluence updates
+4. Configure a crontab entry for this script. A sample crontab entry
+that will process spaces and pages every 5 minutes is:
+*/5 * * * * /usr/local/share/zulip/integrations/confluence/confluence [options]
+
+== Setting up Confluence authentications ==
```
eeshangarg: Replace `authentications` with `authentication`. :)

zulip/integrations/confluence/confluence
```diff
@@ -0,0 +1,115 @@
...
+server_base_url =
+
+To obtain an API token from your Atlassian account:
+
+1. Log in to https://id.atlassian.com.
+2. Click API token.
+3. From the dialog that appears, enter a memorable and concise Label for your token and click Create.
+4. Click Copy to clipboard, then paste the token in .zulip_confluencerc:
+
+"""
```
eeshangarg: Anything that needs to be clicked or is a UI element should be bolded. :)

zulip/integrations/confluence/confluence
```diff
@@ -0,0 +1,115 @@
...
+
+Run this on a personal or trusted machine, because your API key is
+visible to local users through the command line or config file.
+
+Please create a ~/.zulip_confluencerc with the following contents:
+
+[Confluence]
+username =
+token =
+server_base_url =
```
eeshangarg: Could you please expand on what values are to be assigned to these parameters a bit?

zulip/integrations/confluence/confluence
```diff
@@ -0,0 +1,115 @@
...
+streams = [i['name'] for i in streams]
+
+url = '{}/rest/api/content/search?cql=created>=now("-{}m")'.format(server_base_url, opts.frequency)
+auth = (username, token)
+response = requests.get(url, auth=auth)
+response_json = response.json()
+
+if response_json['results']:
+    for page in response_json['results']:
+        content = 'New Page Created\nTitle: {}\nURL: {}{}'.format(page['title'], server_base_url, page['_links']['webui'])
```
eeshangarg: I would make the content template:
``` python
content = 'New page [{title}]({url}) created.`.format(
    title=page['title'],
    url='{}{}'.format(server_base_url, page['_links']['webui']
)
```

zulip/integrations/confluence/confluence
eeshangarg: Actually, ignore this one, for some reason, I thought this was a Markdown file haha.

mayankmadan:
@eeshangarg I have made some changes to the content of the message that will be posted when a new page is created. I wanted to get your opinion on it. I was thinking of further formatting 'created_at'. But before doing that I wanted to ask you if the extra information is worth the extra request sent

timabbott:
Can you provide more details on the question about `created_at`, maybe with an example + screenshot?  Not sure what the product question is.

mayankmadan:
> Can you provide more details on the question about `created_at`, maybe with an example + screenshot? Not sure what the product question is.

@timabbott I'm sorry about that question. At first I somehow missed the 'expand' GET parameter mentioned in the Confluence API's documentation and thought I needed to send a separate request to another endpoint to fetch the author name and 'created_at'. I changed that in the latest commit.

eeshangarg:
@mayankmadan: This looks good to me. I would recommend addressing my latest comments and once we get a thumbs up from @timabbott, I think we can merge this.

But before we merge it, could you please open a PR that documents this integration in the main repo? That would be great! Thanks! :)

zulip/integrations/confluence/confluence
```diff
@@ -0,0 +1,122 @@
...
+url = '{}/rest/api/content/search?cql=created>=now("-{}m")&expand=history'.format(server_base_url, opts.frequency)
+auth = (username, token)
+response = requests.get(url, auth=auth)
+response_json = response.json()
+
+if response_json['results']:
+    for page in response_json['results']:
+        created_by = page['history']['createdBy']['displayName']
+        created_at = page['history']['createdDate'].split('T')
+        content = 'New page [{title}]({url}) created by {created_by} at {time}.'.format(title=page['title'],
```
timabbott: I would line-wrap this after `.format(` for better readability.

mayankmadan:
@eeshangarg @timabbott, Right now, the bot needs to be a generic bot as it needs to get the list of streams to which the bot is subscribed to so the updates are sent on those streams. But since the permissions system for bots isnt great right now (on which I really want to work on) I was wondering if this could be made into an incoming webhook bot and the list of streams can be provided in config file

timabbott:
@mayankmadan can you explain precisely why it needs that list? I would imagine there would need to be configuration somewhere determining where in Zulip to send messages from Confluence, but I would imagine we'd just want that to be e.g. a small function in the configuration file that the user writes, not something that involves pulling data from the Zulip server.  (A list of subscribed streams, which is what one would get from Zulip, wouldn't encode the mapping the user wants of (Confluence article -> Zulip stream+topic).

mayankmadan:
Right now, for the bot to function, the user must subscribe it to a stream so the bot knows in which stream to post the updates. The topic inside the stream is just the title of the new page created. Thats why it needs the list of subscribed streams. Also, the stream wouldn't change depending on the Confluence article; only the topic would. So I was thinking to maybe change that so it doesn't have to pull the list of subscribed stream from Zulip API and just provide it the list through the config file.

PIG208:
I think the whole concept of this bot makes it more suitable to be an incoming webhook that can be integrated into the server. 
It appears that confluence has added support to it in https://jira.atlassian.com/browse/CONFSERVER-52487. But I can't find any relevant documentation for it.

It might be unnecessary to stick with the "subscribe and send messages to subscribed streams" model. I think it should be totally sufficient to have the user hardcode the desired stream to receive notifications for in the config file. We should be able to merge this after tweaking the configuration a little bit and make sure it works with the current version of their REST API.

PR #517, opened 2019-07-27
Bots: Add support for running packaged bots via entrypoints

neiljp:
The entrypoints approach seems like it may make it easier to produce third-party bots, and could potentially be extended to be the default way in which `zulip_bots` accesses built-in bots?

A test is included for the new approach, as well as:
- removal of python 3.4 support and addition of 3.7 to travis (branch failed due to an existing bot requiring lxml with 2.7 or >=3.5)
- addition of `ImportError` as an alternative to `ModuleNotFoundError`, which is only introduced in python 3.6
- changes to improve indication to the user, where the bot module is being sourced

### DIFF
--- a/zulip_bots/setup.py
+++ b/zulip_bots/setup.py
@@ -55,6 +55,7 @@ setuptools_info = dict(
         'html2text',
         'lxml',
         'BeautifulSoup4',
+        'entrypoints',
     ],
 )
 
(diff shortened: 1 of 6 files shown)


neiljp:
If this is a reasonable approach then we can add support for a 'sample' package, which I've done locally but not included so far.

timabbott:
I merged the first 2 commits.

neiljp:
Just adjusted the last commit to take into account 'editable' bot packages, which don't have a version - at least according to entrypoints - so just show the bot module_name details.

timabbott:
Closing, since we've now merged #708.

PR #518, opened 2019-07-30
setup.py: Add supported python versions for all packages.

neiljp:
This PR adds
* trove classifiers to the packages
* `python_requires` limits, which was introduced in July 2016 (?)

The versions for `zulip_bots` and `zulip_botserver` were discussed previously; I'm not sure there is a defined range of expected supported versions for the API package, but I've assumed >=2.7 or >=3.5 for now, as per the classifiers and `python_requires`.

For testing the `python_requires` I've not exhaustively used `pyenv` or similar locally. However, I added a **[DO NO MERGE]** commit; travis doesn't (easily) support 2.6 (or 3.0/1/2/3) right now, but this extra commit adds 3.4 support back in for test-main and extra for the api-only tests, and it fails promptly as `setup.py` would not install `zulip_bots` or `zulip` - though does install the latter on 3.5. That suggests that `python_requires` is working as expected.


timabbott:
This looks good, except that Travis CI seems to not have run?

neiljp:
I just dropped off the 'DO NO MERGE' commit, which should be the cause of the not-running/passing CI.

neiljp:
@timabbott This passes CI now that I removed the intentionally-failing test commit.

PR #520, opened 2019-08-07
Client.get_user_presence: Fix blatantly incorrect type annotation.

andersk:


timabbott:
@andersk do we need to do a release of this for https://github.com/zulip/zulip/pull/12948 to proceed?

timabbott:
(Also, do we want to add a `py.typed` file?)

andersk:
Doing a release isn’t a blocker for zulip/zulip, which is already using this from Git. The `zulip` package already has a `py.typed`. Maybe we want to add it to `zulip_bots` and `zulip_botserver`, though I’m guessing that will uncover some more errors in zulip/zulip—I’ll do some tests.

PR #521, opened 2019-08-08
zulip_bots, zulip_botserver: Add PEP 561 type annotation markers

andersk:


timabbott:
Given that we can't usefully use `typing` at runtime on Python 2.7 anyway, I guess this is a reasonable solution.

andersk:
Oh this is still supposed to support Python 2.7? We do have `typing` imports at top level, so we should put back `'typing;python_version<"3.5"'` in that case.

PR #522, opened 2019-08-10
Restore dependency on typing package for Python < 3.5

andersk:
Apparently we still support Python 2.7 for some reason.

Follow-up to #521.


neiljp:
I've tried to clarify this in #518 based on my contributions this repo.

PR #526, opened 2019-11-01
zulip: Use distro.linux_distribution instead of platform

punchagan:
Python 3.8 removes `platform.linux_distribution` and `distro` is the
recommended replacement to use for this function.


timabbott:
Looks great, merged, thanks @punchagan!

PR #529, opened 2019-11-06
mypy: Upgrade to 0.770

neiljp:
I was trying to work on another issue and realized the mypy version was quite old, hence this PR.

Most changes are fairly straightforward, but perhaps due to some improvements in typeshed, the input/output with `Popen` objects is more clearly typed, and it raises concerns over whether variables are `bytes` or `str`. I've incorporated some of these types of changes already where they seem fairly reasonable, but this approach is used quite a bit more in zephyr - and since I recall it was important to run zephyr code changes past @timabbott, I'd like to get Tim's input on that before completing this.

So: is `universal_newlines` applicable in general, or is it better to encode and decode via `utf-8` to allow for extended characters? An explicit encoding is used at least in `send_zephyr`, which is what made me rethink the first approach. I think an `encoding` can be specified in the `Popen` constructor on 3.6+, but that doesn't help, particularly if the zephyr mirror might still be intended to be compatible with python2?


neiljp:
The commits are shown out of order above, but I've added some commits to fix a few outstanding issues that I missed when testing locally:
* more type-ignores for the `exec_module` issue (can we re-use this code, at least in eg. botserver?)
* switch the botserver into raising exceptions rather than returning them (maybe this isn't tested yet?)

These additions leave only checks failing in the zephyr mirror backend, as per above.

timabbott:
For the Zephyr stuff, we should just be really sure that we're not accidentally breaking it.  It's annoying to test the Zephyr code not in a production setting and I don't expect to do any further development on it in the future, so I'd prefer to minimize churn to its code over other potential goals for it.

neiljp:
Re type-checking zephyr, would it therefore be best to continue to check what we can for now and add ignore/Any in minimal places? Or even perhaps just exclude zephyr from the type checking, if the code isn't expected to change?

timabbott:
I'd probably do the Any/ignore approach if it's not super time-consuming.

neiljp:
@timabbott I just pushed an updated version of this, which now upgrades to mypy 0.770 (the latest)

neiljp:
Once this is merged, then python 3.8 testing is practical (which requires a later `typed-ast` than current `mypy` depends upon).

timabbott:
This is great, merged, thanks @neiljp!

Issue #497, opened 2019-01-21
Zulip client should detect the presence of a zuliprc in $PWD.


rht:
Reading from ~/.zuliprc forces the number of account to be 1. Having to type the API key in a command line is not secure (this leaves trails in the history). Additionally, having a local and global config has been practiced by npm for a while.

If this is to be implemented, another consideration is whether to make the zuliprc visible or not.

note: `zulip-terminal` has already implemented this local config detection.

timabbott:
We already support just having users pass the path to a `.zuliprc` file, right?

rht:
Huh, yeah, I could have used that instead of passing the API key, but again, it is more convenient if the client anticipates that `.zuliprc` is in the current directory. So the remaining argument is for convenience and that npm does it.

orientor:
@zulipbot claim

zulipbot:
Welcome to Zulip, @orientor! We just sent you an invite to collaborate on this repository at https://github.com/zulip/python-zulip-api/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

orientor:
PR added. Please review. Also how should we change the docs in this case?

zulipbot:
Hello @orientor, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

gvarun1:
@zulipbot claim

zulipbot:
[...]

PR #535, opened 2020-02-21
Config file: Add $PWD as default directory

orientor:
The API will first check whether config file exists in the home directory. If not it will take the file from PWD.

This solves issue #497


zulip/zulip/__init__.py
```diff
@@ -270,6 +270,8 @@ def get_default_config_filename():
             os.path.exists(os.path.join(os.environ["HOME"], ".humbugrc"))):
         raise ZulipError("The Zulip API configuration file is now ~/.zuliprc; please run:\n\n"
                          "  mv ~/.humbugrc ~/.zuliprc\n")
+    elif not os.path.exists(config_file):
+        config_file = os.path.join(os.environ["PWD"], ".zuliprc")
```
rht: This doesn't take into account the case when `.zuliprc` is not present at `$PWD`.

orientor: That will be checked later. Even in the original code it only raised an error if zuliprc was not present AND humbgrc was present, otherwise it proceeded nornally and was checked later.

rht: OK, make sense.

rht:
There should be a documentation somewhere that needs to be updated to say that `$PWD` is detected, shouldn't it?

orientor:
@rht Updated the documentation. Kindly check.

zulip/README.md
```diff
@@ -23,9 +23,9 @@ as we make any changes to them.
 The easiest way to use these API bindings is to base your tools off
 of the example tools under zulip/examples/ in this distribution.
 
-If you place your API key in the config file `~/.zuliprc` the Python
-API bindings will automatically read it in. The format of the config
-file is as follows:
+If you place your API key in the config file `~/.zuliprc` or './.zuliprc'
```
rht: It is more explicit to say `$PWD/.zuliprc` than `./.zuliprc`.

orientor: Done.

rht:
@orientor do you know what takes precedence in npm, `$PWD` or `~`? I think it is `$PWD` but I haven't reconfirm. If that's the case, you might make it to check for $PWD first, instead of `~`.

orientor:
@rht Yes it would make sense for local to have more precedence over global so that I can have different directories for different zulip API uses. npm also does the same, checking the project config file if we are in the directory of a project.

orientor:
@rht changed the precedence accordingly.

rht:
The commit messages need to end with fullstop as standard in Zulip commits (check out the examples at https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages). Everything else LGTM.

orientor:
@rht done. Commit messages updated.

timabbott:
I'm not convinced we want to make this change; I worry a bit that it'll replace "Have to type a bit more to specify the config" with "confusing behavior where $PWD is important to whether the tool works"

PR #538, opened 2020-02-23
API: Added random exponential backoff to do_api_query

orientor:
Set delay time on failure in do_api_query according to random exponential backoff.


orientor:
Made a PR for do_api_query. But base and cap need to be decided. Also if this is done, then I can change class RandomExponentialBackoff accordingly.

zulip/zulip/__init__.py
```diff
@@ -525,7 +525,10 @@ def error_retry(error_string):
                     sys.stdout.write(".")
                 sys.stdout.flush()
             query_state["request"]["dont_block"] = json.dumps(True)
-            time.sleep(1)
+            delay_cap = 10
+            delay_base = 0.5
+            delay_time = random.random() * min(delay_cap, delay_base * (2 ** query_state["failures"]))
```
showell: @orientor Thanks for working on this!  The code looks nice and clean, but I have one concern.

The expression `delay_base * (2 ** query_state{'failures'])` will start overwhelming `delay_cap` by the fifth failure, and then we'll be computing exponential factors of 2 for no reason, and they can get big, well, exponentially.

I think the better strategy here is to tweak it so that after N failures, we just using the `delay_cap` value.  I also think there is some point after which we should simply quit, but that can be for a future PR.

showell: See https://chat.zulip.org/#narrow/stream/92-learning/topic/exponential.20backoff.20--.20with.20jitter/near/820056 for more discussion.

orientor: @showell changed the PR. Now it doesn't compute after 5th power of 2. Regarding stopping do_api_query stops after 10 failures.

orientor: @showell should I delay_cap and delay_base be added as function parameters?

zulip/zulip/__init__.py
```diff
@@ -93,9 +95,11 @@ def fail(self):
         super(RandomExponentialBackoff, self).fail()
         # Exponential growth with ratio sqrt(2); compute random delay
         # between x and 2x where x is growing exponentially
-        delay_scale = int(2 ** (self.number_of_retries / 2.0 - 1)) + 1
-        delay = delay_scale + random.randint(1, delay_scale)
-        message = "Sleeping for %ss [max %s] before retrying." % (delay, delay_scale * 2)
+        delay_base = 0.5
+        delay = random.random() * self.delay_cap
+        if math.log2(self.delay_cap) > self.number_of_retries:
```
showell: Ugh, I think my prior suggestion was a bit misleading.  I didn't want us to actually calculate the logarithm at run time; I was just explaining mathematically that after log2N delays, we're always gonna get `delay_cap` here instead of `2 ** retries`, but I didn't mean to actually code it this way.

It's my fault here for probably prematurely optimizing.  (Well, I wasn't prematurely optimizing, I was more mistakenly worried about overflow bugs.)  I think JS can handle large exponents of 2.

orientor: @showell changed the code accordingly. The optimization could be added later if needed.

timabbott:
Can you provide terminal output showing you've tested the error handling behaves correctly?

orientor:
Screenshot from 2020-02-27 07-15-53

I have tested the condition when retry is required. But in the original code retry happens only in two cases:
1. When we receive a 50x status code.
2. If we have connected to the server before but couldn't connect now.

Hence in ALL other cases (with no retry) my code works same as the original code.(Tested it and also an error gets raised in these cases so changing loop doesn't effect them)

For the first 2 cases the output will be like that shown in the image.

orientor:
I also checked if it works normally in normal conditions. 
Screenshot from 2020-03-03 02-19-31

Screenshot from 2020-03-03 02-19-50

orientor:
Some of the other errors:

Screenshot from 2020-03-03 02-20-12
Screenshot from 2020-03-03 02-22-45

orientor:
I have thoroughly checked most of the cases Code working fine in all.

timabbott:
@orientor can you rework your commits here to follow our commit style?  I think what makes sense are:
* A first commit that adjusts the behavior of `RandomExponentialBackoff` to add a `delay_cap`, with a default value of 90 (10 is too short).
* A second commit that does the `delay_base` refactor (nonfunctional).
* A third commit that changes the randomization approach in `RandomExponentialBackoff`, which we might choose to skip.
* A next commit that migrates the `error_retry` logic to use `RandomExponentialBackoff`
* Etc.

That approach will be reviewable, biseactable, and mergable.

orientor:
@timabbott 
I have reworked my changes into two commits.
In the first commit I am improving the Random Exponential Backoff class.
In the second commit I am updating the do_api_query method to use Random Exponential Backoff.

orientor:
@timabbott removed delay_base completely as it changed only one iteration(it was set to 1/2 so it would basically reduce power of 2 by 1) and I think was not required.

Reworked changes into 3 commits.
1. Added delay_cap as class variable to the CountingBackoff class, the superclass for RandomExponentialBackoff. Changed the default delay_cap to 90.

2. Changed the algorithm for Random Exponential Backoff.

3. Changed do_api_query so as to use Random Exponential backoff.

timabbott:
The first commit adds the `delay_cap` field but doesn't have it do anything.  That isn't a coherent commit -- it's just confusing were to merge just that.  

Zulip's development model is around every commit in every PR being mergable incrementally, which is important for bisecting as well as efficient code review.  

Can you fix the first commit to fully implement delay_cap?

orientor:
@timabbott Improved the commits.
The first commit now basically adds delay_cap properly, with the Random Exponential Backoff now using delay_cap.
The second commit improves Random Exponential Backoff algorithm.
The third commit adds Random Exponential Backoff to do_api_query method.

zulip/zulip/__init__.py
```diff
@@ -526,9 +523,6 @@ def error_retry(error_string):
                     sys.stdout.write(".")
                 sys.stdout.flush()
             query_state["request"]["dont_block"] = json.dumps(True)
-            time.sleep(1)
-            query_state["failures"] += 1
-            return True
```
timabbott: Why not just have this function do the `backoff.file()` and `return backoff.keep_going()` ?  It's deduplicate code.

zulip/zulip/__init__.py
```diff
@@ -611,6 +610,7 @@ def end_error_retry(succeeded):
             end_error_retry(False)
             return {'msg': "Unexpected error from the server", "result": "http-error",
                     "status_code": res.status_code}
+        return {'msg': "Unexpected error from the server", "result": "unexpected-error"}
```
timabbott: Why do we need to add this?

timabbott:
I merged the first commit and posted a couple comments.  I'll also note that this doesn't replace the `time.sleep(1)` in `call_on_each_event`, nor does it add a change to detect "invalid authentication" failures and not retry those, which I think we should do.

zulip/zulip/__init__.py
orientor: This is a really nice idea. Implemented this and one other change and now my do_api_query commit is way more readable and has minimal changes.

zulip/zulip/__init__.py
orientor: Removed after previous changes.

orientor:
@timabbott Changed the code to be more readable and have minimal changes. I was not aware of call_on_each_event. Will make the necessary commits soon. Also can you shed some light on invalid authentication? According to me the only way of having invalid authentication is wrong API key. Are there any other methods? I will make the changes accordingly.

orientor:
Added exponential backoff to call_on_each_event. But the function was originally intended to keep retrying to connect the server until successful. So an infinite loop. @timabbott @showell Is that intended or should I cut short the infinite loop and add a max number of retries?

PR #542, opened 2020-03-01
bridge_with_irc: Check if Zulip bot is subscribed to stream before starting IRC bot.

orientor:
Use list_subscriptions method of zulip client to get list of subscriptions and then check whether given stream has been subscribed to or not. Start IRC bot only if subcribed.

Solves #500 .

### DIFF
--- a/zulip/integrations/bridge_with_irc/irc-mirror.py
+++ b/zulip/integrations/bridge_with_irc/irc-mirror.py
@@ -42,6 +42,20 @@ if __name__ == "__main__":
     # Setting the client to irc_mirror is critical for this to work
     options.client = "irc_mirror"
     zulip_client = zulip.init_from_options(options)
+    # Check if bot is subscribed to the stream
+    subscribed = False
+    subscription_dict = zulip_client.list_subscriptions()
+    assert subscription_dict["result"] == "success"
+    subscription_list = subscription_dict["subscriptions"]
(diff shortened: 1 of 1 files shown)


zulip/integrations/bridge_with_irc/irc-mirror.py
```diff
@@ -42,6 +42,20 @@
...
+    subscription_list = subscription_dict["subscriptions"]
+    for subscriptions in subscription_list:
+        if subscriptions["name"] == options.stream:
+            subscribed = True
+        if subscribed:
+            break
+    if not subscribed:
+        traceback.print_exc()
+        print("ERROR: The Zulip bot is not subscribed to the stream.")
+        sys.exit(1)
```
timabbott: I think we could check this more efficiently by just parsing the result of a `/register` call, rather than doing a separate query for this.

rht:
@timabbott this PR is a duplicate of #505 which is essentially the same thing but expressed more succinctly.

> I think we could check this more efficiently by just parsing the result of a /register call, rather than doing a separate query for this.

I can't think of a way other than making a post to /register with a narrow to the said stream name. This is again, yet another separate query.

rht:
This PR can be closed since #505 has been merged.

orientor:
@rht closed.

PR #546, opened 2020-03-11
api: Refactor GET /users endpoint.

akashaviator:
This adds binding for `GET ../users/{id}` endpoint.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -1133,9 +1133,9 @@ class Client:
             request=request
         )
 
-    def get_members(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
+    def get_users(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
         '''
-            See examples/list-members for example usage.
+            See examples/list-users for example usage.
         '''
         return self.call_endpoint(
             url='users',
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -1143,6 +1143,9 @@ class Client:
             request=request,
         )
 
+    def get_members(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
+        return self.get_users(request=request)
+
     def get_alert_words(self) -> Dict[str, Any]:
         '''
             See examples/alert-words for example usage.
--- a/zulip/zulip/examples/list-users
+++ b/zulip/zulip/examples/list-users
@@ -18,5 +18,5 @@ options = parser.parse_args()
 
 client = zulip.init_from_options(options)
 
-for user in client.get_members()["members"]:
+for user in client.get_users()["members"]:
     print(user["full_name"], user["email"])


zulip/zulip/__init__.py
```diff
@@ -1076,6 +1076,21 @@ def delete_stream(self, stream_id):
             method='DELETE',
         )
 
+    def get_member(self, user_id, request=None):
+        # type: (int, Optional[Dict[str, Any]]) -> Dict[str, Any]
+
+        '''
+            Example usage:
+
+            >>> client.get_member(8,{'include_custom_profile_fields': True})
```
timabbott: I'd put a space after the `,` here, as that's standard Python style.  Otherwise lgtm.

akashaviator:
@timabbott added the space after `,` .

zulip/zulip/__init__.py
```diff
@@ -1076,6 +1076,21 @@ def delete_stream(self, stream_id):
             method='DELETE',
         )
 
+    def get_member(self, user_id, request=None):
```
timabbott: Oh, let's call this `get_user`, not `get_member`.  That's more consistent naming with the rest of the product.

(And I think we'll want to rename `get_members` to `get_users`)

zulip/zulip/__init__.py
```diff
@@ -1076,6 +1076,21 @@ def delete_stream(self, stream_id):
...
+        # type: (int, Optional[Dict[str, Any]]) -> Dict[str, Any]
+
+        '''
+            Example usage:
+
+            >>> client.get_member(8, {'include_custom_profile_fields': True})
+            {'result': 'success', 'msg': '', 'user': [{...}, {...}]}
+        '''
+        return self.call_endpoint(
+            url='users/'+str(user_id),
```
timabbott: We generally do space around operators like `+`; see PEP-8 which we follow on this.

But probably it's better to do string formatting, rather than `+`, here.

akashaviator:
@timabbott made those changes. I was trying to rename examples/list-members to list-users , but got "conflicting files" warning here.

timabbott:
I merged the first commit.  For the rename commit, we should make the old function name work for backwards-compatibility.  Otherwise, any existing API clients will break on upgrade.

(And once we merge that and do a release, we should make a PR to update the documentation)

akashaviator:
@timabbott So do you mean that I should create a new wrapper for the same endpoint, named as get_users ?

timabbott:
I was suggesting we rename `get_members` to `get_users` and do the wrappers in the other direction.

(That would be more clear about which is the preferred option, the one that's not a wrapper).

akashaviator:

zulip/zulip/__init__.py
```diff
@@ -1133,16 +1133,19 @@ def update_user_by_id(self, user_id: int, **request: Any) -> Dict[str, Any]:
...
+            See examples/list-users for example usage.
         '''
         return self.call_endpoint(
             url='users',
             method='GET',
             request=request,
         )
 
+    def get_members(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
+        return self.get_users(request=request)
```
timabbott, line 1147: Let's add a comment that this is a backwards-compatibility name.  I can take care of it before merging.

timabbott:
Merged after adding the comment discussed above, thanks @akashaviator!  

We'll probably want to do similar renames to the other items we wanted to change in @orientor's `OpenationId` PR.

PR #549, opened 2020-03-14
api: Fix undefined variable reference from previous commit.

mateuszmandera:
Fixes travis failures.


neiljp:
@timabbott Without this fix, travis is broken for mypy, and so is usage of `RandomExponentialBackoff`.

I haven't looked into this implementation as I assume it was reviewed OK, but this LGTM as a fix.

PR #550, opened 2020-03-15
api: Implement GET /messages/matches_narrow endpoint

akashaviator:
This implements GET /messages/matches_narrow api endpoint.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -687,6 +687,23 @@ class Client(object):
             request=message_filters
         )
 
+    def check_messages_match_narrow(self, **request):
+        # type: (Dict[str, Any]) -> Dict[str, Any]
+
+        '''
+            Example usage:
+
+            >>> client.check_messages_match_narrow(msg_ids=[11, 12],
+                narrow=[{'operator': 'has', 'operand': 'link'}]
+            )
+            {'result': 'success', 'msg': '', 'messages': [{...}, {...}]}
+        '''
+        return self.call_endpoint(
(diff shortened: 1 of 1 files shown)


zulip/zulip/__init__.py
```diff
@@ -708,6 +708,20 @@ def get_messages(self, message_filters):
             request=message_filters
         )
 
+    def get_messages_matching_narrow(self, request):
+        # type: (Dict[str, Any]) -> Dict[str, Any]
+        '''
+            Example usage:
+
+            >>> client.get_messages_matching_narrow(request)
```
timabbott: I think we better provide an example narrow input in `request` rather than a variable name, to explain how to use this.  Might be worth commenting that the format is the same as for `get_messages`.

akashaviator:
I think that the format for `request` here is different from get_messages, apart from the `narrow` key. Did I misunderstand you ?

zulip/zulip/__init__.py
```diff
@@ -688,6 +688,23 @@ def get_messages(self, message_filters):
             request=message_filters
         )
 
+    def get_messages_matching_narrow(self, request):
```
timabbott, line 691: I think this should be called `check_messages_match_narrow` -- the purpose is to determine whether a set of messages appear in a narrow (E.g. to determine whether to add a new message that just arrived to an open `search` narrow in the webapp).

akashaviator:
@timabbott I've updated this PR.

akashaviator:

timabbott:
Yeah, this is great, merged!

Issue #552, opened 2020-03-17
bots: Do not remove self from the list of recipients of the incoming message.


aero31aero:
The current behavior of `send_reply` can be traced back to ad25f278534bc21310c1e3bd8a8021154bc19126.

(This change is part of the larger changes including some in the server repo to improve the UX for using bots in group PMs.)

TAM360:
Hey @aero31aero, can I work on this issue?

aero31aero:
Hey @TAM360 there's already an open PR fixing this issue. If you want a bots related issue and can't find one, you can start a discussion in #integrations stream on chat.zulip.org and we can find you a project. 😃

TAM360:
Best!

Issue #551, opened 2020-03-17
bots: Do not pass all messages to bot in a group PM.


aero31aero:
If a bot is part of a group PM, the conversation can easily get out of hand.

![image](https://user-images.githubusercontent.com/8033238/76867515-9fdc9e00-688b-11ea-8965-a4df63687532.png)

Instead, we should apply the same rules as stream messages for PMs with more than 2 recipients (mandatory @-mention before the bot's name).

TAM360:
@aero31aero does this have an open PR too?

xeruf:
is there a straightforward way for a bot to opt out of this behavior and handle all messages received in a group?

timabbott:
No but we'd definitely accept a PR to add an option to do it. Might be worth discussing naming in #integrations on chat.zulip.org before implementing, but it's not hard to find/replace.

PR #554, opened 2020-03-18
Bots in Group PMs

aero31aero:
Refer to https://github.com/zulip/zulip/issues/14228 for more details. This PR addresses #551 and #552.

### DIFF
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -165,7 +165,7 @@ class ExternalBotHandler:
         if message['type'] == 'private':
             return self.send_message(dict(
                 type='private',
-                to=[x['email'] for x in message['display_recipient'] if self.email != x['email']],
+                to=[x['email'] for x in message['display_recipient']],
                 content=response,
                 widget_content=widget_content,
             ))
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -165,7 +165,7 @@ class ExternalBotHandler:
         if message['type'] == 'private':
             return self.send_message(dict(
                 type='private',
-                to=[x['email'] for x in message['display_recipient']],
+                to=[x['id'] for x in message['display_recipient']],
                 content=response,
                 widget_content=widget_content,
             ))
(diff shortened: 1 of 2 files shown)


zulip_bots/zulip_bots/lib.py
```diff
@@ -166,7 +166,7 @@ def send_reply(self, message: Dict[str, Any], response: str, widget_content: Opt
         if message['type'] == 'private':
             return self.send_message(dict(
                 type='private',
-                to=[x['email'] for x in message['display_recipient'] if self.email != x['email']],
+                to=[x['email'] for x in message['display_recipient']],
```
timabbott: Can we change this to use user IDs instead while we're updating it?

zulip_bots/zulip_bots/lib.py
```diff
@@ -166,7 +166,7 @@ def send_reply(self, message: Dict[str, Any], response: str, widget_content: Opt
         if message['type'] == 'private':
             return self.send_message(dict(
                 type='private',
-                to=[x['email'] for x in message['display_recipient'] if self.email != x['email']],
```
neiljp: This was one of my very early contributions. Looking at it again now, my thought would be that it was intended to avoid a bot triggering itself with its own message, due to auto-mentions in PMs. Can that happen now, assuming it did before?

If so, this behavior is maybe countered by the other commit (though not fully?), so should these commits be squashed or restructured?

neiljp: Does this have to be paired with server logic like zulip/zulip#14241, to not break?

aero31aero: No. This will work fine either way. Paired with the the server change, we address the case when the bot is IN the list of recipients but afterwards, the bot won't be (since the server won't automatically add the sender to recipients list).

zulip_bots/zulip_bots/lib.py
aero31aero: Right!

showell:
@aero31aero What's the latest status on this?

aero31aero:
Got sidetracked with other stuff, I'll work on this today/tomorrow.

Issue #534, opened 2020-02-11
API: Support delete all messages from a topic


IDerr:
Hello,
I think that delete a topic is not supported by the python API, it was added to this commit:
https://github.com/zulip/zulip/commit/ac55a5222c977ae2c507fb34ec5081c6ab018c16

Thanks a lot for your work :)

arunikayadav42:
@zulipbot claim

zulipbot:
Welcome to Zulip, @arunikayadav42! We just sent you an invite to collaborate on this repository at https://github.com/zulip/python-zulip-api/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

arunikayadav42:
@IDerr what exactly is to be done in this issue? Is the pull request to be fixed?

zulipbot:
Hello @arunikayadav42, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

IDerr:
oops, haven't seen your answer, I think there's no method in the python api to delete a topic like in the PR I have linked above.

maanuanubhav999:
@zulipbot claim

zulipbot:
Welcome to Zulip, @maanuanubhav999! We just sent you an invite to collaborate on this repository at https://github.com/zulip/python-zulip-api/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
[...]

PR #556, opened 2020-03-19
#534 delete_topic method and example

maanuanubhav999:
@timabbott please look into the PR. for #534

### DIFF
--- a/zulip/zulip/examples/delete-topic
+++ b/zulip/zulip/examples/delete-topic
@@ -1,0 +1,23 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import argparse
+
+import zulip
+
+usage = """delete-topic <stream_id/stream_name, topic_name>
+
+Example: delete-topic 1 "Denmark1 " or "Denmark" "Denmark1"
+"""
+
+
+parser = zulip.add_default_arguments(argparse.ArgumentParser(usage=usage))
+parser.add_argument('stream_id')
+parser.add_argument('topic_name', type=str)
+options = parser.parse_args()
+
+client = zulip.init_from_options(options)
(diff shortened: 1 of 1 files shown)


zulip/zulip/examples/delete-topic
```diff
@@ -0,0 +1,21 @@
...
+import zulip
+
+usage = """delete-topic <streams_id, topic_name>
+
+Example: delete-topic 1 "Denmark1"
+"""
+
+
+parser = zulip.add_default_arguments(argparse.ArgumentParser(usage=usage))
+parser.add_argument('streams_id', type=int)
```
timabbott: It's clear you haven't tested this, because this should be `stream_id`, not `streams_id`.

timabbott: Also, I think this example should support passing a `stream_name` instead of a `stream_id`; it should be able to detect based on whether what it got is a number, and if it is a string, it can look up that stream by name using the API and act appropriately.

timabbott:
Thanks for working on this!  I posted a comment above; please also test your changes manually and post a transcript of your using this to delete a topic in your Zulip development environment.

I posted a couple comments; you should also clean up your commit history.  Check out the Zulip commit guidelines for more details: https://zulip.readthedocs.io/en/latest/contributing/version-control.html

zulip/zulip/examples/delete-topic
maanuanubhav999: Is there automatic testing for this? or just manual, yeah sure will do that.

maanuanubhav999:
2020-03-22-191729_1832x195_scrot
2020-03-22-182947_739x66_scrot

maanuanubhav999:
had some issue with this pr opening a new pr

PR #557, opened 2020-03-21
twitter_bot: Solve retweet issue.

orientor:
The twitter bot initially posted all kind of tweets. Change it
so that on encountering a retweet either post the original tweet
if it hasn't been posted before. Else try to edit the number of
retweets in the message of the original tweet.


zulip/integrations/twitter/twitter-bot
```diff
@@ -258,15 +268,33 @@ for status in statuses[::-1][:opts.limit_tweets]:
...
-    }
-
-    ret = client.send_message(message)
+    # If we are editing message it implies the current tweet is a retweet
+    # and has been posted before. Hence try to edit the older tweet and
+    # change its number of retweets but if it has passed its edit time
+    # do nothing.
+    if edit_message is True:
+        message_data = {
+            "message_id": message_id,
```
rht: `message_id` exists only when the `try` code above doesn't raise an exception, which means it only exists within that scope. Shouldn't `message_id` be initialized to have a default value of 0 first?

orientor: Added accordingly.

rht:
From my reading of the current state of the PR, there is no adding a number of retweets in the body of the message yet (a). What happens is that the message is being edited with the message_id pointing to the latest message. Is feature (a) too unwieldy to implement?

orientor:
> From my reading of the current state of the PR, there is no adding a number of retweets in the body of the message yet (a). What happens is that the message is being edited with the message_id pointing to the latest message. Is feature (a) too unwieldy to implement?

@rht I am storing the `twitter id: message id` key-value pairs for tweets that have been posted in the config file. So `message_id` will point to the older message and that will be edited. I didn't understand the "pointing to latest message" part.

orientor:
Screenshot from 2020-03-21 23-57-20
@rht Added number of retweets to all messages.

zulip/integrations/twitter/twitter-bot
```diff
@@ -259,14 +258,47 @@ for status in statuses[::-1][:opts.limit_tweets]:
...
+        }
+
+        ret = client.send_message(message)
+        if ret['result'] == 'success':
+            if 'twitter' not in config_internal.sections():
+                config_internal.add_section('twitter')
+            if status.retweeted_status is None:
+                config_internal.set('twitter', str(status.id), str(ret['id']))
+            else:
+                config_internal.set('twitter', str(status.retweeted_status.id), str(ret['id']))
```
rht: This means that the config file content will grow with the number of retweeted messages?

orientor: Yes. I am using it as a kind of database.

rht: Maybe the message id should be stored in a separate dedicated key-value database?

orientor: Yes that would be better. Can you suggest what I should use for the key-value database?

rht: Can't come up with any solution on top of my head. Haven't seen any being used by python-zulip-api bots before. But if other bots will become more complex to require a db, they might be using the same db we would use here.

rht: After a quick search, I found pickledb, a nosql key-value db in pure python. Unfortunately I can't be decisive about this, given that there are several alternatives. Maybe we should discuss in czo.

PIG208:
Perhaps it will be feasible to use bot storage to store the key-value pairs.

PR #560, opened 2020-03-25
State support for & test against python 3.8

neiljp:
This is a minimal commit, but hopefully should make it clear to users that we now support this version.

I assume we're happy with an extra travis job for this.


timabbott:
Closing as I just merged the exact duplicate #561. :)

PR #562, opened 2020-03-26
api: Implement GET /attachments endpoint

akashaviator:
This implements the GET /attachments endpoint.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -718,6 +718,19 @@ class Client(object):
             files=[file]
         )
 
+    def get_attachments(self):
+        # type: () -> Dict[str, Any]
+        '''
+            Example usage:
+
+            >>> client.get_attachments()
+            {'result': 'success', 'msg': '', 'attachments': [{...}, {...}]}
+        '''
+        return self.call_endpoint(
+            url='attachments',
+            method='GET'
+        )
+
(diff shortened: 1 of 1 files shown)


akashaviator:
@timabbott Can you have a look?

zulip/zulip/__init__.py
```diff
@@ -718,6 +718,19 @@ def upload_file(self, file):
             files=[file]
         )
 
+    def get_attachments(self):
+        # type: () -> Dict[str, Any]
+        '''
+            Example usage:
+
+            >>> client.get_attachments()
+            {'result': 'success', 'msg': '', 'attachments':[{...}, {...}]}
```
timabbott, line 727: There should be a space after the `:` here.

Are you generating these manually rather than copy-pasting output?  That tends to result in typos like this.

akashaviator:
@timabbott Other such examples seemed to follow this style to represent the output, rather than being the exact copy pasted outputs. I don't add these without testing the endpoints manually though. I'll make sure to not produce such typos again :).

timabbott:
Yeah, the `...` is fine; I just mean the there's no space between `:` and `attachments`.

akashaviator:
@timabbott I'll keep that in mind.

timabbott:
(Let me know when this is updated).

PR #563, opened 2020-03-27
bots: Fix tictactoe, connect_four emoji and computer play message

fishfred:
Some of the emojis that the tictactoe and connect_four bots were using were removed
I've replaced them with alternative existing emojis

The bots also say `You cannot play against the computer in this game.` even if you can. This has also been fixed


showell:
Merged, thanks @fishfred!

PR #570, opened 2020-04-01
api: Implement PATCH /users/{user_id} endpoint

akashaviator:
This implements PATCH /users/{user_id} endpoint.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -1084,6 +1084,25 @@ class Client(object):
             method='DELETE',
         )
 
+    def update_user_by_id(self, user_id, **request):
+        # type: (int, **Any) -> Dict[str, Any]
+
+        '''
+            Example usage:
+
+            >>> client.update_user_by_id(8, full_name="New Name")
+            {'result': 'success', 'msg': ''}
+        '''
+
+        for key, value in request.items():
+            request[key] = json.dumps(value)
+
+        return self.call_endpoint(
+            url='users/{}'.format(user_id),
+            method='PATCH',
+            request=request
+        )
+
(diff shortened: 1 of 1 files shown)


akashaviator:
@timabbott This is how we pass request to this endpoint currently -
```request={'full_name': json.dumps("New Name")}```
Should we refactor it to accept 
```request={'full_name': "New Name"}```

**Comment by timabbott** [comment]
Yes, definitely.  We should be exposing a clean, Pythonic interface in the API.

**Comment by akashaviator** [comment]
@timabbott Can you have a look?

File: zulip/zulip/__init__.py [hunk 1]
```diff
@@ -1071,6 +1071,24 @@ def get_user_by_id(self, user_id, request=None):
...
+            >>> client.update_user_by_id(8, {'full_name': "NewName"})
+            {'result': 'success', 'msg': ''}
+        '''
+        for key, value in user_data.items():
+            user_data[key] = json.dumps(value)
+
+        return self.call_endpoint(
+            url='users/{}'.format(user_id),
+            method='PATCH',
+            request=user_data
```
timabbott, line 1089: Why not do `**request` so that the user can just use `client.update_user_by_id(8, full_name="NewName")`?  The `get_streams` endpoint uses this pattern, and I think it's a good one.  

(We should change your other new endpoints, like `get_user_by_id`, to use this as well, while they're still new and we can freely change the Python API)

akashaviator:
@timabbott I have updated the PR.

timabbott:
@akashaviator can you rebase past the merge conflicts?

akashaviator:

PR #572, opened 2020-04-01
api: Implement DELETE /users/{user_id} api endpoint

akashaviator:
This implements DELETE /users/{user_id} api endpoint.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -1068,6 +1068,20 @@ class Client(object):
             request=request,
         )
 
+    def deactivate_user_by_id(self, user_id):
+        # type: (int) -> Dict[str, Any]
+
+        '''
+            Example usage:
+
+            >>> client.deactivate_user_by_id(8)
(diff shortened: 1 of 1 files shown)


zulip/zulip/__init__.py
```diff
@@ -1071,6 +1071,20 @@ def get_user_by_id(self, user_id, request=None):
             request=request,
         )
 
+    def deactivate_user_by_id(self, user_id):
+        # type: (int) -> Dict[str, Any]
+
+        '''
+            Example usage:
+
+            >>> client.delete_user_by_id(8)
```
rht, line 1080: Why is this called `delete_user_by_id` when the function name is `deactivate_user_by_id`?

akashaviator: Yeah, that's a typo. I'll fix that.

akashaviator:
@rht  Thanks for pointing that out !

akashaviator:
@rht Fixed it. Thanks for pointing that out !

PR #574, opened 2020-04-02
lint: Remove explicit python invocation check.

rht:
https://chat.zulip.org/#narrow/stream/127-integrations/topic/python-zulip-api.20to.20use.20python3/near/846057


rht:
@showell

Issue #573, opened 2020-04-02
Migrate the codebase to Python 3.


rht:
Python 2 is EOL since Jan 1st 2020.

See also https://chat.zulip.org/#narrow/stream/127-integrations/topic/python-zulip-api.20to.20use.20python3.

PR #575, opened 2020-04-02
integrations: Migrate to Python 3.

rht:
Partially addresses #573.


rht:
For the six imports, I have only tested that the imports work, which hopefully is sufficient.

PR #576, opened 2020-04-03
bridges: Remove sender field from message data.

rht:
For the reason why, see https://chat.zulip.org/#narrow/stream/127-integrations/topic/bridge.20between.20zulips.


showell:
Thanks for merging!

PR #579, opened 2020-04-03
api: Refactor get_user_by_id to take keyword arguments

akashaviator:
This refactors get_user_by_id to take arguments like
`include_custom_profile_fields = True` instead of
{include_custom_profile_fields: True}.


timabbott:
Merged after fixing the `.` at the end of the commit message.

Can you now submit a change to update the /api docs to use this updated form?

PR #583, opened 2020-04-11
api: Implement POST /users/{user_id}/reactivate endpoint.

akashaviator:
This implements POST /users/{user_id}/reactivate endpoint.


akashaviator:

PR #585, opened 2020-04-17
pypi: Prepare for the next major release.

eeshangarg:
@timabbott: FYI :)


timabbott:
Merged.  I think we should add an additional note about the package being available for Python 2 with (whatever the last released version supported that was)?

PR #588, opened 2020-04-19
Cleanups to pass Flake8, and replace pycodestyle with Flake8

andersk:


andersk:
It’s possible that some of the expressions behind by the two “remove unused variables” commits should also be removed.

timabbott:
This is great, merged, thanks @andersk!  I agree it may be worth looking more at the stuff adjacent to the unused variable commits; some of it smells like bugs.

Issue #533, opened 2020-01-29
api: Give quicker feedback if your credentials are wrong


showell:
When you use the api and instantiate a Client, it will succeed as long as the credentials are well-formed, even if the credentials aren't actually valid.  Of course, as soon as you try to use them, you'll get a proper error, but it might make sense to have some kind of "ping" method that you can call that literally just makes sure you can connect to the server, so that program can handle that error immediately before starting real work.

More discussion here:

https://chat.zulip.org/#narrow/stream/127-integrations/topic/python.20client/near/813065

Even though this might be a quick fix in terms of the amount of code, I recommend that you work on this only after you've worked with the API a bit and have a feel for the use cases.  A big part of this issue is figuring out the strategy that makes the most sense here.

We'll also want to improve the docs here to reflect any new methods we add or best practices we decide to recommend.

Kushal-kothari:
@zulipbot claim

zulipbot:
Welcome to Zulip, @Kushal-kothari! We just sent you an invite to collaborate on this repository at https://github.com/zulip/python-zulip-api/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

Kushal-kothari:
@showell  . Am I on the correct path to interpret the issue Eg: whenever a user try to use an api for some random reason "x"  (let say to create a bot and get it integrated).
After adding the  wrong credentials lets say (incorrect email id / incorrect token) we need a method handler that checks and give the feedback that   the entered credentials are valid or not  immediately and not after the whole process.

zulipbot:
Hello @Kushal-kothari, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

[...]

PR #592, opened 2020-04-20
zulip: Add authentication before initializing client.

TAM360:
Before sending any message to a user/stream via CLI, user's credentials are verified first. The idea is to update the users before they start using it (#533). This is done using **validate_credentials()** method which returns returns the result of **get_profile** API. The status of result is used to determine if the credentials are valid or not.


zulip/zulip/__init__.py
```diff
@@ -331,7 +341,7 @@ def __init__(self, email: Optional[str] = None, api_key: Optional[str] = None, c
             if cert_bundle is None and config.has_option("api", "cert_bundle"):
                 cert_bundle = config.get("api", "cert_bundle")
             if insecure is None and config.has_option("api", "insecure"):
-                # Be quite strict about what is accepted so that users don't
+                # Be quite strict about what is ainvalidccepted so that users don't
```
aero31aero, line 344: Accidental change?

zulip/zulip/send.py
```diff
@@ -71,28 +71,32 @@ def main() -> int:
...
+                'content': options.message,
+                'to': options.recipients,
+            }
+
+        if not do_send_message(client, message_data):
+            return 1
+        return 0
+    except zulip.InvalidCredentialsError as error:
+        logging.error(error)
+        return 0
```
aero31aero, line 99: We should return a non-zero value here. 0 implies the process went smoothly, without errors.

zulip/zulip/__init__.py
```diff
@@ -226,12 +226,18 @@ def init_from_options(options: Any, client: Optional[str] = None) -> 'Client':
...
-                  client_cert_key=options.client_cert_key)
+
+    new_client = Client(email=options.zulip_email, api_key=options.zulip_api_key,
+                        config_file=options.zulip_config_file, verbose=options.verbose,
+                        site=options.zulip_site, client=client,
+                        cert_bundle=options.cert_bundle, insecure=options.insecure,
+                        client_cert=options.client_cert,
+                        client_cert_key=options.client_cert_key)
+
+    if new_client.get_profile()["result"] == "error":
```
aero31aero, line 237: Awesome, this sounds like the code we want. Just one thing I'd change here: we have some more places that also call `Client()` to create a new client. (`git grep 'Client(' | wc -l` shows 31 cases, most of which are calls to `zulip.Client()`. We also want to throw this error in thjose cases, thus, we should have this block of code in the Client class, just before it returns the client object.

zulip/zulip/__init__.py
TAM360: Yup

zulip/zulip/__init__.py
TAM360: I thought 0 meant an error. does -1 work in this case?

zulip/zulip/send.py
```diff
@@ -71,28 +71,32 @@ def main() -> int:
...
+            message_data = {
+                'type': 'private',
+                'content': options.message,
+                'to': options.recipients,
+            }
+
+        if not do_send_message(client, message_data):
+            return 1
+        return 0
+    except zulip.InvalidCredentialsError as error:
```
neiljp, line 97: It might be clearer if we moved the try block to just be around the relevant code that raises the exception?

TAM360: I think it makes sense to keep the `client = zulip.init_from_options(options)` part in try/except block and move others outside.

showell:
@TAM360 Do you think you have time in the next couple days to address the remaining comments?  It looks like it's mostly fixing a typo and limiting the scope of the try/except block.

TAM360:
@showell yeah I do. I am gonna try to finish this by Sunday.

showell:
@TAM360 Can you rebase these commits into a single commit?  We don't really want or need the merge commits or the commits that are just code-review fixes.

TAM360:
I've already pushed 3-4 small commits to my branch. Will they get affected if I rebase my commits now and then push it?

showell:
> 
> 
> @showell I've already pushed 3-4 small commits to my branch. Will they get affected if I rebase my commits now and then push it?

Yes, but that's the idea behind rebasing--I **want** you to replace those.

TAM360:
@showell all right.

aero31aero:
From my previous comment:

> We should have this block of code in the Client class, just before it returns the client object.

It looks like you've added the code to throw the error in multiple places; we only need it in the zulip Client's constructor.

TAM360:
@aero31aero I've added the validation logic in Client's constructor but as a result it's failing 4 of the unit test. Should I update those tests or add new ones?

TAM360:
@zulipbot abandon

andersk:
(I’m not sure this PR is still active, but it’s apparently still being referenced.)

`"result": "error"` doesn’t necessarily imply that the credentials are wrong. We also need to check for HTTP status 401 or `"code": "INVALID_API_KEY"`.

Also, `sys.exit()` returns a successful exit code that’s inappropriate for an error condition. I think in most cases we just want to let the exception propagate up.

showell:
I am gonna close this since @TAM360 has abandoned it.  @TAM360 Thanks for working on this!

PR #593, opened 2020-04-21
ExternalBotHandler: Support adding reactions to message for a bot..

iishiishii:
This adds a react() function that allows a bot to react to a message with an example and tests.


showell:
@iishiishii The fix here seems fine.  I think we want to add an example of this--maybe in the `helloworld` bot we can add a reaction?

iishiishii:
@showell Thank you for your review. I'm writing the example in the `helloworld` bot and the test for this function in `test_lib.py`.

showell:
Merged, thanks @iishiishii and @aero31aero!

PR #596, opened 2020-04-27
Add search and JQL functions to Jira bot

pemontto:
This PR adds functionality for searching issue summaries with `search` and full blown JQL search with `jql`.

It also supports the user defining their own HTTP scheme in the `domain`, this is useful for circumstances where Jira is served behind an authenticating proxy. This allows you to connect to a secondary listener that might not be HTTPS. An optional configuration parameter was also added `display_url` so the URL for the output links can also be controlled.

E.g. Users access Jira through a proxy at `https://jira.mydomain.com`. The actual Jira host listens on `http://jira1.mydomain.com:8081`. You can now configure this using
```
[jira]
username = test
password = test_password
domain = http://jira1.mydomain.com:8081
display_url = https://jira.mydomain.com
```

### DIFF
--- a/zulip_bots/zulip_bots/bots/jira/jira.conf
+++ b/zulip_bots/zulip_bots/bots/jira/jira.conf
@@ -2,3 +2,4 @@
 username = <your Jira email address>
 password = <your Jira password>
 domain = <your Jira domain>
+# display_url = [optional] <your front facing Jira URL to use in links>
--- a/zulip_bots/zulip_bots/bots/jira/jira.py
+++ b/zulip_bots/zulip_bots/bots/jira/jira.py
@@ -27,6 +27,7 @@ EDIT_REGEX = re.compile(
     '( by making due "(?P<due_date>.+?)")?'
     '$'
 )
+SEARCH_REGEX = re.compile('search "(?P<search_term>.+)"$')
 HELP_REGEX = re.compile('help$')
 
 HELP_RESPONSE = '''
@@ -49,6 +50,23 @@ Jira Bot:
  > - Project: *Bots*
  > - Priority: *Medium*
  > - Status: *To Do*
+
+---
+
+**search**
+
+`search` takes in a search term and returns issues with matching summaries. For example,
+
+you:
+
+ > @**Jira Bot** search "XSS"
+
+Jira Bot:
+
+ > **Search results for *"XSS"*:**
+ >
+ > - ***BOTS-5:*** Stored XSS **[Published]**
+ > - ***BOTS-6:*** Reflected XSS **[Draft]**
 
 ---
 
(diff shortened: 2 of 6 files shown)


zulip_bots/zulip_bots/bots/jira/jira.conf
```diff
@@ -2,3 +2,4 @@
 username = <your Jira email address>
 password = <your Jira password>
 domain = <your Jira domain>
+# domain_url = [optional] <your front facing Jira URL to use in links>
```
aero31aero, line 5: Missing new line at the and of this file.

zulip_bots/zulip_bots/bots/jira/jira.py
```diff
@@ -127,7 +163,40 @@ def initialize(self, bot_handler: Any) -> None:
             raise KeyError('No `domain` was specified')
 
         self.auth = make_jira_auth(username, password)
-        self.domain_with_protocol = 'https://' + domain
+
+        # Allow users to override the HTTP scheme
```
aero31aero, line 167: It would be better to split this into its own commit. I'm not sure if this a good change; we should try to encourage using https and not provide an easy way to override that.

zulip_bots/zulip_bots/bots/jira/jira.py
```diff
@@ -127,7 +163,40 @@ def initialize(self, bot_handler: Any) -> None:
...
+
+        # Use the front facing URL in output
+        self.display_url = config.get('display_url')
+        if not self.display_url:
+            self.display_url = self.domain_with_protocol
+
+    def jql_search(self, jql_query:str) -> str:
+        UNKNOWN_VAL = '*unknown*'
+        jira_response = requests.get(
+            f'{self.domain_with_protocol}/rest/api/2/search?jql={jql_query}&fields=key,summary,status',
```
aero31aero, line 181: Unfortunately, we cannot use f-strings yet. We have to support python 3.5 as well, thus, we should use `"{}".format()` for now.

zulip_bots/zulip_bots/bots/jira/jira.py
```diff
@@ -127,7 +163,40 @@ def initialize(self, bot_handler: Any) -> None:
...
+        ).json()
+
+        url = self.display_url + '/browse/'
+        errors = jira_response.get('errorMessages', [])
+        results = jira_response.get('total', 0)
+
+        if errors:
+            response = 'Oh no! Jira raised an error:\n > ' + ', '.join(errors)
+        else:
+            response = f'*Found {results} results*\n\n'
```
aero31aero, line 192: Similarly here.

zulip_bots/zulip_bots/bots/jira/jira.py
```diff
@@ -127,7 +163,40 @@ def initialize(self, bot_handler: Any) -> None:
...
-        self.domain_with_protocol = 'https://' + domain
+
+        # Allow users to override the HTTP scheme
+        if re.match(r'^https?://', domain, re.IGNORECASE):
+            self.domain_with_protocol = domain
+        else:
+            self.domain_with_protocol = 'https://' + domain
+
+        # Use the front facing URL in output
+        self.display_url = config.get('display_url')
```
aero31aero, line 174: Similarly, it makes sense to have adding display_url as a separate commit.

zulip_bots/zulip_bots/bots/jira/jira.conf
aero31aero: You probably meant to write `display_url` here.

pemontto:
Thanks for the quick review, I was hesitant to spam the commit history. Will address those issues now.

zulip_bots/zulip_bots/bots/jira/jira.conf
pemontto: Fixed

timabbott:
Merged, after squashing that last documentation fixup commit, thanks @pemontto!

PR #597, opened 2020-05-02
Remove six/typing dependencies & add project_urls

neiljp:
Even though the this repo is now python3.5+, we have retained a `six` and `typing` dependency, which we can now remove.

Also, the URL for each package currently points to zulip.org -> zulipchat.com. The second commit should add links in the PyPI sidebar to the git repo and the zulipchat.com/api page, which otherwise are not particularly obvious.


neiljp:
@eeshangarg Was the requirements change left out in 0.7.0, or am I missing something?

eeshangarg:
@neiljp You mean this change? Yes, I think it was, but we can make another one.

PR #600, opened 2020-06-05
api: Update type annotation in 'remove_subscriptions'.

akashaviator:
This updates the type annotation for the 'principals' argument
to accept list of email addresses as well as user ids.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -15,7 +15,7 @@ import distro
 from configparser import SafeConfigParser
 import urllib.parse
 import logging
-from typing import Any, Callable, Dict, Iterable, IO, List, Mapping, Optional, Text, Tuple, Union
+from typing import Any, Callable, Dict, Iterable, IO, List, Mapping, Optional, Text, Tuple, Union, Sequence
 
 __version__ = "0.7.0"
 
@@ -1270,13 +1270,11 @@ class Client:
             request=request,
         )
 
-    def remove_subscriptions(self, streams: Iterable[str], principals: Optional[Iterable[str]] = None) -> Dict[str, Any]:
+    def remove_subscriptions(self, streams: Iterable[str],
+                             principals: Union[Sequence[str], Sequence[int]] = []) -> Dict[str, Any]:
         '''
             See examples/unsubscribe for example usage.
         '''
-        if principals is None:
-            principals = []
-
         request = dict(
             subscriptions=streams,
             principals=principals


zulip/zulip/__init__.py
```diff
@@ -1270,7 +1270,8 @@ def add_subscriptions(self, streams: Iterable[Dict[str, Any]], **kwargs: Any) ->
             request=request,
         )
 
-    def remove_subscriptions(self, streams: Iterable[str], principals: Optional[Iterable[str]] = None) -> Dict[str, Any]:
+    def remove_subscriptions(self, streams: Iterable[str],
+                             principals: Optional[List[Any]] = None) -> Dict[str, Any]:
```
timabbott, line 1274: This should be typed `Union[List[str], List[int]]`.

akashaviator:

zulip/zulip/__init__.py
```diff
@@ -1270,7 +1270,8 @@ def add_subscriptions(self, streams: Iterable[Dict[str, Any]], **kwargs: Any) ->
             request=request,
         )
 
-    def remove_subscriptions(self, streams: Iterable[str], principals: Optional[Iterable[str]] = None) -> Dict[str, Any]:
+    def remove_subscriptions(self, streams: Iterable[str],
+                             principals: Optional[List[Union[str, int]]] = None) -> Dict[str, Any]:
```
timabbott, line 1274: It should be a `Union[List`, not `List[Union`, since all the things in a list must be the same type.

akashaviator:

zulip/zulip/__init__.py
```diff
@@ -1270,16 +1270,15 @@ def add_subscriptions(self, streams: Iterable[Dict[str, Any]], **kwargs: Any) ->
...
         )
 
-    def remove_subscriptions(self, streams: Iterable[str], principals: Optional[Iterable[str]] = None) -> Dict[str, Any]:
+    def remove_subscriptions(self, streams: Iterable[str],
+                             principals: Optional[Union[List[str], List[int]]] = None) -> Dict[str, Any]:
         '''
             See examples/unsubscribe for example usage.
         '''
-        if principals is None:
-            principals = []
```
timabbott, line 1278: Why did you change this?  I think it was better with this on its own line.

showell: @timabbott The reason the change happened here was to work around `mypy` being finicky about shadowing a variable and thinking we've changed the type of it to `List[Any]` (figuratively).

More discussion here: https://chat.zulip.org/#narrow/stream/92-learning/topic/mypy.20error/near/902548

Are we enforcing that the Python versions >= 3.6 yet for the API?  This would influence our decision here on how best to deal with mypy.

akashaviator: @timabbott I was facing an issue. I've explained it here: https://chat.zulip.org/#narrow/stream/92-learning/topic/mypy.20error

timabbott: We probably shouldn't limit this to disallow 3.5 before it reaches EOL.  Seems really cheap to keep support and I'm sure someone will have a bad experience if we do.

akashaviator:

timabbott:
I don't think this latest change is an improvement over what we have in master.

akashaviator:

timabbott:
Merged, thanks @akashaviator!  `Sequence` is the right way to do this.

Issue #607, opened 2020-07-12
Potential code quality issues found


pnijhara:
I have forked this [repo](https://github.com/pnijhara/python-zulip-api/tree/pnijhara-patch-1) and later to check code quality, I ran a [DeepSource](https://deepsource.io) analysis on it. There I found around 700+ code quality issues, some of them can be easily fixed in a single click.

Those issues are categorized as:

Anti-pattern: 425 
Bug Risk: 49
Performance: 188
Security: 57

Find a detailed report of the issues found here -> https://deepsource.io/gh/pnijhara/python-zulip-api/issues/?category=recommended

I have also added a deepsource config file to let DeepSource analyze my forked repo. The link of that file can be found [here](https://github.com/pnijhara/python-zulip-api/blob/pnijhara-patch-1/.deepsource.toml).

I want to send a patch with some fix.

Also do let me know which of the above issues you think are false positives and which of them are potential issues.

timabbott:
We love taking patches that improve the codebase, and use other code scanners to look for them.  

Based on past experience, I'd be surprised if many of those "700 issues" are real problems -- this class of tool is known for generating a lot of spam because it helps sell that class of software.  And 188 "performance" issues in a small codebase like this API library is a sign that they're very aggressive about flagging things.

PR #608, opened 2020-07-20
Fix some code qulity issues

pnijhara:
Fixing some of the code quality mentioned in the issue #607.

This patch:
- Fixes dangerous default argument
- Uses literal syntax instead of function calls to create data structure
- Iterates dictionary directly
- Remove length check in favour of truthiness of the object

These are some of the antipatterns, bug-risk and performance issues that I have solved in this PR.
These issues are solved with the help of DeepSource autofix tool.

As discussed in the issue that many of the issues can be false positives. Those issues can be silenced by ignoring the issues in the issue category in the dashboard.

### DIFF
--- a/zulip/integrations/perforce/git_p4.py
+++ b/zulip/integrations/perforce/git_p4.py
@@ -648,7 +648,7 @@ def findUpstreamBranchPoint(head = "HEAD"):
     branches = p4BranchesInGit()
     # map from depot-path to branch name
     branchByDepotPath = {}
-    for branch in branches.keys():
+    for branch in branches:
         tip = branches[branch]
         log = extractLogMessageFromGitCommit(tip)
         settings = extractSettingsGitLog(log)
@@ -1352,7 +1352,7 @@ class P4Submit(Command, P4UserMap):
             p4_delete(f)
 
         # Set/clear executable bits
-        for f in filesToChangeExecBit.keys():
+        for f in filesToChangeExecBit:
             mode = filesToChangeExecBit[f]
             setP4ExecBit(f, mode)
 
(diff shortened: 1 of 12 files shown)


neiljp:
From my perspective the primary points worth addressing here would be the empty list default argument case.

I would suggest having each type of change in their own commit, to make review and selection of each easier, though personally I would not consider most of the other changes significant.

pnijhara:
@neiljp so should I send different PR for different type fo changes and make this one as fixing the default argument case?

andersk:
Rather than introducing `Optional` and `None` where it isn’t needed, I think it’s better to guard against the potential problems with mutable default parameters using read-only mypy types, as I did in zulip/zulip#15349.

```diff
-    def create_game_lobby(self, message: Dict[str, Any], users: List[str] = []) -> None:
+    def create_game_lobby(self, message: Dict[str, Any], users: Sequence[str] = []) -> None:

-    def join_game(self, game_id: str, user_email: str, message: Dict[str, Any] = {}) -> None:
+    def join_game(self, game_id: str, user_email: str, message: Mapping[str, Any] = {}) -> None:
```

pnijhara:
Ok! So should I revert back these changes from files?
@andersk could you please tell me which files do not require this type of change?

pnijhara:
ping! @andersk 
Need help!

andersk:
My comment is about all of the changes of this type, not about particular files.

pnijhara:
Sure! I will revert this change.

showell:
@pnijhara Do you still have time to clean up this PR?  I think we would need to get the build passing again (I haven't looked to see what's broken) and then do another round of review.

If you don't have time, no worries, we can probably have somebody take this over.

pnijhara:
@showell I sincerely apologise for the delay. I will revert the suggested change soon.

showell:
No apology necessary...thanks for the update.

PR #611, opened 2020-08-01
zephyr: Use exponential backoffs in retry loops.

alexmv:
This reduces the number of retries that might spam APIs.

There is some complexity here which is left un-managed -- for
instance, maybe_restart_mirroring_script does a number of restart
attempts, and then fails, but will be retried every 15s by the
surrounding `process_loop`.  Previously, it would merely have looped
forever inside maybe_restart_mirroring_script.

Three loops are intentionally left as infinite `while True` loops,
that merely cap their backoff at the default 90s.  Their callers do
not expect, or have any way to handle more gracefully, a failure of
the expected-infinite-loop in `process_loop` or `zulip_to_zephyr`.
They maintain their previous behavior of retrying forever, albeit more
slowly.

### DIFF
--- a/zulip/integrations/zephyr/zephyr_mirror_backend.py
+++ b/zulip/integrations/zephyr/zephyr_mirror_backend.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-from typing import Any, Dict, IO, List, Optional, Set, Text, Tuple, cast
+from typing import Any, Dict, IO, List, NoReturn, Optional, Set, Text, Tuple, cast
 from types import FrameType
 
 import sys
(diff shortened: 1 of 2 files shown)


alexmv:
See also #586.

zulip/integrations/zephyr/zephyr_mirror_backend.py
```diff
@@ -759,12 +770,13 @@ def maybe_forward_to_zephyr(message: Dict[str, Any]) -> None:
 def zulip_to_zephyr(options: int) -> None:
     # Sync messages from zulip to zephyr
     logger.info("Starting syncing messages.")
+    backoff = RandomExponentialBackoff(timeout_success_equivalent=120)
```
timabbott, line 773: I also note `timeout_success_equivalent` is undocumented; do you understand what it's for?

I guess 4e02777aaa9e788843f581868291e8c4b3696ace explains it pretty clearly, so maybe we just want to move some of that commit message into the comments.

And I think 120 is a good value for that here.

timabbott:
Posted some comments; otherwise lgtm.

Issue #612, opened 2020-08-03
Matrix bridge: Event object no longer has "membership" key.


rht:
The event object is [this line](https://github.com/zulip/python-zulip-api/blob/396ef1d11adfd401ff807bcb77abb109669725a1/zulip/integrations/bridge_with_matrix/matrix_bridge.py#L107) only has these keys:
`['unsigned', 'origin_server_ts', 'type', 'room_id', 'content', 'sender', 'event_id']`. `membership` is gone.

timabbott:
Is there any documentation for this change in the Matrix.org APIs?

rht:
The example in https://github.com/matrix-org/matrix-python-sdk/blob/master/samples/SimpleChatClient.py still uses "membership" which no longer works. The Matrix API documentation doesn't have changelog that says any deprecation. Because of #604, the library used in the Matrix bridge is no longer actively maintained.

@timabbott is it worth it to modernize the Matrix bridge, or should I make the Zulip Go API work?

timabbott:
Can you open an issue with the Matrix.org folks for that `membership` bug?  I think @Hypro999 might be interested in collaborating on fixing Matterbridge.

rht:
This will be closed by #723 since it uses matrix-nio.

PR #613, opened 2020-08-07
Matrix bridge: Change the condition of not_from_zulip_bot.

rht:


zulip/integrations/bridge_with_matrix/matrix_bridge.py
```diff
@@ -75,10 +75,7 @@ def _matrix_to_zulip(room: Any, event: Dict[str, Any]) -> None:
         zulip_bot_user = ('@%s:matrix.org' % matrix_config['username'])
         # We do this to identify the messages generated from Zulip -> Matrix
         # and we make sure we don't forward it again to the Zulip stream.
-        not_from_zulip_bot = (
-            'body' not in event['content']
```
neiljp, line 79: Was this a previous way to identify the user as a bot? Is this only in a certain version of the matrix library? If so, should we set a bound on it in `setup.py` with this change?

rht: The Matrix Python library hasn't been maintained since 2019, and so has been unchanged. It is the server that is changing its behavior, just like when `membership` key is gone.

rht: By the `membership` key bug, I was referring to #612.

PR #614, opened 2020-08-08
zulip: Add hash_util_decode() to decode server encoded URL excerpts

preetmishra:
This adds `hash_util_decode()` to decode a `hash_util_encode()` [present in [zulip/zulip's zerver/lib/url_encoding.py](https://github.com/zulip/zulip/blob/master/zerver/lib/url_encoding.py)] encoded string.

The intent is to facilitate code sharing among various python clients (primarily, Zulip Terminal).

The string replacement before the `unquote` is to recoup for the custom string replacements in [zulip/zulip's zerver/lib/url_encoding.py](https://github.com/zulip/zulip/blob/master/zerver/lib/url_encoding.py).

Conversation link: https://chat.zulip.org/#narrow/stream/127-integrations/topic/Decode.20server.20URLs/near/977374

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -1544,3 +1544,15 @@ class ZulipStream:
 
     def flush(self) -> None:
         pass
+
+def hash_util_decode(string: str) -> str:
+    """
+    Returns a decoded string given a hash_util_encode() [present in zulip/zulip's zerver/lib/url_encoding.py] encoded string.
+
+    Example usage:
+    >>> zulip.hash_util_decode('test.20here')
+    'test here'
+    """
+    # Acknowledge custom string replacements in zulip/zulip's zerver/lib/url_encoding.py before unquoting.
+    # NOTE: urllib.parse.unquote already does .replace('%2E', '.').
+    return urllib.parse.unquote(string.replace('.', '%'))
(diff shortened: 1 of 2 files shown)


zulip/zulip/__init__.py
```diff
@@ -1544,3 +1544,15 @@ def write(self, content: str) -> None:
...
+    """
+    Returns a decoded string given a hash_util_encode() [present in zulip/zulip's zerver/lib/url_encoding.py] encoded string.
+
+    Example usage:
+    >>> zulip.hash_util_decode('test.20here')
+    'test here'
+    """
+    # Acknowledge custom string replacements in zulip/zulip's zerver/lib/url_encoding.py before unquoting.
+    # NOTE: urllib.parse.unquote already does .replace('%2E', '.').
+    return urllib.parse.unquote(string.replace('.', '%'))
```
timabbott, line 1558: Can you add tests for this?  I think that'd help a lot in being confident in it.

preetmishra:
@timabbott Thanks for the comment. I have updated the PR with tests.

preetmishra:
@timabbott Thanks for the comment. I have updated the PR with tests.

timabbott:
Merged, thanks @preetmishra!  I think ultimately we're going to want some higher-level functions on top of this that take a URL and decode it, with a documented interface, but happy to merge this as progress.

PR #615, opened 2020-08-09
[BUGFIX] IRC bridge: Use connection initialized from aio_reactor and document --nickserv-pw.

rht:
Additionally, pin the irc library version to 18.0 because the newer
version has an error in the AioReactor:
```
  File "/home/rht/code/venv/lib/python3.8/site-packages/irc/bot.py", line 108, in run
    self.bot.reactor.scheduler.execute_after(intvl, self.check)
AttributeError: 'AioReactor' object has no attribute 'scheduler'
```


lu-zero:
I confirm it works :)

timabbott:
Merged, thanks @rht and @lu-zero!

PR #616, opened 2020-08-11
IRC bridge: Remove "_zulip" nickname append.

rht:
The current behavior is to automatically append "_zulip" to user-specified nick prefix, hence forcing the bridge nick to end with "_zulip". This current behavior is confusing the user, especially if the user is using a password to connect to an IRC server. Additionally, if nickname is already in use, simply raise an error instead of silently adding more underscores to the nick.

Testing plan: I have tested this PR on my own laptop.


timabbott:
How do we prevent mirroring loops without the `_zulip` suffix?

rht:
> How do we prevent mirroring loops without the `_zulip` suffix?

! I forgot about that. I suppose I can just exactly match the bot nickname.

rht:
@timabbott added the fix in the 3rd commit.

rht:
My testing worked because I was coincidentally using a nick that ends with "_zulip". Now it is no longer the case and still works.

zulip/integrations/bridge_with_irc/irc_mirror_backend.py
```diff
@@ -84,7 +85,7 @@ def forward_to_irc(msg: Dict[str, Any]) -> None:
     def on_privmsg(self, c: ServerConnection, e: Event) -> None:
         content = e.arguments[0]
         sender = self.zulip_sender(e.source)
-        if sender.endswith("_zulip@" + self.IRC_DOMAIN):
+        if sender.endswith(self.nickname + "@" + self.IRC_DOMAIN):
```
timabbott, line 88: Should these be `endswith` checks, or exact checks?

rht: I have replaced this line with exact check, and have tested it working.

rht:
@timabbott comment addressed.

timabbott:
Great; can you clean up the commit history?

(And I guess we'll want to document this, maybe as a first/preferred option above the Matrix version?)

rht:
Commits squashed.

I looked at https://github.com/zulip/zulip/blob/master/templates/zerver/integrations/irc.md. It looks fine and doesn't recommend Matrix at all. Even the sentence at https://github.com/zulip/zulip/blob/master/templates/zerver/integrations/matrix.md recommends the direct IRC mirror.

rht:
IMO, we should instead advice people to use Matterbridge, if they only want to mirror only to 1 topic in Zulip.
I have seen the Matterbridge in action in julialang.zulipchat.com, and it mirrors multiple protocol with only 1 configuration file.
The IRC bridge works, but with unavoidable error messages, that makes the script looks bad.

rht:
@timabbott ping

rht:
Rebased.

PR #619, opened 2020-08-31
New Zulip bot named bugzilla

miredo:
New bot to directly add new comments in Bugzilla from Zulip.

It works nice with https://github.com/lemoineat/Zulipzilla a Bugzilla extension that posts into Zulip bug changes.


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "python.pythonPath": "zulip-api-py3-venv/bin/python3.7"
+}
```
timabbott, line 3: Probably this should go in `.gitignore`; the Python version at least isn't fixed.

zulip_bots/zulip_bots/bots/bugzilla/doc.md
```diff
@@ -0,0 +1,30 @@
+# Bugzilla bot
+
+This bot allows to update directly Bugzilla from Zulip
+
+## Setup
+
+To use Bugzilla Bot, first set up `bugzilla.conf`. `bugzilla.conf` takes 2 options:
+ - site (the site like `https://bugs.xxx.net` that includes both the protocol and the domain)
+ - api_key (a Bugzilla API key)
```
timabbott, line 9: I think it'd be best to have an example in a markdown code block; without that, it isn't clear what format this file should be in.

zulip_bots/zulip_bots/bots/bugzilla/doc.md
```diff
@@ -0,0 +1,30 @@
...
+### comment
+
+With no argument, by default, a new comment is added to the bug that is associated to the topic.
+For example, on topic Bug 123,
+
+you:
+
+  > @**Bugzilla** A new comment
+
+Then `A new comment` is added to bug 123
```
timabbott, line 30: Would it be worth having explicit syntax for commenting on a bug when the topic has some other format?

.vscode/settings.json
miredo: In my last commit, the vs code settings file is ignored

zulip_bots/zulip_bots/bots/bugzilla/doc.md
miredo: I thought it was not required because an example was provided in the source code. But sure if it helps, it is worth adding it.
It is done in the last commit.

zulip_bots/zulip_bots/bots/bugzilla/doc.md
miredo: For sure, it could be a nice feature, but since it was mainly aimed to work with https://github.com/lemoineat/Zulipzilla I had no use of it. And I did not want to make the bot more complex if the feature is not really required. Using a topic to track a bug record is pretty convenient.

But if anybody wants to improve it  with that feature it would be nice ! (I don't have much more time to add new features on it for the moment).

.gitignore
```diff
@@ -57,3 +57,7 @@ botserverrc
 # Pycharm
 \.DS_Store
 \.idea/
+
+# VS Code
+.vscode/settings.json
```
LoopThrough-i-j, line 62: I think we can ignore `.vscode`.

LoopThrough-i-j:
Moreover, would you mind going through the [docs](https://zulip.readthedocs.io/en/latest/contributing/version-control.html) and fixing the commit history?

PR #620, opened 2020-09-09
FEAT: add method to move topics between streams

tmichela:
Move a topic from one stream to another,

Example usage:
```python
>>> client.move_topic('stream_a', 'stream_b', 'my_topic')
{'result': 'success', 'msg': ''}
```

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -1554,6 +1554,83 @@ class Client:
             request=request
         )
 
+    def move_topic(
+        self,
+        stream: str,
+        new_stream: str,
+        topic: str,
+        new_topic: Optional[str] = None,
+        message_id: Optional[int] = None,
+        propagate_mode: str = 'change_all',
+        notify_old_topic: bool = True,
+        notify_new_topic: bool = True
+    ) -> Dict[str, Any]:
+        '''
+            Move a topic from ``stream`` to ``new_stream``
+
+            The topic will be renamed if ``new_topic`` is provided.
+            message_id and propagation_mode let you control which messages
+            should be moved. The default behavior moves all messages in topic.
+
+            propagation_mode must be one of: `change_one`, `change_later`,
+            `change_all`. Defaults to `change_all`.
+
+            Example usage:
+
+            >>> client.move_topic('stream_a', 'stream_b', 'my_topic')
+            {'result': 'success', 'msg': ''}
+        '''
+        # get IDs for source and target streams
+        result = self.get_stream_id(stream)
+        if result['result'] != 'success':
+            return result
+        stream = result['stream_id']
+
+        result = self.get_stream_id(new_stream)
+        if result['result'] != 'success':
+            return result
+        new_stream = result['stream_id']
+
+        if message_id is None:
+            if propagate_mode != 'change_all':
+                raise AttributeError('A message_id must be provided if '
+                                     'propagate_mode isn\'t "change_all"')
+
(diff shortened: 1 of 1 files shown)


zulip/zulip/__init__.py
```diff
@@ -1554,6 +1554,53 @@ def set_typing_status(self, request: Dict[str, Any]) -> Dict[str, Any]:
...
+        new_topic: Optional[str] = None
+    ) -> Dict[str, Any]:
+        """Move a topic from ``stream`` to ``new_stream``
+
+        The topic will be renamed if ``new_topic`` is provided
+
+        Example usage::
+
+            >>> client.move_topic('stream_a', 'stream_b', 'my_topic')
+            {'result': 'success', 'msg': ''}
```
timabbott, line 1571: This is over-indented, I think.

zulip/zulip/__init__.py
```diff
@@ -1554,6 +1554,53 @@ def set_typing_status(self, request: Dict[str, Any]) -> Dict[str, Any]:
             request=request
         )
 
+    def move_topic(
+        self,
+        stream: str,
+        new_stream: str,
+        topic: str,
+        new_topic: Optional[str] = None
```
timabbott, line 1562: Ideally, this would support the `propagate_mode` option and a message ID threshold.  https://zulipchat.com/api/update-message#parameters has some basic documentation

zulip/zulip/__init__.py
```diff
@@ -1554,6 +1554,53 @@ def set_typing_status(self, request: Dict[str, Any]) -> Dict[str, Any]:
...
+        """
+        # ask the server for the latest message ID in the topic.
+        request = {
+            'anchor': 'newest',
+            'narrow': [{'operator': 'stream', 'operand': stream},
+                       {'operator': 'topic', 'operand': topic}],
+            'num_before': 1,
+            'num_after': 0,
+        }
+        reply = self.get_messages(request)
```
timabbott, line 1581: Let's call this `result`; we use that local variable name for API results in the codebase.

Also, can you add a few blank lines in this function to match the style of the codebase?

timabbott:
@tmichela thanks for working on this!  I posted a few comments.  It'd also be great if you wrote a proper commit message; 
Check out the Zulip commit message guidelines for more details: https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages.

@akashaviator can you do a review pass on this as well?

tmichela:
@timabbott thanks for the review and suggestions!

I think I've addressed your comments.

I also changed/fixed:
- get stream IDs from their name (got some issues while testing using the stream names, returning sometimes an error)
- added arguments for `message_id` and `propagate_mode` as you suggested
- added arguments to silence notification to old/new topic

zulip/zulip/__init__.py
tmichela: I've changed the docstring to match the exiting style

timabbott:
This is great, merged, thanks @tmichela!  

Would you be up for submitting a PR to document this in the examples in https://zulip.com/api/update-message?  I expect https://zulip.readthedocs.io/en/latest/documentation/api.html should cover the background you need; let me know if it's missing anything!

tmichela:
Sure, I'll look into this.

Do you want me to simply update the python example code?
It looks like this documentation reflects more of the direct restAPI call while this method wraps a few different calls to make it more convenient to use.

timabbott:
Yeah, exactly; I'd just include it in the same block where we document the more messy/direct way to do it.

PR #622, opened 2020-09-20
provision: Fix source command on Windows

QEDK:
Currently, the provision script gives us something like this on Windows:
```bash
Run the following to enter into the virtualenv:

  source C:\Users\qedk\python-zulip-api\zulip-api-py3-venv\Scripts\activate
```
However, `bash` does not work with  backslashes and will give us an error saying `No such file or directory`.
This commit addresses that so we get the correct command with the forward slashes:
```bash
Run the following to enter into the virtualenv:

  source C:/Users/qedk/python-zulip-api/zulip-api-py3-venv/Scripts/activate
```

### DIFF
--- a/tools/provision
+++ b/tools/provision
@@ -116,6 +116,9 @@ the Python version this command is executed with."""
                                     venv_dir,
                                     venv_exec_dir,
                                     'activate')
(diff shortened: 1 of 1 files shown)


showell:
Merged, thanks @QEDK!

PR #627, opened 2020-10-01
zulip: Give quicker feedback when credentials are wrong

### LINKED ISSUE #533: api: Give quicker feedback if your credentials are wrong

QEDK:
Fixes #533

### DIFF
--- a/zulip/integrations/bridge_between_zulips/run-interrealm-bridge
+++ b/zulip/integrations/bridge_between_zulips/run-interrealm-bridge
@@ -5,10 +5,13 @@ import os
 import argparse
 import multiprocessing as mp
 import zulip
+import logging
 import interrealm_bridge_config
 
 from typing import Any, Callable, Dict
 
+logging.basicConfig()
+log = logging.getLogger("zulip-interrealm-bridge")
 
 def create_pipe_event(to_client: zulip.Client, from_bot: Dict[str, Any],
                       to_bot: Dict[str, Any], stream_wide: bool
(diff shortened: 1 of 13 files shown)


QEDK:
@zulipbot add "area: integrations"

QEDK:
@zulipbot add "needs review"

QEDK:
@showell @neiljp I'm not sure if I'm not using zulipbot right but I cannot add labels to mark this for review so I'm pinging both of you as both of you reviewed code in the other related pull request.

andersk:
This causes many errors to be logged and then ignored (with `sys.exit()`, `return 0`, or falling off the end of an `except` block), and it doesn’t change the `Client` constructor at all. I’m pretty sure this is not what the author of #533 had in mind.

QEDK:
> This causes many errors to be logged and then ignored (with `sys.exit()`, `return 0`, or falling off the end of an `except` block), and it doesn’t change the `Client` constructor at all. I’m pretty sure this is not what the author of #533 had in mind.

This is modeled after #592 which is mostly what they wanted in #533 (per comments on that pull). I fixed the structuring, typos and rebasing.

andersk:
#592 has the same problems. One of them was already [mentioned there](https://github.com/zulip/python-zulip-api/pull/592#issuecomment-619693314), and I’ve commented there about the other.

QEDK:
> #592 has the same problems. One of them was already [mentioned there](https://github.com/zulip/python-zulip-api/pull/592#issuecomment-619693314), and I’ve commented there about the other.

I'll open a new PR to address it in the other way.

PR #633, opened 2020-11-08
zulip-botserver: Allow passing config via environment variables

### LINKED ISSUE #485: Allow passing config via environment variables to zulip-botserver
**Component:** `zulip-botserver`
**Package:** https://pypi.org/project/zulip-botserver/
**Code:** https://github.com/zulip/python-zulip-api/tree/master/zulip_botserver

Instead of requiring that config be passed as a file with the `--config-file` option, it would be great if we could pass config via environment variables.
This would make our secure key storage system much easier, as right now we use env variables to pass down secure keys to containers, and creating a config file to store them on disk is suboptimal.

```ini
[helloworld]
email=foo-bot@hostname
key=dOHHlyqgpt5g0tVuVl6NHxDLlc9eFRX4
site=http://hostname
token=aQVQmSd6j6IHphJ9m1jhgHdbnhl5ZcsY
```

Becomes:

```bash
env BOT_EMAIL=foo-bot@hostname BOT_KEY=abc... BOT_SITE=https://hostname BOT_TOKEN=abc... zulip-botserver
```

I would require adding ~20 LOC to https://github.com/zulip/python-zulip-api/blob/master/zulip_botserver/zulip_botserver/server.py, if this proposal sounds reasonable, I can probably submit a PR for it myself without too much difficulty.

LoopThrough-i-j:
Fixes #485

### DIFF
--- a/zulip_botserver/tests/test_server.py
+++ b/zulip_botserver/tests/test_server.py
@@ -4,6 +4,7 @@ from typing import Any, Dict
 import unittest
 from .server_test_lib import BotServerTestCase
 import json
+from collections import OrderedDict
 from importlib import import_module
 from types import ModuleType
 
@@ -131,6 +132,34 @@ class BotServerTests(BotServerTestCase):
         assert opts.bot_config_file is None
         assert opts.hostname == '127.0.0.1'
         assert opts.port == 5002
+
+    def test_read_config_from_env_vars(self) -> None:
+        # We use an OrderedDict so that the order of the entries in
+        # the stringified environment variable is standard even on
+        # Python 3.7 and earlier.
+        bots_config = OrderedDict()
+        bots_config['hello_world'] = {
+            'email': 'helloworld-bot@zulip.com',
+            'key': 'value',
+            'site': 'http://localhost',
+            'token': 'abcd1234',
+        }
+        bots_config['giphy'] = {
+            'email': 'giphy-bot@zulip.com',
+            'key': 'value2',
+            'site': 'http://localhost',
+            'token': 'abcd1234',
+        }
+        os.environ['ZULIP_BOTSERVER_CONFIG'] = json.dumps(bots_config)
+
+        # No bot specified; should read all bot configs
+        assert server.read_config_from_env_vars() == bots_config
+
+        # Specified bot exists; should read only that section.
+        assert server.read_config_from_env_vars("giphy") == {'giphy': bots_config['giphy']}
+
+        # Specified bot doesn't exist; should read the first section of the config.
+        assert server.read_config_from_env_vars("redefined_bot") == {'redefined_bot': bots_config['hello_world']}
 
     def test_read_config_file(self) -> None:
         with self.assertRaises(IOError):
--- a/zulip_botserver/zulip_botserver/input_parameters.py
+++ b/zulip_botserver/zulip_botserver/input_parameters.py
@@ -7,12 +7,18 @@ def parse_args() -> argparse.Namespace:
     '''
 
     parser = argparse.ArgumentParser(usage=usage)
-    parser.add_argument(
+    mutually_exclusive_args = parser.add_mutually_exclusive_group(required=True)
+    # config-file or use-env-vars made mutually exclusive to prevent conflicts
+    mutually_exclusive_args.add_argument(
         '--config-file', '-c',
         action='store',
-        required=True,
         help='Config file for the Botserver. Use your `botserverrc` for multiple bots or'
              '`zuliprc` for a single bot.'
+    )
+    mutually_exclusive_args.add_argument(
+        '--use-env-vars', '-e',
+        action='store_true',
+        help='Load configuration from JSON in ZULIP_BOTSERVER_CONFIG environment variable.'
     )
     parser.add_argument(
         '--bot-config-file',
--- a/zulip_botserver/zulip_botserver/server.py
+++ b/zulip_botserver/zulip_botserver/server.py
@@ -7,6 +7,7 @@ import os
 import sys
 import importlib.util
 
+from collections import OrderedDict
 from configparser import MissingSectionHeaderError, NoOptionError
 from flask import Flask, request
 from importlib import import_module
@@ -28,6 +29,32 @@ def read_config_section(parser: configparser.ConfigParser, section: str) -> Dict
     }
     return section_info
 
+def read_config_from_env_vars(bot_name: Optional[str] = None) -> Dict[str, Dict[str, str]]:
+    bots_config = {}  # type: Dict[str, Dict[str, str]]
+    json_config = os.environ.get('ZULIP_BOTSERVER_CONFIG')
+
+    if json_config is None:
+        raise OSError("Could not read environment variable 'ZULIP_BOTSERVER_CONFIG': Variable not set.")
+
+    # Load JSON-formatted environment variable; use OrderedDict to
+    # preserve ordering on Python 3.6 and below.
+    env_config = json.loads(json_config, object_pairs_hook=OrderedDict)
+    if bot_name is not None:
+        if bot_name in env_config:
+            bots_config[bot_name] = env_config[bot_name]
+        else:
+            # If the bot name provided via the command line does not
+            # exist in the configuration provided via the environment
+            # variable, use the first bot in the environment variable,
+            # with name updated to match, along with a warning.
+            first_bot_name = list(env_config.keys())[0]
+            bots_config[bot_name] = env_config[first_bot_name]
+            logging.warning(
+                "First bot name in the config list was changed from '{}' to '{}'".format(first_bot_name, bot_name)
+            )
+    else:
+        bots_config = dict(env_config)
+    return bots_config
 
 def read_config_file(config_file_path: str, bot_name: Optional[str] = None) -> Dict[str, Dict[str, str]]:
     parser = parse_config_file(config_file_path)
@@ -178,16 +205,20 @@ def handle_bot() -> str:
 def main() -> None:
     options = parse_args()
     global bots_config
-    try:
-        bots_config = read_config_file(options.config_file, options.bot_name)
-    except MissingSectionHeaderError:
-        sys.exit("Error: Your Botserver config file `{0}` contains an empty section header!\n"
-                 "You need to write the names of the bots you want to run in the "
-                 "section headers of `{0}`.".format(options.config_file))
-    except NoOptionError as e:
-        sys.exit("Error: Your Botserver config file `{0}` has a missing option `{1}` in section `{2}`!\n"
-                 "You need to add option `{1}` with appropriate value in section `{2}` of `{0}`"
-                 .format(options.config_file, e.option, e.section))
+
+    if options.use_env_vars:
+        bots_config = read_config_from_env_vars(options.bot_name)
+    elif options.config_file:
+        try:
+            bots_config = read_config_file(options.config_file, options.bot_name)
+        except MissingSectionHeaderError:
+            sys.exit("Error: Your Botserver config file `{0}` contains an empty section header!\n"
+                     "You need to write the names of the bots you want to run in the "
+                     "section headers of `{0}`.".format(options.config_file))
+        except NoOptionError as e:
+            sys.exit("Error: Your Botserver config file `{0}` has a missing option `{1}` in section `{2}`!\n"
+                     "You need to add option `{1}` with appropriate value in section `{2}` of `{0}`"
+                     .format(options.config_file, e.option, e.section))
 
     available_bots = list(bots_config.keys())
     bots_lib_modules = load_lib_modules(available_bots)
--- a/zulip_botserver/README.md
+++ b/zulip_botserver/README.md
@@ -1,16 +1,29 @@
 ```
 zulip-botserver --config-file <path to botserverrc> --hostname <address> --port <port>
-
 ```
 
 Example: `zulip-botserver --config-file ~/botserverrc`
 
 This program loads the bot configurations from the
-config file (botserverrc here) and loads the bot modules.
+config file (`botserverrc`, here) and loads the bot modules.
 It then starts the server and fetches the requests to the
 above loaded modules and returns the success/failure result.
 
-Please make sure you have a current botserverrc file with the
-configurations of the required bots.
-Hostname and Port are optional arguments. Default hostname is
-127.0.0.1 and default port is 5002.
+The `--hostname` and `--port` arguments are optional, and default to
+127.0.0.1 and 5002 respectively.
+
+The format for a configuration file is:
+
+    [helloworld]
+    key=value
+    email=helloworld-bot@zulip.com
+    site=http://localhost
+    token=abcd1234
+
+Is passed `--use-env-vars` instead of `--config-file`, the
+configuration can instead be provided via the `ZULIP_BOTSERVER_CONFIG`
+environment variable.  This should be a JSON-formatted dictionary of
+bot names to dictionary of their configuration; for example:
+
+    ZULIP_BOTSERVER_CONFIG='{"helloworld":{"email":"helloworld-bot@zulip.com","key":"value","site":"http://localhost","token":"abcd1234"}}' \
+      zulip-botserver --use-env-vars


LoopThrough-i-j:
@timabbott, or @showell  can you please take a look at this PR

showell:
@LoopThrough-i-j Thanks for working on this!  I appreciate that you took the time to write tests here.

I am not sure having the parallel comma-separated lists of values here is ideal.  I wonder if there's a better way to represent hierarchical data in environment variables.

I think we need to wait and see what Tim says here.  (He and his wife just welcomed a baby into the world, so he's gonna be pretty busy for the next couple weeks, so apologies in advance if this doesn't get reviewed for a while.)

LoopThrough-i-j:
Not a problem, thanks for the reply.

timabbott:
I agree with the concern about comma-separated values in environment variables.

I'm not convinced that it's worth supporting multiple bots with this environment variables model; if we did, probably it'd be better to just have each bot under a prefix (E.g. `GIPHY_ZULIP_BOT_EMAIL` or something like that) and pass which prefixes to look for on the command line.

@pirate FYI for thoughts on that question.

pirate:
I don't mind CSV strings personally, but a reasonable alternative is requiring proper json strings and just using `json.loads` (like `ZULIP_CUSTOM_SETTINGS`).

LoopThrough-i-j:
Thanks for the reply, I agree with JSON strings as well, will update the pr soon.

LoopThrough-i-j:
@pirate what do you think. Does it look fine🙂?

zulip_botserver/zulip_botserver/input_parameters.py
```diff
@@ -40,4 +39,9 @@ def parse_args() -> argparse.Namespace:
         type=int,
         help='Port on which you want to run the Botserver. (default: %(default)d)'
     )
+    parser.add_argument(
+        '--use-env-vars', '-ev',
```
pirate, line 43: `'-ev'` the short flag with two letters seems unusual to me, could it be just `-e` or maybe `-n`? if those are already taken maybe we can just remove the short flag, the verbose one is not painfully long.

pirate (approved):

pirate:
looks awesome, thanks for this work @LoopThrough-i-j!

zulip_botserver/zulip_botserver/input_parameters.py
LoopThrough-i-j: Okay sure ```-e``` is fine, will make it ```-e```

LoopThrough-i-j:
> looks awesome, thanks for this work @LoopThrough-i-j!

No problem looking forward to work more.

zulip_botserver/zulip_botserver/server.py
```diff
@@ -178,8 +212,16 @@ def handle_bot() -> str:
...
     global bots_config
+
     try:
-        bots_config = read_config_file(options.config_file, options.bot_name)
+        if options.use_env_vars:
+            bots_config = read_config_from_env_vars(options.bot_name)
+        elif options.config_file:
+            bots_config = read_config_file(options.config_file, options.bot_name)
+        else:
+            sys.exit("""Configuration Error: Please pass --use-env-vars or -ev along with properly formatted environment varibales
```
pirate, line 222:
```suggestion
            sys.exit("""Configuration Error: Please pass --use-env-vars or -e along with properly formatted environment varibales
```

pirate (approved):

zulip_botserver/zulip_botserver/server.py
LoopThrough-i-j: Thanks for this probably I was a bit overexcited for my 1st PR at Zulip

LoopThrough-i-j:
@timabbott can you please merge this if it is done.

showell:
I haven't tested this or read it thoroughly, but the overall approach seems reasonable now.  Is there a place in the readme that we should be updating too?

LoopThrough-i-j:
> I haven't tested this or read it thoroughly, but the overall approach seems reasonable now. Is there a place in the readme that we should be updating too?

The readme does not talk about the new feature, of passing config via environment variables. Then the docs will need an update as well.

zulip_botserver/zulip_botserver/server.py
```diff
@@ -28,6 +29,39 @@ def read_config_section(parser: configparser.ConfigParser, section: str) -> Dict
...
     return section_info
 
+def read_config_from_env_vars(bot_name: Optional[str] = None) -> Dict[str, Dict[str, str]]:
+    bots_config = {}  # type: Dict[str, Dict[str, str]]
+    env_config = OrderedDict()   # type: Dict[str, Dict[str, str]]
+
+    # Check for presence of environment variables
+    # otherwise throws appropriate errors
+    if os.environ.get('ZULIP_BOTSERVER_CONFIG') is None:
+        raise OSError("Could not read environment varibale 'ZULIP_BOTS': Variable not set.")
```
alexmv, line 39: `varibale` -> `variable`

And the variable looks to be named `ZULIP_BOTSERVER_CONFIG ` now, not `ZULIP_BOTS`

zulip_botserver/zulip_botserver/server.py
```diff
@@ -28,6 +29,39 @@ def read_config_section(parser: configparser.ConfigParser, section: str) -> Dict
...
+
+    # The bots emails keys sites and tokens
+    # are stored in their respective lists
```
alexmv, line 42: I don't think this comment is accurate anymore?

zulip_botserver/zulip_botserver/server.py
```diff
@@ -28,6 +29,39 @@ def read_config_section(parser: configparser.ConfigParser, section: str) -> Dict
...
+    env_config = json.loads(str(os.environ.get('ZULIP_BOTSERVER_CONFIG')), object_pairs_hook=OrderedDict)
```
alexmv, line 43: Do we need the `str` here?

zulip_botserver/zulip_botserver/server.py
```diff
@@ -28,6 +29,39 @@ def read_config_section(parser: configparser.ConfigParser, section: str) -> Dict
...
+
+    # If bot name is provided we only use
+    # the provided name and return it's config
+    if bot_name is not None:
+        if bot_name in env_config:
+            bots_config[bot_name] = env_config[bot_name]
+        else:
+            # If bot_config doesn't match
+            # If the bot name doesnot match
```
alexmv, line 52: These lines look like they're duplicates.

zulip_botserver/zulip_botserver/server.py
```diff
@@ -28,6 +29,39 @@ def read_config_section(parser: configparser.ConfigParser, section: str) -> Dict
...
+            # name of the first bot is replaces the bot name
+            bots_config[bot_name] = env_config[list(env_config.keys())[0]]
```
alexmv, line 54: Can we pull `list(env_config.keys())[0]` out into a variable, since we're using it a couple times, and it's a bit of a mouthful?

zulip_botserver/zulip_botserver/server.py
```diff
@@ -178,8 +212,16 @@ def handle_bot() -> str:
...
+
     try:
-        bots_config = read_config_file(options.config_file, options.bot_name)
+        if options.use_env_vars:
+            bots_config = read_config_from_env_vars(options.bot_name)
+        elif options.config_file:
+            bots_config = read_config_file(options.config_file, options.bot_name)
+        else:
+            sys.exit("""Configuration Error: Please pass --use-env-vars or -e along with properly formatted environment varibales
+                     Or pass the path to a properly formatted config file with --config-file path/to/config/file""")
```
alexmv, line 223: We can do this a little more cleanly by making the `--config-file` and `--use-env-vars` options [be mutually exclusive and required](https://docs.python.org/3/library/argparse.html#mutual-exclusion).

zulip_botserver/zulip_botserver/server.py
```diff
@@ -28,6 +29,39 @@ def read_config_section(parser: configparser.ConfigParser, section: str) -> Dict
...
+    # otherwise throws appropriate errors
+    if os.environ.get('ZULIP_BOTSERVER_CONFIG') is None:
+        raise OSError("Could not read environment varibale 'ZULIP_BOTS': Variable not set.")
+
+    # The bots emails keys sites and tokens
+    # are stored in their respective lists
+    env_config = json.loads(str(os.environ.get('ZULIP_BOTSERVER_CONFIG')), object_pairs_hook=OrderedDict)
+
+    # If bot name is provided we only use
+    # the provided name and return it's config
```
alexmv, line 46: `return its config`, not `return it's config`, in this case.

I also notice your comment lines are wrapped fairly short; you should wrap them at 80-ish characters.

zulip_botserver/zulip_botserver/input_parameters.py
```diff
@@ -40,4 +39,9 @@ def parse_args() -> argparse.Namespace:
         type=int,
         help='Port on which you want to run the Botserver. (default: %(default)d)'
     )
+    parser.add_argument(
+        '--use-env-vars', '-e',
+        action='store_true',
+        help='Optional argument to parse values through environment variables. (default: %(default)d)'
```
alexmv, line 45: Can we move this up to be under `--config-file`?  It is also now only one environment variable, not more than one, so we should name it here.

Finally, see below for a how to deal more cleanly with `--config-file`.

alexmv (changes requested):
Thanks for this work!  I left some comments about a couple things to iterate on.

Having a JSON dump in an environment variable doesn't thrill me, because it's not really all that easy to set or edit by hand, but it does solve the problem of passing in a somewhat-complex datastructure.

Please update the commit message to say "environment variable" and not "environment variables"; it would be nice to have a mention of JSON in there as well.

One thing that this doesn't address, that would be an improvement in my eyes, is combining the config file with the environment variables.  That is, let me define all but my token inside the configuration file, and pass in just that via a JSON dict, which then get flattened together.

That adds a bit more complexity, though, and I think this approach is fine for a first pass.

zulip_botserver/zulip_botserver/server.py
LoopThrough-i-j: Yes because lint throws an error, as an environment variable can be None: resulting in type mismatch.
Though I have a check for None before, lint does not understand that.

LoopThrough-i-j:
@alexmv Thanks a lot for the suggestions. Can you please take a look if everything looks good? 

The checks fail because in python 3.5 a random bot is chosen to replace the name instead of the 1st one. For some reason, Ordered Dict is not working. 

Can you help to improve or shall I remove the test for checking the 1st python bot is replaced while using the environment variables?

Thanks a lot in advance.

zulip_botserver/zulip_botserver/input_parameters.py
```diff
@@ -7,13 +7,19 @@ def parse_args() -> argparse.Namespace:
...
         '--config-file', '-c',
         action='store',
-        required=True,
         help='Config file for the Botserver. Use your `botserverrc` for multiple bots or'
              '`zuliprc` for a single bot.'
     )
+    mutually_exclusive_args.add_argument(
+        '--use-env-vars', '-e',
+        action='store_true',
+        help='Optional argument to parse values through environment variables. (default: %(default)d)'
```
alexmv, line 21: We should make explicit here which environment variable this is:
```suggestion
        help='Load configuration from JSON in ZULIP_BOTSERVER_CONFIG environment variable.'
```

zulip_botserver/zulip_botserver/server.py
alexmv: If you pull this out to its own variable, and use it in the `if` as well as here, then mypy will propagate that `is None` constraint.  It'll repeat the name of the environment variable less (for fewer potential copy/paste mistakes in the future), and also have fewer arbitrary variable casts!

zulip_botserver/zulip_botserver/server.py
```diff
@@ -28,6 +29,33 @@ def read_config_section(parser: configparser.ConfigParser, section: str) -> Dict
...
+        raise OSError("Could not read environment variable 'ZULIP_BOTSERVER_CONFIG': Variable not set.")
+
+    # Load JSON formatted Environment Variable
+    env_config = json.loads(str(os.environ.get('ZULIP_BOTSERVER_CONFIG')), object_pairs_hook=OrderedDict)
+    # If bot name is provided we only use the provided name and return the config
+    if bot_name is not None:
+        if bot_name in env_config:
+            bots_config[bot_name] = env_config[bot_name]
+        else:
+            # If the bot name doesnot match, the bot name replaces the name of the first bot
```
alexmv, line 47:
```suggestion
            # If the bot name does not match, the bot name replaces the name of the first bot
```

zulip_botserver/zulip_botserver/server.py
alexmv: We can't hit this `else` case anymore, because the mutually-exclusive-but-required options we now configured above.

zulip_botserver/zulip_botserver/server.py
```diff
@@ -179,7 +207,13 @@ def main() -> None:
...
+        if options.use_env_vars:
+            bots_config = read_config_from_env_vars(options.bot_name)
+        elif options.config_file:
+            bots_config = read_config_file(options.config_file, options.bot_name)
+        else:
+            sys.exit("""Configuration Error: Please pass --use-env-vars or -e along with properly formatted environment varibales
+                     Or pass the path to a properly formatted config file with --config-file path/to/config/file""")
     except MissingSectionHeaderError:
         sys.exit("Error: Your Botserver config file `{0}` contains an empty section header!\n"
                  "You need to write the names of the bots you want to run in the "
```
alexmv, line 219: It occurs to me on re-read hat this `except` (and the one below it) are specific to the `read_config_file` branch only; we can't raise them from `read_config_from_env_vars`.  We should move the `try .. except` from _outside_ the `if` to _inside` the `elif options.config_file:` block.

zulip_botserver/tests/test_server.py
```diff
@@ -132,6 +132,64 @@ def test_argument_parsing_defaults(self) -> None:
...
+                'token': 'abcd1234',
+            },
+            'giphy': {
+                'email': 'giphy-bot@zulip.com',
+                'key': 'value2',
+                'site': 'http://localhost',
+                'token': 'abcd1234',
+            }
+        }
+        assert json.dumps(bot_conf1, sort_keys=True) == json.dumps(expected_config1, sort_keys=True)
```
alexmv, line 167: We don't need to serialize these to test equality -- we can just `assert bot_conf1 == expected_config1`

alexmv (commented):
> The checks fail because in python 3.5 a random bot is chosen to replace the name instead of the 1st one. For some reason, Ordered Dict is not working.

Starting in Python 3.6, standard dicts keep their order (though that's not a thing one should rely on until 3.7).  That means that yeah, we have a dict someplace we're expecting to have an OrderedDict.

> Can you help to improve or shall I remove the test for checking the 1st python bot is replaced while using the environment variables?

We shouldn't remove the test -- it's doing real work, telling us that the behavior on Python 3.5 is unreliable!  Can you do some more debugging and figure out what might be the cause of that?

zulip_botserver/tests/test_server.py
LoopThrough-i-j: Can this stay as it is for consistency?

alexmv: Consistency with what?

alexmv: Oh, I see that `test_read_config_file` also serializes to JSON.  I'd rather fix that, than perpetuate the unnecessary code.

zulip_botserver/zulip_botserver/server.py
```diff
@@ -28,6 +29,32 @@ def read_config_section(parser: configparser.ConfigParser, section: str) -> Dict
     }
     return section_info
 
+def read_config_from_env_vars(bot_name: Optional[str] = None) -> Dict[str, Dict[str, str]]:
+    bots_config = {}  # type: Dict[str, Dict[str, str]]
+    env_config = OrderedDict()   # type: Dict[str, Dict[str, str]]
+    json_config = os.environ.get('ZULIP_BOTSERVER_CONFIG')   # type: Optional[str]
```
alexmv, line 35: mypy can deduce the type of this, so we don't need the `# type:` comment here -- and I don't think it it adds much.

zulip_botserver/zulip_botserver/server.py
```diff
@@ -28,6 +29,32 @@ def read_config_section(parser: configparser.ConfigParser, section: str) -> Dict
...
+
+    # Check for presence of environment variable otherwise throws appropriate errors
```
alexmv, line 37: I think we can drop this comment.  Comments are good for explaining the "why" of something, or for particularly gnarly algorithms.  This is neither.

alexmv (commented):
This looks pretty good!  Just a few minor cleanups.

Did you have any luck tracking down the Python 3.5 problem?

LoopThrough-i-j:
> This looks pretty good! Just a few minor cleanups.
> 
> Did you have any luck tracking down the Python 3.5 problem?

I tried a lot of things out even printing down types and trying to find out leakages, but nothing worked(Stackoverflow couldn't help as well). Later, I decided to drop the test for python3.5 and the test happens only for python3.6 or higher.

LoopThrough-i-j:
@alexmv what do you think

ryanreh99:
> The checks fail because in python 3.5 a random bot is chosen to replace the name instead of the 1st one. For some reason, Ordered Dict is not working.

I didn't run this locally but I think this maybe the reason:-
In the docs for [json.load](https://docs.python.org/3/library/json.html#json.load) it is mentioned: `All optional parameters are now keyword-only.`
And in the [OrderedDict](https://docs.python.org/3/library/json.html#json.load) docs, it is mentioned:
```
Changed in version 3.6: With the acceptance of PEP 468, order is retained for keyword arguments passed to the OrderedDict constructor and its update() method.
```

LoopThrough-i-j:
@timabbott can you please take a look at this PR

alexmv:
Using `object_pairs_hook=list` to preserve the order is an interesting solution, that nicely steps around the OrderedDict problem.  Since it's not necessarily completely intuitively obvious, I think it probably deserves extending the comment above it about "the first bot" that we re-parse to extract key order, in order to support Python 3.6 and before, since 3.7 and later would preserve key ordering information.

Otherwise, this is looks in great shape -- sorry for the long delay in the review!

LoopThrough-i-j:
> Using `object_pairs_hook=list` to preserve the order is an interesting solution, that nicely steps around the OrderedDict problem. Since it's not necessarily completely intuitively obvious, I think it probably deserves extending the comment above it about "the first bot" that we re-parse to extract key order, in order to support Python 3.6 and before, since 3.7 and later would preserve key ordering information.

sure will fix the issue and will be working on my next PR! btw the tests fail for now, will it be fine I take the responsibility to migrate it to Github actions.

alexmv:
> btw the tests fail for now, will it be fine

If the tests are still failing, that means the workaround of using `list` _isn't working_ -- so no, that's not fine.

I installed python 3.5 myself to investigate, by putting some print statements into the test and running it several times.  What jumped out is that the order of the `list` was still changing!  So I added a print statement for the environment variable itself that was being fed into the test.  Which showed that the problem isn't actually that the OrderedDict trick doesn't work during loading on 3.5 -- it's that _we're using `json.dumps` to construct the JSON we put in the environment,_ which is _unordered on Python 3.5_, so we're feeding the program unreliable input!

The fix is to swap back to the much less hacky OrderedDict, and _also_ to use that in the tests to make sure we pass in a consistent value for the environment variable.  With that in place, tests pass consistently on 3.5, and later.  I've pushed a commit that does this, as well as add some documentation.

> will it be fine

Tests should always pass.  In general, it's not a good idea to break functionality, even if you're planning on coming back and to fix it later -- particularly not when just adding new features, like this.  If tests are failing when someone arrives at a project, folks will assume that _any_ failure is probably not due to their commit, and thus failures (including real important ones!) will accumulate over time.  Keeping confidence that tests don't lie is _really_ important.

>  I take the responsibility to migrate it to Github actions.

I'd love to take you up on migrating to Github Actions, however!   But let's do that in a separate PR.

LoopThrough-i-j:
Thanks a lot for your help and detailed comment and for mentoring me throughout the PR. I would love to work on the migration of tests to GitHub actions. Thank you once more for helping me to become a better developer.

PR #634, opened 2020-11-12
list_subscriptions: Encoded list_subscriptions parameters.

aryanshridhar:
Added <code>default_parameter</code> within <code><a href = 'https://zulip.com/api/get-subscriptions#get-subscribed-streams'>list_subscriptions</a></code> api call and deprecating the dict parameter .

<strong>Earlier :</strong>
<pre>
client.list_subscriptions({include_subscribers=True})
</pre>
<strong>Now :</strong>
<pre>
client.list_subscriptions(include_subscribers=True)
</pre>
As discussed within zulip/zulip#16698


andersk:
Deprecating is not the same as removing—we still want to support the old syntax with a warning for some deprecation period.

Also, this is not the only function affected by this issue.

aryanshridhar:
Hey @andersk , upon <code> grep -rnw . -e 'list_subscriptions'</code> , I see the following functions calling the <code>list_subscriptions</code> function : 
<pre>
/integrations/jabber/jabber_mirror_backend.py:278:        stream_infos = get_stream_infos("subscriptions", zulipToJabber.client.list_subscriptions)
./integrations/bridge_with_irc/irc_mirror_backend.py:39:        resp = self.zulip_client.list_subscriptions()
./build/lib/integrations/jabber/jabber_mirror_backend.py:278:        stream_infos = get_stream_infos("subscriptions", zulipToJabber.client.list_subscriptions)
./build/lib/integrations/bridge_with_irc/irc_mirror_backend.py:39:        resp = self.zulip_client.list_subscriptions()
./build/lib/zulip/examples/list-subscriptions:21:print(client.list_subscriptions())
./build/lib/zulip/__init__.py:1281:    def list_subscriptions(self, request: Optional[Dict[str, Any]] = None, **kwargs: Any) -> Dict[str, Any]:
./zulip/examples/list-subscriptions:21:print(client.list_subscriptions())
./zulip/__init__.py:1281:    def list_subscriptions(self, request: Optional[Dict[str, Any]] = None, **kwargs: Any) -> Dict[str, Any]:
</pre>

But none of them have a parameter of <code>{'include_subscribers' : bool_value}</code> . Could you guide me on mentioning the functions that need to be altered due to this issue .

zulip/zulip/__init__.py
```diff
@@ -1278,14 +1278,17 @@ def remove_alert_words(self, alert_words: List[str]) -> Dict[str, Any]:
             }
         )
 
-    def list_subscriptions(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
+    def list_subscriptions(self, request: Optional[Dict[str, Any]] = None, **kwargs: Any) -> Dict[str, Any]:
         '''
             See examples/list-subscriptions for example usage.
         '''
+        if request is not None:
+            logging.warning('')
```
aryanshridhar, line 1286: Does <code>include_subscribers as a dictionary parameter will be deprecated in future versions .</code> looks like a fine warning ? Am open to other suggestions as well .

andersk:
These all share the same antipattern:

```python
    def get_profile(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
    def update_presence(self, request: Dict[str, Any]) -> Dict[str, Any]:
    def get_users(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
    def get_members(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
    def list_subscriptions(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
    def mute_topic(self, request: Dict[str, Any]) -> Dict[str, Any]:
    def render_message(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
    def create_user(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
    def update_storage(self, request: Dict[str, Any]) -> Dict[str, Any]:
    def get_storage(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
    def set_typing_status(self, request: Dict[str, Any]) -> Dict[str, Any]:
```

aryanshridhar:
Is there any changes to be made @andersk ? Or should I go ahead to fix failing <code>mypy</code> failures .

zulip/zulip/__init__.py
```diff
@@ -1072,17 +1072,20 @@ def deregister(self, queue_id: str, timeout: Optional[float] = None) -> Dict[str
...
-    def get_profile(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
+    def get_profile(self, request: Optional[Dict[str, Any]] = None, **kwargs: Any) -> Dict[str, Any]:
         '''
             Example usage:
 
             >>> client.get_profile()
             {u'user_id': 5, u'full_name': u'Iago', u'short_name': u'iago', ...}
         '''
+        if request is not None:
+            logging.warning('Passing parameters as a dictionary will be deprecated in future versions.')
```
neiljp, line 1083: Due to the duplication I would extract this as a very simple function with a descriptive name., or at minimum the string.

Regarding the message, aren't we referring to particularly *request* parameters in a dictionary being deprecated?

zulip/zulip/__init__.py
```diff
@@ -1072,17 +1072,20 @@ def deregister(self, queue_id: str, timeout: Optional[float] = None) -> Dict[str
...
+
         return self.call_endpoint(
             url='users/me',
             method='GET',
-            request=request,
+            request=request or kwargs,
```
neiljp, line 1088: This ignores `kwargs` if `request` is passed, which may be confusing and doesn't follow the logic @andersk and I discussed in the issue.

neiljp (commented):
@aryanshridhar I left some feedback on what you have so far.

neiljp:
@andersk Would a decorator to handle the migration work here, do you think?

zulip/zulip/__init__.py
andersk: “Will be removed”, not “will be deprecated”. It’s deprecated now (which [means](https://en.wikipedia.org/wiki/Deprecation#Software) that it still exists for compatibility, but users should stop using it); it will be removed in the future.

andersk:
A decorator might have the disadvantage of making the pydoc confusing and the mypy types poor (all mypy can guarantee about a decorator is basically “congratulations, you have *a function*!”).

A helper function pattern like this would be fine:

```python
def deprecated_request(request: Optional[Dict[str, Any]]) -> Dict[str, Any]:
    if request is not None:
        logger.warning(…)
        return request
    return {}

def list_subscriptions(self, request: Optional[Dict[str, Any]] = None, **kwargs: Any) -> Dict[str, Any]:
    return self.call_endpoint(…, request={**deprecated_request(request), **kwargs})
```

neiljp:
@andersk I agree that this approach is simple - and almost suggested something similar, ie. putting all of the deprecation-migration code together.

Re decorators, I did think mypy could handle them - since we have them annotated in the main repo - though I suppose it's one thing to annotate and another to know what is actually validated via them being present. I'd also have thought `@wraps` would handle docstrings, and would therefore take it that docstring-consumers like pydoc would be OK too.

zulip/zulip/__init__.py
aryanshridhar: Oh Sorry , My bad .

aryanshridhar:
Hey , @andersk @neiljp , Could you review it again ?

timabbott:
I think this change makes sense to me -- @aryanshridhar can you rebase this?

I'm a bit reluctant to merge this while we're unable to update `python-zulip-api` in zulip/zulip, which we'll definitely need to do in order to update documentation for the modified endpoints. 

(Or did we resolve that issue and I just didn't notice?)

aryanshridhar:
Rebased it @timabbott, As far as I remember, there was no further work done related to this pr in zulip/zulip. 
Opened zulip/zulip#17804 to update the docs.

timabbott:
I'm going to hold off on this until after https://github.com/zulip/python-zulip-api/pull/692; let's chat about it next week.

timabbott:
@aryanshridhar can you rebase this?

PR #635, opened 2020-11-19
Following changes are done:

dmarar:
The __init__.py file of Client was changed to thrown an error during instantiation if the credentials are incorrect.
send.py was modified to handle the error thrown by the Client.

Adding an entry in the log for the invalidcredentialException raised


dmarar:
@zulipbot codereview

PR #640, opened 2020-12-21
Changes to make rss bot working with zulip 3.3 and Debian 10.7 :

fulljackz:
- Add init method
- Change return hashlib.md5 synthax

Signed-off-by: Manu LN <manu+github@lacavernedemanu.fr>

Before changes, while trying to get rss infos in RSS stream the script output was 

```
  File ".local/share/zulip/integrations/rss/rss-bot", line 190, in <module>
    entry_hash = compute_entry_hash(entry)  # type: str
  File ".local/share/zulip/integrations/rss/rss-bot", line 131, in compute_entry_hash
    return hashlib.md5(entry_id + str(entry_time)).hexdigest()
TypeError: Unicode-objects must be encoded before hashing
```

### DIFF
--- a/zulip/integrations/rss/rss-bot
+++ b/zulip/integrations/rss/rss-bot
@@ -111,6 +111,7 @@ def log_error_and_exit(error: str) -> None:
 
 class MLStripper(HTMLParser):
     def __init__(self) -> None:
+        super().__init__()
         self.reset()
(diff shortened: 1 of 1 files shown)


timabbott:
Merged, thanks for the PR @fulljackz!

PR #643, opened 2021-01-07
Setup CI using Github Actions

LoopThrough-i-j:


.github/workflows/zulip-tests.yml
```diff
@@ -0,0 +1,36 @@
...
+        sudo apt-get update
+        sudo apt-get install virtualenv
+        tools/provision --force
+
+    - name: Running Test-Suite
+      run: |
+        source zulip-api-py3-venv/bin/activate
+        tools/test-main
+
+    - name: Run Codecov
```
rht, line 32: You should use the standard action for this
```yml
    - name: Codecov
      uses: codecov/codecov-action@v1.2.1
```

.github/workflows/zulip-tests.yml
```diff
@@ -0,0 +1,36 @@
+name: Zulip Tests
+
+on: [pull_request, push]
```
rht, line 3: This should be
```yml
on:
  push:
    branches:
    - master
  pull_request:
```

otherwise it would be wasteful.

.github/workflows/zulip-tests.yml
```diff
@@ -0,0 +1,36 @@
+name: Zulip Tests
```
rht, line 1: The name should be just "build" because this name will show up in the build badge, and we want it short, just "build", like on Travis.

rht:
@LoopThrough-i-j  add a commit to remove .travis.yml at the end.

.github/workflows/zulip-tests.yml
```diff
@@ -0,0 +1,36 @@
...
+
+on: [pull_request, push]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8]
```
rht, line 10: 3.5 is already EOL and should be removed.

rht:
@LoopThrough-i-j you haven't included the static analysis https://github.com/zulip/python-zulip-api/blob/master/.travis.yml#L9-L14.

rht:
Your commit messages should end with a fullstop ('.').

rht:
For other reviewers wanting to check the build: https://github.com/LoopThrough-i-j/python-zulip-api/actions.

.github/workflows/zulip-tests.yml
LoopThrough-i-j: okay sure

.github/workflows/zulip-tests.yml
LoopThrough-i-j: sure

.github/workflows/zulip-tests.yml
LoopThrough-i-j: I added this because I thought we might want the tests for all the branches.

LoopThrough-i-j:
> @LoopThrough-i-j you haven't included the static analysis https://github.com/zulip/python-zulip-api/blob/master/.travis.yml#L9-L14.

Thanks a lot for your review, will add the static analysis

>Your commit messages should end with a fullstop ('.'). 

I am sorry for this, probably the lint for git commit is not set.

.github/workflows/zulip-tests.yml
```diff
@@ -0,0 +1,62 @@
...
+      run: |
+        sudo rm -vf /etc/apt/sources.list.d/*
+        sudo apt-get update
+        sudo apt-get install virtualenv
+        tools/provision --force
+
+    - name: Running Test-Suite
+      run: |
+        source zulip-api-py3-venv/bin/activate
+        tools/test-static-analysis
```
rht, line 32: I think it would be more efficient if you do this inside test, but somehow use `if` ( https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsif ) to make sure it runs only on 3.6.

rht (changes requested):

.github/workflows/zulip-tests.yml
alexmv: We currently only have a `master` branch, and I'd prefer to default to running tests on new branches, rather than not.

.github/workflows/zulip-tests.yml
alexmv: We can also add 3.9 into the mix now.

.github/workflows/zulip-tests.yml
alexmv: This is a great idea -- it'll save the time of setting up python and installing deps an extra time.

.github/workflows/zulip-tests.yml
rht: The GA workflow will be replicated across all the forks. So in `on: [push, pull_request]` whenever someone pushes a new branch in their own fork, the build will be triggered in their forks. This is not Travis' behavior since CI is by default turned off, but it'd be wasteful energy consumption to the planet, even though it is MSFT's money that is being burned.

alexmv:
Thanks for doing this, @LoopThrough-i-j!  As I mentioned in the other PR, having confidence the the current state of the tests is really key, so  not having it show up on PRs is really important to get fixed -- so thank you for tackling that.  @rht's commented on more or less everything I can think of, so I don't have much to add. :)

rht:
@LoopThrough-i-j  nit: your Dropbox commit message is more concise if you just say: 'dropbox_share: Pin version of `dropbox` dependency.' and then in the secondary description of the commit message say that the version after 10.10.0 fails the test.

.github/workflows/zulip-tests.yml
neiljp: Can it generate a separate output, though? "3.6 & static checks" is not ideal, even if it does reduce setup, whereas if it could be run in the 3.6 section but generate a static test output section, that would be clearer for devs to read - particularly new ones.

.github/workflows/zulip-tests.yml
alexmv: Aha!  Thanks for teaching me that!  Given that, this makes sense to keep as-is.

neiljp: To clarify, with the version in @rht 's first comment above, would this mean Actions would not be run in users' own forks, unless their PRs happen to be on the master branch? I get the feeling this is something that devs find useful before opening PRs, though enabling the setting to build forks separately via Travis could be something easily missed for starting devs.

What about something like: https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662/10 (or similar)

rht:
> @LoopThrough-i-j you should add a commit to remove .travis.yml at the end.

You forgot to react to this.

.github/workflows/zulip-tests.yml
```diff
@@ -0,0 +1,62 @@
+name: Build
```
rht, line 1: Lowercase "build"; should look like this: https://img.shields.io/circleci/build/github/zulip/zulip/master.svg

.github/workflows/zulip-tests.yml
rht: The step is going to be named "Static analysis" so I think it will be unambiguous for devs to read the nature of the error.

.github/workflows/zulip-tests.yml
rht: > What about something like: https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662/10 (or similar)
```yml
  push:
  pull_request:
    branches:
    - '**:**'
```
seems to exclude PRs created with non-fork branches, which is not what we want? We want to test every PR, regardless of the source repo.

.github/workflows/zulip-tests.yml
LoopThrough-i-j: okay, yeah I hope that will be good.

.github/workflows/zulip-tests.yml
LoopThrough-i-j: I was wondering as well to have the static tests on every python version and run the rest only if it succeeds. @neiljp I am not sure of what you are talking about is possible, but I think having it on all the tests will make more sense.

.github/workflows/zulip-tests.yml
LoopThrough-i-j: Okay. Will make it.

LoopThrough-i-j:
> > @LoopThrough-i-j you should add a commit to remove .travis.yml at the end.
> 
> You forgot to react to this.

I didn't remove the .travis.yml to keep it for reference. I will remove it from my final commit once everyone is fine with it.

.github/workflows/zulip-tests.yml
LoopThrough-i-j: Though Having a separate Job for lint is good as well, because the jobs run in parallel and won't take more time. Rather Adding it as a step with increase the time of execution. I don't think we need to take care of its load of execution in the MS Servers.

rht: I suppose my concern was a relic of having to deal with Travis' limited resource (only 4 jobs max!). At this point it becomes a matter of whether being considerate to MSFT as not to waste their resource or not.

rht:
Typo in the Dropbox commit message: "verion".

rht:
Also, you must use imperative statement, not "fixed". I think it is more relevant to say "Pin" than "Fix".

.github/workflows/zulip-tests.yml
LoopThrough-i-j: Actually, this will help us find a separate report for the Static tests.

rht:
You forgot full stop for the first commit again.

LoopThrough-i-j:
> You forgot full stop for the first commit again.

Yeah, I added it now. I think we don't have git lint set here. Can I work on that next? @rht

rht:
> I think we don't have git lint set here. Can I work on that next?

Sounds good. [This](https://github.com/zulip/zulip/issues/1131) is the issue on zulip/zulip, and [this](https://github.com/zulip/zulip/commit/ec8e47192adddf9fc0d76eb7f7189704d1bd1ded) the commit that closed it.

rht:
Another nit: in the commit message "pin" and "setup" should be capitalized. E.g. see https://github.com/zulip/zulip/commit/42dfd9860702ed4dc450023fbf22e7bbd88ae8de.

LoopThrough-i-j:
> Another nit: in the commit message "pin" and "setup" should be capitalized. E.g. see [zulip/zulip@42dfd98](https://github.com/zulip/zulip/commit/42dfd9860702ed4dc450023fbf22e7bbd88ae8de).

@rht This is updated I hope everything looks good now. We can probably resolve the unresolved issue above and Keep the Jobs separate for separate reports.

> > I think we don't have git lint set here. Can I work on that next?
> 
> Sounds good. [This](https://github.com/zulip/zulip/issues/1131) is the issue on zulip/zulip, and [this](https://github.com/zulip/zulip/commit/ec8e47192adddf9fc0d76eb7f7189704d1bd1ded) the commit that closed it.

Yes, I will create a new PR for Git Lint Setup.

.github/workflows/zulip-tests.yml
neiljp: @rht Agreed. This does solve a problem with duplication, but not what we want here.

If I understand this correctly I'm not sure there's a way around this without some way of removing duplicated work between forks (for PRs + fork-branches). However, it would be useful for users to have the ability to enable Actions on their own forks not just for master branches (like with Travis) - without adding a custom commit to their local trees!

.github/workflows/zulip-tests.yml
neiljp: @rht I agree that it would show the Step failed as Static Analysis, but it would show overall (only) within the named python 3.6 test job, not a specific Static Analysis job. That was my point about whether the result could be saved into a separate job as simply a result.

In zulip-terminal we currently have separate jobs for each part of static testing even, to make this clearer - or hopefully so - without too much overhead. Our installation dependencies and setup for each are lower though?

.github/workflows/zulip-tests.yml
LoopThrough-i-j: I would suggest it to allow on all push and PR as the rest of the Zulip does it? Let's not think much about duplicate builds because we would obviously not want to remove user privileges unnecessarily which we provided earlier, just because we want to increase efficiency that too in an unconstrained environment.

rht: > However, it would be useful for users to have the ability to enable Actions on their own forks not just for master branches (like with Travis) - without adding a custom commit to their local trees!

When a dev working on their fork pushes a WIP to a branch, it is not always the case that they want their commits to be tested right away.
My solution to this: if a dev explicitly wants to CI-test the fork-branch once they are confident that the code has stabilized, they can just make a PR to the upstream and label it WIP. If this is not feasible, a small commit to `on: [push, pull_request` can be added easily. tl;dr: think of all the wasted builds if you do `on: [push, pull_request` by default.

.github/workflows/zulip-tests.yml
rht: I have no strong opinion on this separation of static analysis, so am fine with making it separate. Going forward, maybe it is worth it to cache pip to save time and resource during installation? See e.g. https://github.com/django-ftl/fluent-compiler/blob/a6048b856a3db0761ae05a9d3049489f6bb11cbe/.github/workflows/pythonpackage.yml#L29

.github/workflows/zulip-tests.yml
rht: See e.g. how Mypy does it: https://github.com/python/mypy/blob/master/.github/workflows/test.yml.

LoopThrough-i-j: This is a good start then, But we need to update the doc to make people aware of this.

rht: Mypy is not the only major Python-based FOSS project restricting CI build on push to master. Flask does it too: https://github.com/pallets/flask/blob/master/.github/workflows/tests.yaml#L2-L10. Several others either use CircleCI or their own CI (e.g. Google), so it's a bit hard to find more examples. My point is that please consider the wasted resource.

LoopThrough-i-j: Great, I do appreciate your thought on resource utilization. Most people consider resources as free or paid and only consider the way it is utilized if there is some restriction or payment involved, but thanks for making me realized that we are developing for the future and it's our duty not to misuse what we get for free or probably it won't last forever. Thanks a lot, @rht.

LoopThrough-i-j: @rht, @neiljp can we continue with this.

timabbott:
@LoopThrough-i-j I don't see CI as having run on this PR -- any idea why?

LoopThrough-i-j:
@timabbott because it runs on push to master
```
on:
  push:
    branches:
    - master
```
and for all pull requests
```
  pull_request:
```

The branch I committed to in my fork is not master but named CI. 
I have tested it out thoroughly to make it work, and have got all the issues fixed.

timabbott:
Merged the first commit, after copying the commit message into a comment, as e995e52896a9fa93678eedb777d6d63927913bb5.

Can you post a link to the CI run output with GitHub Actions?

LoopThrough-i-j:
> Merged the first commit, after copying the commit message into a comment, as [e995e52](https://github.com/zulip/python-zulip-api/commit/e995e52896a9fa93678eedb777d6d63927913bb5).
> 
> Can you post a link to the CI run output with GitHub Actions?

This Actions probably won't run till it is merged in the original repository. Because the Repo does not have the required build files to run it. I can run it on my form by removing master from it. Hopefull I could explain it to you.

neiljp:
@timabbott zulip/zulip has `on [push, pull_request]` in almost all workflows (except production). There is discussion above on this alternative, but I'd be interested to hear what you think since the choice in this PR is different to zulip/zulip.

rht:
> I can run it on my fork by removing master from it.

@LoopThrough-i-j alternatively you can make a PR to the master branch in your repo, so you don't have to modify any code.

rht:
FTR @andersk restricted the push to only master (https://github.com/zulip/zulip-desktop/commit/a1bb6da4fb3ffd2007b147c184f64300711094f4) at zulip/zulip-desktop.

LoopThrough-i-j:
@timabbott [here](https://github.com/LoopThrough-i-j/python-zulip-api/actions/runs/516393976) is a sample build I triggered to test if everything still works, for clarity.

.travis.yml
```diff
@@ -1,24 +0,0 @@
...
-  - "3.6"
-  - "3.7"
-  - "3.8"
-env: TEST_SUITE=test-main
-matrix:
-  include:
-  - python: "3.6"
-    env: TEST_SUITE=test-static-analysis
-    addons:
-      postgresql: "9.3"
```
neiljp, line 14: Is this relevant to the new version?

README.md
```diff
@@ -1,6 +1,7 @@
 # Zulip API
 
-[![Build status](https://travis-ci.com/zulip/python-zulip-api.svg?branch=master)](https://travis-ci.com/github/zulip/python-zulip-api)
+[![Build status](https://github.com/zulip/python-zulip-api/workflows/build/badge.svg?branch=master)]
+(https://github.com/zulip/python-zulip-api/actions?query=workflow%3Abuild)
```
neiljp, line 4: Perhaps minor, but should this link be to master too?

neiljp (commented):
@LoopThrough-i-j I've left a few comments which you could maybe address.

I think it'd be good to get this repo back running CI ASAP so having something present would be a great start which we could iterate on later - changing the on push/pull_request, for example.

.travis.yml
LoopThrough-i-j: I removed Postgres because it didn't need any Postgres instance while setup and probably uses, mock for testing those that need Postgres. (In short, all the tests run without Postgres).

README.md
LoopThrough-i-j: yes, you are right thanks for this.

alexmv:
Thank you for being persistent on this, and apologies for the delay.  I've reformatted the yml with prettier and merged.

LoopThrough-i-j:
> Thank you for being persistent on this, and apologies for the delay. I've reformatted the yml with prettier and merged.

No problem thanks.

PR #646, opened 2021-01-27
Provision: Replace virtualenv by python native venv.

### LINKED ISSUE #625: Move away from virtualenv
Since python3.5, the recommended methodology for creating virtual environments is to use the native venv module.

Now that we no longer support older versions of python, we no longer need virtualenv.

More discussion here: https://chat.zulip.org/#narrow/stream/2-general/topic/page.20not.20found/near/1022651

LoopThrough-i-j:
Fixes #625

### DIFF
--- a/tools/provision
+++ b/tools/provision
@@ -6,8 +6,6 @@ import argparse
 import subprocess
 import glob
 
-from importlib import import_module
-
 CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
 ZULIP_BOTS_DIR = os.path.join(CURRENT_DIR, '..', 'zulip_bots')
 sys.path.append(ZULIP_BOTS_DIR)
@@ -48,16 +46,19 @@ the Python version this command is executed with."""
     venv_dir = os.path.join(base_dir, venv_name)
     if not os.path.isdir(venv_dir):
         try:
-            return_code = subprocess.call(['virtualenv', '-p', options.python_interpreter, venv_dir])
+            return_code = subprocess.call([options.python_interpreter, '-m', 'venv', venv_dir])
         except OSError:
-            if subprocess.call(['which', 'virtualenv']):
-                print("{red}Please install the virtualenv package and try again.{end_format}"
-                      .format(red='\033[91m', end_format='\033[0m'))
-                sys.exit(1)
+            print("{red}Installation with venv failed. Probable errors are: "
+                  "You are on Ubuntu and you haven't installed python3-venv,"
+                  "or you are running an unsupported python version"
+                  "or python is not installed properly{end_format}"
+                  .format(red=red, end_format=end_format))
+            sys.exit(1)
             raise
         else:
+            # subprocess.call returns 0 if a script executed successfully
             if return_code:
-                raise OSError("The command `virtualenv -p {} {}` failed. Virtualenv not created!"
+                raise OSError("The command `{} -m venv {}` failed. Virtualenv not created!"
                               .format(options.python_interpreter, venv_dir))
             print("New virtualenv created.")
     else:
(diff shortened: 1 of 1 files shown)


LoopThrough-i-j (commented):
I am not sure about this line. Can anyone help?

tools/provision
```diff
@@ -39,7 +37,7 @@ the Python version this command is executed with."""
     py_version = tuple(int(num) for num in py_version_list[0:2])
     venv_name = 'zulip-api-py{}-venv'.format(py_version[0])
 
-    if py_version <= (3, 1) and (not options.force):
+    if py_version <= (3, 5) and (not options.force):
```
rht, line 40: This version constraint update should be in a separate commit.

tools/provision
```diff
@@ -48,37 +46,31 @@ the Python version this command is executed with."""
...
     if not os.path.isdir(venv_dir):
         try:
-            return_code = subprocess.call(['virtualenv', '-p', options.python_interpreter, venv_dir])
+            return_code = subprocess.call([options.python_interpreter, '-m', 'venv', venv_dir])
         except OSError:
-            if subprocess.call(['which', 'virtualenv']):
-                print("{red}Please install the virtualenv package and try again.{end_format}"
-                      .format(red='\033[91m', end_format='\033[0m'))
-                sys.exit(1)
+            print("{red}Installation with venv failed."
```
rht, line 51: I think on Ubuntu, `python3-venv` is a separate package (but not sure, haven't checked). If so, this err message doesn't take into account of the fact that the library hasn't been installed yet.

tools/provision
```diff
@@ -48,37 +46,31 @@ the Python version this command is executed with."""
...
     else:
         venv_exec_dir = 'bin'
 
-    # In order to install all required packages for the venv, we need to activate it. Since
-    # the activation script sets environmental variables, it needs to be executed inline with
-    # `import_module`.
-    activate_module_dir = os.path.abspath(os.path.join(venv_dir, venv_exec_dir))
-    sys.path.append(activate_module_dir)
-
-    import_module('activate_this')
```
rht, line 80: Why is this no longer necessary?

tools/provision
rht: Yes, confirmed that it is a separate package: https://packages.ubuntu.com/search?keywords=python3-venv. Note that this is specific to Ubuntu only.

tools/provision
LoopThrough-i-j: 1. There is no activate_this in venv 
2. If you check the way packages are getting installed we don't actually need to activate the environment. To install packages we are fetching the path to pip in the venv and installing packages using that, which limits the package installation to the environment.

tools/provision
LoopThrough-i-j: Yes, I know that as well. So do I mention that in the error message or run a subprocess for that? I think we can just add an extra line for that.

tools/provision
LoopThrough-i-j: okay, will do sure.

tools/provision
rht: Yes, you should say in error message that if you are on Ubuntu, you haven't installed `python3-venv`.

tools/provision
rht, line 40: I grepped for 3.5 in the codebase. It's not only in this line that the version needs to be updated. @LoopThrough-i-j I think you should do the version constraint thing in a separate PR.

e.g.
```
zulip_botserver/setup.py
27:        'Programming Language :: Python :: 3.5',
32:    python_requires='>=3.5',
```

and several more.

LoopThrough-i-j: ok makes sense will do.

timabbott:
Merged, thanks @LoopThrough-i-j!  I made a few edits to the commit message, mainly for technical English ("Replace virtualenv with python native venv.", i.e. I did s/by/with/).

tools/provision
rht: @LoopThrough-i-j have you made the PR to drop 3.5 yet?

rht: #655

LoopThrough-i-j:
Thanks, will keep it in mind.

Issue #637, opened 2020-12-08
Add remove_storage helper to the Python API.


showell:
This should be fairly straightforward for you if you have worked with extending our API before.  The actual code should be trivial, but you'll want to add tests for it and documentation.

https://chat.zulip.org/#narrow/stream/127-integrations/topic/Storage.20Python.20API/near/1074914

R0cker77:
@zulipbot claim

zulipbot:
Hello @R0cker77!

Thanks for your interest in Zulip! You have attempted to claim an issue without the labels "help wanted", "good first issue". Since you're a new contributor, you can only claim and submit pull requests for issues with the [help wanted](https://github.com/zulip/python-zulip-api/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22help+wanted%22) or [good first issue](https://github.com/zulip/python-zulip-api/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22good+first+issue%22) labels.

If this is your first time here, we recommend reading our [guide for new contributors](https://zulip.readthedocs.io/en/latest/overview/contributing.html) before getting started.

R0cker77:
@zulipbot claim

zulipbot:
Welcome to Zulip, @R0cker77! We just sent you an invite to collaborate on this repository at https://github.com/zulip/python-zulip-api/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

zulipbot:
Hello @R0cker77, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

[...]

PR #648, opened 2021-02-01
Added remove_storage helper to the Python API.

R0cker77:
A pr for issue #637.
Added `remove_storage` for the zulip API, and `storage.remove` for the bot framework, I'm not sure if I did the tests the right way, need some feedback on it.

i intend to add `storage.remove` to the [bot framework documentation](https://zulip.com/api/writing-bots) if this pr is good.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ and [commit guidelines](https://zulip.readthedocs.io/en/latest/contributing/vers
    `git clone https://github.com/<your_username>/python-zulip-api.git`
 
 2. Make sure you have [pip](https://pip.pypa.io/en/stable/installing/)
-   and [virtualenv](https://virtualenv.pypa.io/en/stable/installation/)
+   and [virtualenv](https://virtualenv.pypa.io/en/stable/installation.html)
    installed.
 
 3. `cd` into the repository cloned earlier:


showell:
LGTM.

timabbott:
@R0cker77 can you clean up the commit history for this PR?  

Check out our [GitHub guide](https://zulip.readthedocs.io/en/latest/git-guide.html) and [commit guidelines](https://zulip.readthedocs.io/en/latest/contributing/version-control.html) for more details.

zulip/zulip/__init__.py
```diff
@@ -1539,6 +1539,23 @@ def get_storage(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any
...
+    def remove_storage(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
+        '''
+            Example usage:
+
+            >>> client.update_storage({'storage': {"entry 1": "value 1", "entry 2": "value 2", "entry 3": "value 3"}})
+            >>> client.get_storage()
+            {'result': 'success',  'msg': '', 'storage': {"entry 1": "value 1", "entry 2": "value 2", "entry 3": "value 3"}}
+            >>> client.remove_storage({'keys': ["entry 1", "entry 3"]})
+            >>> client.get_storage()
+            {'result': 'success',  'msg': '', 'storage': {'entry 1': 'value 1'}}
```
timabbott, line 1551: I don't understand this example -- shouldn't only entry 2 be left?  

It's best to generate these from actual output from testing the tool to avoid bugs like this.

timabbott:
Thanks for working on this @R0cker77!  A few notes:
* I'd prefix the commit with `zulip:`, since that's the relevant subpackage.
* Generally you need to post a comment in GitHub when you've resolved the last comments -- otherwise maintainers won't know to look at your PR again.

zulip/zulip/__init__.py
R0cker77: Sorry about that, I edited the example to have only entry 2 left.
I checked the given example with the actual output of the tool, and everything checks out.
I hope it meets zulip standards.

R0cker77:
@timabbott Thank you for the feedback.
I added `zulip :` , prefix to commit, and changed the example to the correct form.
I’m eager to receive your feedback.

zulip/zulip/__init__.py
```diff
@@ -1539,6 +1539,23 @@ def get_storage(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any
...
+            {'result': 'success',  'msg': '', 'storage': {"entry 1": "value 1", "entry 2": "value 2", "entry 3": "value 3"}}
+            >>> client.remove_storage({'keys': ["entry 1", "entry 3"]})
+            >>> client.get_storage()
+            {'result': 'success',  'msg': '', 'storage': {'entry 2': 'value 2'}}
+        '''
+        return self.call_endpoint(
+            url='bot_storage',
+            method='DELETE',
+            request=request,
+        )
```
timabbott, line 1557: Can you change the `request` format to use the kwargs format for passing parameters (see e.g. #579 )?

Otherwise, this lgtm.

R0cker77: @timabbott I have changed `request` to use kwargs format and changed the rest of the pr accordingly.
Hope it meets your description.

zulip/zulip/__init__.py
```diff
@@ -1539,6 +1539,23 @@ def get_storage(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any
             request=request,
         )
 
+    def remove_storage(self, **request: Any) -> Dict[str, Any]:
```
LoopThrough-i-j, line 1542: I think we should type `request` properly, `Any` is not the way. Please go through [this](https://chat.zulip.org/#narrow/stream/92-learning/topic/mypy.20error/near/978210) message to understand how it should be done. Will highlight the same for the return type.

I think in this case it should be a `List[str]`, with a return of `Dict[str, str]`.

Moreover, I am not sure if we want to use `**request` as a param, `keys` I think would be better in terms of code clarity as we won't probably need more params for this function in the future.

```python
    return self.call_endpoint(
            url='bot_storage',
            method='DELETE',
            request={"keys": keys},
        )
```

LoopThrough-i-j (changes requested):

zulip/zulip/__init__.py
R0cker77: Hello @LoopThrough-i-j, sorry for the late response.
concerning the `**request` parameter and changing it to `keys`, do you mean something like this:
```
def remove_storage(self, keys: Optional[List[str]] = None) -> Dict[str, str]:
      
        return self.call_endpoint(
            url='bot_storage',
            method='DELETE',
            request={"keys": keys},
        )
```
And if so, why not just keep it at `request=keys` instead of `request={"keys": keys}`, to keep the consistency throughout the codebase, don't you think it will get a bit confusing to keep up on which function uses `keys=["entry 1"]` instead of `{'keys': ["entry 1"]}`  as a parameter, why make this the exception?

LoopThrough-i-j: I would like the keys to be:
```
keys: List[str]  # Not Optional
```
Why optional.

Moreover, I don't think it is reasonable to use `**request` on a function that can just have 1 parameter, just for consistency. Rather defining a proper parameter would be more reasonable.

> And if so, why not just keep it at request=keys instead of request={"keys": keys}
because that's how the `self.call_endpoint` works. It accepts a `Dict` and it is reasonable to do so.

LoopThrough-i-j: > don't you think it will get a bit confusing to keep up on which function uses keys=["entry 1"] instead of {'keys': ["entry 1"]} as a parameter

That's what the typing is for, and moreover, it won't be much of a difference from a user perspective.

LoopThrough-i-j: > why make this the exception?

This is not about making an exception, but about not following what is done everywhere just for consistency. Using the `keys` variable instead of `**request` will help the developers know exactly what are the arguments and what is getting passes in `request`, though this is not required or reasonable everywhere it makes sense here, knowing that there is not much scope for additional params.

R0cker77: > Why optional.


well, if left empty it removes the entire bot storage, but if you don't want it optional how can we remove the entire bot storage?

LoopThrough-i-j: we can send an empty list.

R0cker77: okay, or we can make this the parameter, what do you think?
`keys: List[str] = None`

LoopThrough-i-j: This will be a type error.

LoopThrough-i-j:
```python
return self.call_endpoint(
            url='bot_storage',
            method='DELETE',
            request={"keys": keys} if keys != [] else None,
        )
```

R0cker77: I tested out putting `keys: List[str] = []` and `keys: List[str] = None` and both of them work just fine, and i have noticed that other functions in the codebase uses None, without any issues, so i believe both of them work alike, what do you think?

LoopThrough-i-j: When Using `None` it is fine to use optional then, but I don't think we should make `keys: List[str] = None`, either use 
`keys: List[str] = []` or `keys: Optional[List[str]] = None`

LoopThrough-i-j: Still, I think sending an empty list is better, as the `bot_storage` API supports an empty list as well.

R0cker77: Okay, I changed it to `keys: List[str] = []`, and changed the rest of the pr accordingly, I hope it meets your standards.

LoopThrough-i-j (approved):

zulip/zulip/__init__.py
LoopThrough-i-j: This looks fine, thanks.

LoopThrough-i-j: These are not exactly my standards, just a small try to make sure the new code pushed in does not need reactors while cleaning up things in the future.

zulip/zulip/__init__.py
```diff
@@ -1539,6 +1539,23 @@ def get_storage(self, request: Optional[Dict[str, Any]] = None) -> Dict[str, Any
             request=request,
         )
 
+    def remove_storage(self, keys: List[str] = []) -> Dict[str, str]:
```
andersk, line 1542: I disagree with @LoopThrough-i-j: I think this default of `[]` is wrong and should be `None`, because the behavior of `remove_storage()` with no arguments should be to remove all keys, which is different than removing no keys. See https://zulip.readthedocs.io/en/latest/testing/mypy.html#use-optional-and-none-correctly.

It happens that our `DELETE /api/v1/bot_storage` with `keys = []` also has the effect of removing all keys, but that’s a side effect of its misuse of falsiness in `keys = keys or get_keys_in_bot_storage(user_profile)`, and I think that should be fixed.

LoopThrough-i-j: Okay then keys should be `keys: Optional[List[str]] = None`, and we should put a proper check-in the actual API. 

This line `keys = keys or get_keys_in_bot_storage(user_profile)`, bothered me about the proper usage.

LoopThrough-i-j: `[]` this should ideally not delete anything then.

R0cker77: I'm very sorry for the inconvenience, but I accidentally forced pushed a wrong commit, which closed this pull request, and I can not reopen it to redo my errors, it was an honest mistake from my end, and I believe I can not reopen it, because it got merged?, I'm completely lost at the moment.

andersk: @R0cker77 Yeah, force-pushing the PR to an ancestor of the target branch had the effect of closing it in such a way that it can’t be reopened. You’ll need to open a new PR.

R0cker77: Yeah, I'm going to make a new pr with `keys: Optional[List[str]] = None`.
And again I'm sorry about this mess.

PR #653, opened 2021-02-23
zulip/zulip/__init__.py: Update `reaction_data` type.

LoopThrough-i-j:
Update `reaction_data` type in add_reaction and
remove_reaction, from Dict[str, str] to Dict[str, Any]

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -829,12 +829,12 @@ class Client:
             method='GET'
         )
 
-    def add_reaction(self, reaction_data: Dict[str, str]) -> Dict[str, Any]:
+    def add_reaction(self, reaction_data: Dict[str, Any]) -> Dict[str, Any]:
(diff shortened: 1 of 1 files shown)


amanagr:
For both the PRs, you don't need to write the whole file name in the commit message.  See https://zulip.readthedocs.io/en/latest/contributing/version-control.html#commit-messages

Something like `reactions_api` would be more appropriate.

LoopThrough-i-j:
Actually, I saw Anders Kaseorg doing this for single file changes, hence thought that could be a better practice for a single file commit.

rht:
Posting the relevant discussion thread for future reference: https://chat.zulip.org/#narrow/stream/127-integrations/topic/msg.20id.20in.20add_emoji.20doc.

timabbott:
Merged, thanks @LoopThrough-i-j!

PR #656, opened 2021-02-25
zulip-bots: Add `BotHandler` Protocol.

### LINKED ISSUE #639: Make a base class or protocol for ExternalBotHandler that zulip/zulip’s EmbeddedBotHandler can subclass
zulip/zulip currently needs an [unchecked cast](https://github.com/zulip/zulip/blob/a054f57af68637ad8b96290b3e13b2a6d8a8a359/zerver/worker/queue_processors.py#L748) to convert a custom `EmbeddedBotHandler` class into an `ExternalBotHandler`. Since `EmbeddedBotHandler` is not a subclass of `ExternalBotHandler` (because it works by invoking `internal_send_message` rather than making API calls), this is unsafe, and it breaks in practice when we try to upgrade python-zulip-api to 0.7.1 (zulip/zulip#16511).

To fix this, we need a common base class or [Mypy protocol](https://mypy.readthedocs.io/en/stable/protocols.html#simple-user-defined-protocols), say `BotHandler`, that can be subclassed by both `ExternalBotHandler` and `EmbeddedBotHandler`. The `zulip_bots` functions must be relaxed to accept `BotHandler` instead of requiring `ExternalBotHandler`.

If the `StubBotHandler` class used in tests could subclass `BotHandler` too, I think this would also let us correct various `bot_handler: Any` types to `bot_handler: BotHandler`.

LoopThrough-i-j:
- zulip-bots: Add `CommonStorage` Protocol.
   -  The `CommonStorage` Protocol is created to add a
  Common type to all Storage Classes.
   -  The Storages used in `BotHandler`s are typed as `CommonStorage`

    Note: Protocol is imported from `typing_extensions` as `typing `doesn't
    provide Protocol for python <= 3.7.

- zulip-bots: Add `BotHandler` Protocol.

     The `BotHandler` Protocol has methods which allow
     any of the Current BotHandlers to use `BotHandler` Protocol
     as a mypy type.

    Fixes part of #639

- zulip-bots: Set `bot_handler` type to `BotHandler`.
    - Set `bot_handler` type to `BotHandler`.
    - Fix mypy issues in improperly typed variables and returns.

    Fixes #639

### DIFF
--- a/zulip_bots/setup.py
+++ b/zulip_bots/setup.py
@@ -65,6 +65,7 @@ setuptools_info = dict(
         'html2text',
         'lxml',
         'BeautifulSoup4',
+        'typing_extensions',
     ],
 )
 
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -112,6 +112,37 @@ class BotStorage(Protocol):
     def contains(self, key: Text) -> bool:
         ...
 
+class BotHandler(Protocol):
+
+    user_id: int
+    email: str
+    full_name: str
+
+    @property
+    def storage(self) -> BotStorage:
+        ...
+
+    def identity(self) -> BotIdentity:
+        ...
+
+    def react(self, message: Dict[str, Any], emoji_name: str) -> Dict[str, Any]:
+        ...
+
+    def send_message(self, message: Dict[str, Any]) -> Optional[Dict[str, Any]]:
+        ...
+
+    def send_reply(self, message: Dict[str, Any], response: str, widget_content: Optional[str] = None) -> Optional[Dict[str, Any]]:
+        ...
+
+    def update_message(self, message: Dict[str, Any]) -> Optional[Dict[str, Any]]:
+        ...
+
+    def get_config_info(self, bot_name: str, optional: bool = False) -> Dict[str, str]:
+        ...
+
+    def quit(self, message: str = "") -> None:
+        ...
+
 class ExternalBotHandler:
     def __init__(
         self,
@@ -168,7 +199,7 @@ class ExternalBotHandler:
                                               emoji_name=emoji_name,
                                               reaction_type='unicode_emoji'))
 
-    def send_message(self, message: (Dict[str, Any])) -> Dict[str, Any]:
+    def send_message(self, message: Dict[str, Any]) -> Dict[str, Any]:
         if not self._rate_limit.is_legal():
             self._rate_limit.show_error_and_exit()
         resp = self._client.send_message(message)
@@ -198,7 +229,7 @@ class ExternalBotHandler:
             self._rate_limit.show_error_and_exit()
         return self._client.update_message(message)
 
-    def get_config_info(self, bot_name: str, optional: Optional[bool] = False) -> Dict[str, Any]:
+    def get_config_info(self, bot_name: str, optional: bool = False) -> Dict[str, str]:
         if self._bot_config_parser is not None:
             config_parser = self._bot_config_parser
         else:
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -9,7 +9,7 @@ import re
 
 
 from typing import Any, Optional, List, Dict, IO, Text
-
+from typing_extensions import Protocol
 from zulip import Client, ZulipError
 
 
@@ -101,6 +101,16 @@ class BotIdentity:
         self.name = name
         self.email = email
         self.mention = '@**' + name + '**'
+
+class BotStorage(Protocol):
+    def put(self, key: Text, value: Any) -> None:
+        ...
+
+    def get(self, key: Text) -> Any:
+        ...
+
+    def contains(self, key: Text) -> bool:
+        ...
 
 class ExternalBotHandler:
     def __init__(
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -295,7 +295,7 @@ class ExternalBotHandler:
         sys.exit(message)
 
 
-def extract_query_without_mention(message: Dict[str, Any], client: ExternalBotHandler) -> Optional[str]:
+def extract_query_without_mention(message: Dict[str, Any], client: BotHandler) -> Optional[str]:
     """
     If the bot is the first @mention in the message, then this function returns
     the stripped message with the bot's @mention removed.  Otherwise, it returns None.
@@ -314,7 +314,7 @@ def extract_query_without_mention(message: Dict[str, Any], client: ExternalBotHa
     return None
 
 
-def is_private_message_but_not_group_pm(message_dict: Dict[str, Any], current_user: ExternalBotHandler) -> bool:
+def is_private_message_but_not_group_pm(message_dict: Dict[str, Any], current_user: BotHandler) -> bool:
     """
     Checks whether a message dict represents a PM from another user.
 
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -335,7 +335,7 @@ def display_config_file_errors(error_msg: str, config_file: str) -> None:
     print('\nMore details here:\n\n{}\n'.format(error_msg))
 
 
-def prepare_message_handler(bot: str, bot_handler: ExternalBotHandler, bot_lib_module: Any) -> Any:
+def prepare_message_handler(bot: str, bot_handler: BotHandler, bot_lib_module: Any) -> Any:
     message_handler = bot_lib_module.handler_class()
     if hasattr(message_handler, 'validate_config'):
         config_data = bot_handler.get_config_info(bot)
--- a/zulip_bots/zulip_bots/test_lib.py
+++ b/zulip_bots/zulip_bots/test_lib.py
@@ -28,6 +28,7 @@ class StubBotHandler:
         self.storage = SimpleStorage()
         self.full_name = 'test-bot'
         self.email = 'test-bot@example.com'
+        self.user_id = 0
         self.message_server = SimpleMessageServer()
         self.reset_transcript()
 
@@ -69,7 +70,7 @@ class StubBotHandler:
     def quit(self, message: str = "") -> None:
         raise self.BotQuitException()
 
-    def get_config_info(self, bot_name: str, optional: bool = False) -> Dict[str, Any]:
+    def get_config_info(self, bot_name: str, optional: bool = False) -> Dict[str, str]:
         return {}
 
     def unique_reply(self) -> Dict[str, Any]:
--- a/zulip_bots/zulip_bots/bots/baremetrics/baremetrics.py
+++ b/zulip_bots/zulip_bots/bots/baremetrics/baremetrics.py
@@ -1,10 +1,11 @@
 # See readme.md for instructions on running this code.
 
 from typing import Any, List, Dict
+from zulip_bots.lib import BotHandler
 import requests
 
 class BaremetricsHandler:
-    def initialize(self, bot_handler: Any) -> None:
+    def initialize(self, bot_handler: BotHandler) -> None:
         self.config_info = bot_handler.get_config_info('baremetrics')
         self.api_key = self.config_info['api_key']
 
@@ -27,7 +28,7 @@ class BaremetricsHandler:
 
         self.check_api_key(bot_handler)
 
-    def check_api_key(self, bot_handler: Any) -> None:
+    def check_api_key(self, bot_handler: BotHandler) -> None:
         url = "https://api.baremetrics.com/v1/account"
         test_query_response = requests.get(url, headers=self.auth_header)
         test_query_data = test_query_response.json()
@@ -46,7 +47,7 @@ class BaremetricsHandler:
         Version 1.0
         '''
 
-    def handle_message(self, message: Dict[str, Any], bot_handler: Any) -> None:
+    def handle_message(self, message: Dict[str, Any], bot_handler: BotHandler) -> None:
         content = message['content'].strip().split()
 
         if content == []:
--- a/zulip_bots/zulip_bots/bots/beeminder/beeminder.py
+++ b/zulip_bots/zulip_bots/bots/beeminder/beeminder.py
@@ -1,6 +1,7 @@
 import requests
 import logging
-from typing import Any, Dict
+from typing import Dict
+from zulip_bots.lib import BotHandler
 from requests.exceptions import ConnectionError
 
 help_message = '''
@@ -79,7 +80,7 @@ class BeeminderHandler:
     towards their beeminder goals via zulip
     '''
 
-    def initialize(self, bot_handler: Any) -> None:
+    def initialize(self, bot_handler: BotHandler) -> None:
         self.config_info = bot_handler.get_config_info('beeminder')
         # Check for valid auth_token
         auth_token = self.config_info['auth_token']
@@ -93,7 +94,7 @@ class BeeminderHandler:
     def usage(self) -> str:
         return "This plugin allows users to add datapoints towards their Beeminder goals"
 
-    def handle_message(self, message: Dict[str, str], bot_handler: Any) -> None:
+    def handle_message(self, message: Dict[str, str], bot_handler: BotHandler) -> None:
         response = get_beeminder_response(message['content'], self.config_info)
         bot_handler.send_reply(message, response)
 
--- a/zulip_bots/zulip_bots/bots/chessbot/chessbot.py
+++ b/zulip_bots/zulip_bots/bots/chessbot/chessbot.py
@@ -3,6 +3,7 @@ import chess.uci
 import re
 import copy
 from typing import Any, Optional, Dict
+from zulip_bots.lib import BotHandler
 
 START_REGEX = re.compile('start with other user$')
 START_COMPUTER_REGEX = re.compile(
@@ -22,7 +23,7 @@ class ChessHandler:
             'Stockfish program on this computer.'
         )
 
-    def initialize(self, bot_handler: Any) -> None:
+    def initialize(self, bot_handler: BotHandler) -> None:
         self.config_info = bot_handler.get_config_info('chess')
 
         try:
@@ -38,7 +39,7 @@ class ChessHandler:
     def handle_message(
         self,
         message: Dict[str, str],
-        bot_handler: Any
+        bot_handler: BotHandler
     ) -> None:
         content = message['content']
 
@@ -93,7 +94,7 @@ class ChessHandler:
                 last_fen
             )
 
-    def start(self, message: Dict[str, str], bot_handler: Any) -> None:
+    def start(self, message: Dict[str, str], bot_handler: BotHandler) -> None:
         """Starts a game with another user, with the current user as white.
         Replies to the bot handler.
 
@@ -115,7 +116,7 @@ class ChessHandler:
     def start_computer(
         self,
         message: Dict[str, str],
-        bot_handler: Any,
+        bot_handler: BotHandler,
         is_white_user: bool
     ) -> None:
         """Starts a game with the computer. Replies to the bot handler.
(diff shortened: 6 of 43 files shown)


zulip_bots/zulip_bots/lib.py
```diff
@@ -102,7 +102,74 @@ def __init__(self, name: str, email: str) -> None:
...
+    def upload_file(self, file: IO[Any]) -> Dict[str, Any]:
+        pass
+
+    def unique_reply(self) -> Dict[str, Any]:
+        pass
+
+    def unique_response(self) -> Dict[str, Any]:
+        pass
+
+    def ensure_unique_response(
```
neiljp, line 163: Aren't these just test helpers? Similarly reset_transcript and maybe others?

zulip_bots/zulip_bots/lib.py
```diff
@@ -102,7 +102,74 @@ def __init__(self, name: str, email: str) -> None:
...
+    ) -> Dict[str, Any]:
+        pass
+
+    def send_reply(
+        self,
+        message: Dict[str, Any],
+        response: str,
+        widget_content: Optional[str] = None
+    ) -> Dict[str, Any]:
+        pass
```
neiljp, line 136: Should this be `...` ?

zulip_bots/zulip_bots/lib.py
```diff
@@ -9,7 +9,7 @@
 
 
 from typing import Any, Optional, List, Dict, IO, Text
-
+from typing_extensions import Protocol
```
neiljp, line 12: This module has not been added to dependencies to be installed?

neiljp (commented):
@LoopThrough-i-j Some initial thoughts from a quick look. I think the methods could be one-liners in some cases with the ellipsis.

zulip_bots/zulip_bots/lib.py
LoopThrough-i-j: I don't think we need a separate installation for this library.

zulip_bots/zulip_bots/lib.py
LoopThrough-i-j: These are test helpers and they were added because there were a few failing tests when `bot_handler` was assigned the type `BotHandler` in the next commit.

zulip_bots/zulip_bots/lib.py
neiljp: I don't believe typing_extensions is included in python3.6?
I tested this in a venv and `zulip-run-bot` rightly says
```
ModuleNotFoundError: No module named 'typing_extensions'

File: zulip_bots/zulip_bots/lib.py (hunk 1 above)
Comment (LoopThrough-i-j) [reply zulip_bots/zulip_bots/lib.py:163]: Will still take a look at improving it if possible.

File: zulip_bots/zulip_bots/lib.py (hunk 3 above)
Comment (LoopThrough-i-j) [reply zulip_bots/zulip_bots/lib.py:12]: Will add that.

File: zulip_bots/zulip_bots/lib.py (hunk 1 above)
Comment (LoopThrough-i-j) [reply zulip_bots/zulip_bots/lib.py]: @neiljp yeah, I considered removing the helper functions which were not really necessary.

File: zulip_bots/zulip_bots/lib.py (hunk 2 above)
Comment (LoopThrough-i-j) [reply zulip_bots/zulip_bots/lib.py:133]: I am not sure about what you are saying! Do you mean I should use *args or **kwargs for this because of the EmbeddedBotHandler in Zulip?

Comment (LoopThrough-i-j) [reply zulip_bots/zulip_bots/lib.py:133]: I am not sure of what is `...`!! Do you mean these should be in 1 line?

Comment (neiljp) [reply zulip_bots/zulip_bots/lib.py:121]: https://mypy.readthedocs.io/en/stable/protocols.html#simple-user-defined-protocols

Comment (LoopThrough-i-j) [reply zulip_bots/zulip_bots/lib.py:121]: Ahhh, ok I understand now. Probably missed this part while going through the library.

Comment (LoopThrough-i-j) [reply zulip_bots/zulip_bots/lib.py]: Thanks, I resolved it.

**Comment by LoopThrough-i-j** [comment]
@neiljp thanks for pointing out some improvements. I had actually added inherited the `BotHandler` class and added some helper functions in the `BotHandler`, confused by the innumerable number of errors thrown by the bots after the 2nd commit, which I slowly debugged and figured out and then considered keeping the `BotHandler` the same.

**Review by neiljp — COMMENTED** [review]
@LoopThrough-i-j This looks tidier already:+1: I'd check the server bots for the methods in common (quit?), or their types (Any?).

File: zulip_bots/zulip_bots/lib.py [hunk 4]
```diff
@@ -102,6 +102,33 @@ def __init__(self, name: str, email: str) -> None:
         self.email = email
         self.mention = '@**' + name + '**'
 
+class BotHandler(Protocol):
+
+    email: str
+    full_name: str
+    storage: Any
```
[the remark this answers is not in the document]
LoopThrough-i-j: Okay, am taking a look.

zulip_bots/zulip_bots/lib.py
```diff
@@ -102,6 +102,33 @@ def __init__(self, name: str, email: str) -> None:
...
+
+    def identity(self) -> BotIdentity:
+        ...
```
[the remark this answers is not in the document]
LoopThrough-i-j: Flake8 will throw errors :
```flake8    | zulip_bots/zulip_bots/lib.py:111:5: E704 multiple statements on one line (def)```

File: zulip_bots/zulip_bots/lib.py [hunk 6]
```diff
@@ -188,7 +215,7 @@ def update_message(self, message: Dict[str, Any]) -> Dict[str, Any]:
             self._rate_limit.show_error_and_exit()
         return self._client.update_message(message)
 
-    def get_config_info(self, bot_name: str, optional: Optional[bool] = False) -> Dict[str, Any]:
+    def get_config_info(self, bot_name: str, optional: bool = False) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
LoopThrough-i-j: Yeah, anyways Optional[bool] doesn't make sense as we are assigning it False by default. I will add it to the commit message.

zulip_bots/zulip_bots/lib.py
```diff
@@ -102,6 +102,33 @@ def __init__(self, name: str, email: str) -> None:
...
+    def send_reply(self, message: Dict[str, Any], response: str, widget_content: Optional[str] = None) -> Dict[str, Any]:
+        ...
+
+    def update_message(self, message: Dict[str, Any]) -> Optional[Dict[str, Any]]:
+        ...
+
+    def get_config_info(self, bot_name: str, optional: bool = False) -> Dict[str, Any]:
+        ...
+
+    def quit(self, message: str = "") -> None:
```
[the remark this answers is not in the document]
LoopThrough-i-j: Yeah, probably quit is also not required.

zulip_bots/zulip_bots/lib.py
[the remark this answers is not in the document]
neiljp: The default value is independent of the type. I doubt the type is `Union[None, bool]`, but that's a separate point.

zulip_bots/zulip_bots/lib.py
```diff
@@ -102,6 +102,33 @@ def __init__(self, name: str, email: str) -> None:
...
+    full_name: str
+    storage: Any
+
+    def identity(self) -> BotIdentity:
+        ...
+
+    def react(self, message: Dict[str, Any], emoji_name: str) -> Dict[str, Any]:
+        ...
+
+    def send_message(self, message: (Dict[str, Any])) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
LoopThrough-i-j: Ohh, it is because I copied the function from `ExternalBotHandler` and it had the same.

zulip_bots/zulip_bots/lib.py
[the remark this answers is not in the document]
LoopThrough-i-j: okay will still take a look, tests pass though.

zulip_bots/zulip_bots/lib.py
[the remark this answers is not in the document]
LoopThrough-i-j: getting this error when removing quit.
```
mypy      | zulip_bots/zulip_bots/bots/idonethis/idonethis.py:140: error: "BotHandler" has no attribute "quit"
```

zulip_bots/zulip_bots/lib.py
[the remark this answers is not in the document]
LoopThrough-i-j: Changed it to the original.

zulip_bots/zulip_bots/lib.py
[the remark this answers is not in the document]
LoopThrough-i-j: Decided to create a `CommonStorage` Protocol to solve this.

zulip_bots/zulip_bots/lib.py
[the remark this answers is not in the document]
neiljp: I wasn't meaning it shouldn't be changed, but we can leave it for now, pending what others think.

zulip_bots/zulip_bots/lib.py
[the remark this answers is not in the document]
neiljp: Did you check the server code? It does seem like it is there.

zulip_bots/setup.py
```diff
@@ -65,6 +65,7 @@
         'html2text',
         'lxml',
         'BeautifulSoup4',
+        'typing_extensions',
```
neiljp, line 68: Is the change to requirements.txt necessary?
Do we need a specific/minimum version of typing_extensions?

zulip_bots/zulip_bots/test_lib.py
```diff
@@ -25,7 +26,7 @@
 
 class StubBotHandler:
     def __init__(self) -> None:
-        self.storage = SimpleStorage()
+        self.storage = SimpleStorage()  # type: CommonStorage
```
neiljp, line 29: This looks good, though I'm not sure if we need a separate protocol for this or if this will be a common implementation.
I would put this refactor into a separate commit.
We should be able to use an inline type annotation.

zulip_bots/zulip_bots/lib.py
```diff
@@ -102,6 +102,45 @@ def __init__(self, name: str, email: str) -> None:
...
+
+    def react(self, message: Dict[str, Any], emoji_name: str) -> Dict[str, Any]:
+        ...
+
+    def send_message(self, message: Dict[str, Any]) -> Dict[str, Any]:
+        ...
+
+    # For ExternalBotHandler and StubBotHandler returns Dict[str, Any]
+    # Whereas EmbeddedBotHandler returns None
+    def send_reply(self, message: Dict[str, Any], response: str, widget_content: Optional[str] = None) -> Any:
```
neiljp, line 132: There is also variation in other methods?

neiljp (commented):
@LoopThrough-i-j This is looking better, though please check the server code for comparison if you've not already. Some of the errors you found are likely since the implementation is using a BotHandler and that doesn't supply the method, it's possible only an ExternalBotHandler does.

There seem to be quite a few deviations between the External and Embedded cases, so I'm not sure where to align to here, though since embedded bots presumably are upgraded with the server they could be adapted more easily?

@andersk I've given this some passes, but I'd appreciate your input based on the original motivation and how you see the two versions coming together.

zulip_bots/zulip_bots/lib.py
andersk: We should avoid `Any` wherever possible because it’s completely unchecked; mypy will silently allow you to put in one type and get out another. Consider one of

* `-> object` (see [`Any` vs. `object`](https://mypy.readthedocs.io/en/stable/dynamic_typing.html#any-vs-object));
* `-> Optional[Dict[str, Any]]`;
* `-> Dict[str, Any]` and add the return value in all cases;
* `-> None` and remove the return value in all cases.

LoopThrough-i-j: @andersk 
If we use any of:
- `Optional[Dict[str, Any]]`;
- `Dict[str, Any]`;
- `None`

The method will start expecting the same return declaration in the actual class methods.
Tried Overloading as well but didn't work.

Thanks for the object vs Any, comparison. I think I should update it as an object.

zulip_bots/zulip_bots/test_lib.py
LoopThrough-i-j: The only way to provide a static type to storage is by using a Protocol.
Otherwise making it an object is fine as well, but I think it is a better way to go.

zulip_bots/setup.py
LoopThrough-i-j: Yeah, they aren't, noticed it later.

zulip_bots/zulip_bots/lib.py
andersk: It’s perfectly [valid](https://mypy-play.net/?mypy=latest&python=3.9&gist=dd4e030f4136c7cd74d074e04fd2c8d8) to implement a protocol method returning `Optional[Dict[str, Any]]` with an implementation method returning `Dict[str, Any]` or `None`, as these are both subtypes of `Optional[Dict[str, Any]]`.

For the other two suggestions, you’ve left off the important part: “and add/remove the return value in all cases”. I think one of these is probably the right plan. Even though all four plans would be accepted by mypy, there’s no actual reason to a return a value in some implementations but not others. What’s the point if the caller can’t rely on it being there?

Avoid overloading; it’s also completely unchecked.

LoopThrough-i-j: But I think we cannot use objects here because it says objects are not indexable, but Dicts are. So, I think Any is the only way to go. The same reason why we can't use `Optional[Dict[str, Any]]`, Optional will give it a possibility of None making it non-indexable as well.

andersk: `None` doesn’t become any more indexable just because you hid it away in an `Any` type. Mypy won’t be able to find the error, but it’s still an error.

andersk: (Can you reply with normal replies instead of editing your original comment? It’s confusing to try to have this discussion out of order.)

I don’t see why you’re saying “that is not true for a `Dict` return” when I just demonstrated that it is.

zulip_bots/zulip_bots/test_lib.py
LoopThrough-i-j: Pushed in a separate commit.

zulip_bots/zulip_bots/lib.py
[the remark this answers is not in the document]
LoopThrough-i-j: yeah, they quit as well.

zulip_bots/zulip_bots/lib.py
```diff
@@ -102,6 +102,47 @@ def __init__(self, name: str, email: str) -> None:
...
+        ...
+
+    def send_message(self, message: Dict[str, Any]) -> Dict[str, Any]:
+        ...
+
+    # ExternalBotHandler and StubBotHandler returns Dict[str, Any]
+    # Whereas EmbeddedBotHandler returns None
+    # Optional[Dict[str, Any]] or object cannot be used because there are occasions where
+    # function calls to send_reply strictly expects Dict[str, Any].
+    def send_reply(self, message: Dict[str, Any], response: str, widget_content: Optional[str] = None) -> Any:
```
andersk, line 134: If there are callers that expect `Dict[str, Any]`, and there are implementation that provide `None`, then that is a bug. The correct resolution isn’t to reach for `Any` and prevent mypy from seeing that there is a bug. The correct resolution is to fix the bug. Either the implementations should be fixed to provide `Dict[str, Any]`, or the callers should be fixed to check for `None`.

zulip_bots/zulip_bots/lib.py
LoopThrough-i-j: I am sorry to edit my message, I thought of condensing everything in one message. Won't do that from now. 
[this](https://mypy-play.net/?mypy=latest&python=3.9&gist=dd4e030f4136c7cd74d074e04fd2c8d8) defines the error.

LoopThrough-i-j: oops it didn't save

LoopThrough-i-j: I am not sure how to share mypy playground:
The error is thrown when we run this:

try this out:
```
from typing import Any, Dict, Optional

from typing_extensions import Protocol

class P(Protocol):
    def foo(self) -> Optional[Dict[str, Any]]:
        ...

class A:
    def foo(self) -> Dict[str, Any]:
        return {"test": "test"}

class B:
    def foo(self) -> None:
        pass

new = A() # type: P
d = new.foo()
print(d["test"])
```

zulip_bots/zulip_bots/lib.py
LoopThrough-i-j: The `ExternalBotHandler` always returns a `Dict` [see this](https://github.com/zulip/python-zulip-api/blob/3887ad102e26efd1a9e3ed80e95a3ca4ad6600f8/zulip_bots/zulip_bots/lib.py#L161), but the `EmbeddedBotHandler` always returns `None` [see this](https://github.com/zulip/zulip/blob/829f9272d2c4299a0c0a37a09802248d8136c0a8/zerver/lib/bot_lib.py#L111). Our Protocol should be compatible to both I hope. If not then probably I understood it wrong.

LoopThrough-i-j: SO I don't think it is a problem for caller's side

zulip_bots/zulip_bots/lib.py
andersk: Right: `A()` is a valid implementation of `P`, but `d["test"]` is not a valid use of `P`.

If you want `B()` to be a valid implementation, then `d["test"]` cannot be a valid use. If you want `d["test"]` to be a valid use, then `B()` cannot be a valid implementation. One or the other will need to be fixed.

LoopThrough-i-j: Yes, that's my point. I have commented on the reason behind me choosing `Any` is in reply to your next comment.

andersk: But this is not a reason to choose `Any`!

We are clearly failing to communicate somewhere, so let me try to simplify. Let’s take all the protocols and classes out of this and just look at some basic types.

Mypy rejects this program because it is wrong:

```python
d = None
print(d["test"])
```

Mypy accepts this program, but it is **still just as wrong**:

```python
d: Any = None
print(d["test"])
```

Using `Any` has **not fixed the bug**. It has only silenced mypy.

Our goal is not to silence mypy. Our goal is to fix the bug.

We can fix the bug by having the consumer check for `None`:

```python
d: Optional[Dict[str, str]] = None
if d is not None:
    print(d["test"])
```

or we can fix the bug by having the producer avoid producing `None`:

```python
d: Dict[str, str] = {"test": "yay"}
print(d["test"])
```

LoopThrough-i-j: I thought we shouldn't put `Optional` because `bot_handler` was an instance of `StubBotHandler` which returns `Dict` strictly, but we have made `bot_handler` an instance of `BotHandler` which made the type `Optional[Dict]`.

will fix this error:
```
mypy      | zulip_bots/zulip_bots/bots/incrementor/incrementor.py:37: error: Value of type "object" is not indexable
```
by putting a check and type as Optional[Dict].

I think fix 1:
```python
d: Optional[Dict[str, str]] = None
if d is not None:
    print(d["test"])
```
is better, then we would ideally be able to send any kind of bot_handler to `bot_handler` parameter.

Thanks for helping soo much, and sorry for being a dumb kid.

andersk:
In order to use this in zulip/zulip, we also need to adjust `zulip_bots.lib.extract_query_without_mention` to take `BotHandler` instead of `ExternalBotHandler`. There may be other such adjustments that make sense.

LoopThrough-i-j:
> In order to use this in zulip/zulip, we also need to adjust `zulip_bots.lib.extract_query_without_mention` to take `BotHandler` instead of `ExternalBotHandler`. There may be other such adjustments that make sense.

Thanks, I didn't know. Will get things fixed soon.

andersk (approved):
I renamed `CommonStorage` to `BotStorage`. Otherwise looks good.

PR #663, opened 2021-03-06
Initialize CLI interface for python-zulip-api.

rht:


zulip/zulip/cli.py
```diff
@@ -0,0 +1,195 @@
...
+    log_exit(response)
+
+
+# Streams API
+
+
+@cli.command()
+def get_subscriptions() -> None:
+    """Get all streams that the user is subscribed to."""
+    # TODO python-zulip-api deviates from zulip.yaml here
```
timabbott, line 189: I think python-zulip-api is wrong, and we should migrate it.  

Can you open an issue or quick PR?

timabbott:
This is cool!  I think it'd really helpful to have a README (that perhaps we can then port into the /api documentation) with a handful of examples.  (I imagine it mostly self-documents with `--help`?)

zulip/zulip/cli.py
rht: This comment is obsolete. I already made the PR in #665. Already merged.

rht:
@timabbott rebased, and used `get_subscriptions` instead of `list_subscriptions`.

timabbott:
Merged, thanks @rht!  We'll want to do a bit more playing with this and then add it to the documentation.

@PIG208 @LoopThrough-i-j @showell are any of you up for spending some time playing with this, perhaps live-blogging the experience in #integrations?

zulip/zulip/cli.py
```diff
@@ -0,0 +1,194 @@
+#!/usr/bin/env python3
+import logging
+import sys
+
+from typing import Any, Dict, List
+
+import zulip
+import click
```
andersk, line 8: This `click` requirement is not declared in `setup.py`.

zulip/zulip/cli.py
```diff
@@ -0,0 +1,194 @@
...
+
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+log = logging.getLogger("zulip-cli")
+
+client = zulip.Client(config_file="~/zuliprc")
```
andersk, line 13: This seems like a poor location for a config file, especially since it can’t be overridden.

rht: Not sure the best way to handle this. There is already https://github.com/zulip/python-zulip-api/blob/717a549f4b097b7d08470bb860f0346c30d4ecd8/zulip_bots/zulip_bots/run.py#L55, but it's buried down inside zulip_bots. Should we move it to zulip library instead?

zulip/zulip/cli.py
rht: #694

zulip/zulip/cli.py
timabbott: I think we should just add the argparse logic from the core library for how to find a zuliprc file.

PR #665, opened 2021-03-10
API: Rename list_subscriptions to get_subscriptions.

rht:
I replaced every occurrences of list_subscriptions and list-subscriptions except for https://github.com/zulip/python-zulip-api/blob/5b5fda2354f4488693f0f00def68eb8a9b426e67/zulip_bots/zulip_bots/bots/baremetrics/test_baremetrics.py#L68-L75 and list-subscriptions in baremetrics.

### DIFF
--- a/zulip/integrations/bridge_with_irc/irc_mirror_backend.py
+++ b/zulip/integrations/bridge_with_irc/irc_mirror_backend.py
@@ -36,7 +36,7 @@ class IRCBot(irc.bot.SingleServerIRCBot):
         print("Listening now. Please send an IRC message to verify operation")
 
     def check_subscription_or_die(self) -> None:
-        resp = self.zulip_client.list_subscriptions()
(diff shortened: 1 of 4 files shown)


timabbott:
Nice, merged, thanks @rht!

I suspect `list_members` has the same issue.

Oh, and now that I think about it, check the filenames in `api/examples`.

rht:
I can't find `list_members`. There is `get_members`, but is marked as deprecated in https://github.com/zulip/python-zulip-api/blob/61abb4f59cefdf57fe5cce0c1b6fdaf15d93524e/zulip/zulip/__init__.py#L1241
```
        # This exists for backwards-compatibility; we renamed this
        # function get_users for consistency with the rest of the API.
        # Later, we may want to add a warning for clients using this
        # legacy name.
```

rht:
>  Later, we may want to add a warning for clients using this legacy name.

Why later? Why not now?

Issue #647, opened 2021-01-30
Provide a mechanism for passing `all_public_streams` to `call_on_each_*`


timabbott:
I suppose the specific flag isn't important -- I named an undocumented one in the title, but `call_on_each_*` need to support passing additional parameters to `/register`, and currently seem not to.  Probably simplest is to just support an additional parameters dictionary?  But mainly, we should try to be consistent with existing functions.

Once we add the feature, we should remember to update zulip/zulip to document it.

LoopThrough-i-j:
@timabbott I wanted to work on this issue, but I have some PRs waiting to be reviewed. Can I work on this?

LoopThrough-i-j:
@zulipbot claim

zulipbot:
Hello @LoopThrough-i-j!

Thanks for your interest in Zulip! You have attempted to claim an issue without the labels "help wanted", "good first issue". Since you're a new contributor, you can only claim and submit pull requests for issues with the [help wanted](https://github.com/zulip/python-zulip-api/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22help+wanted%22) or [good first issue](https://github.com/zulip/python-zulip-api/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22good+first+issue%22) labels.

If this is your first time here, we recommend reading our [guide for new contributors](https://zulip.readthedocs.io/en/latest/overview/contributing.html) before getting started.

zulipbot:
Hello @LoopThrough-i-j!

Thanks for your interest in Zulip! You have attempted to claim an issue without the labels "help wanted", "good first issue". Since you're a new contributor, you can only claim and submit pull requests for issues with the [help wanted](https://github.com/zulip/python-zulip-api/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22help+wanted%22) or [good first issue](https://github.com/zulip/python-zulip-api/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22good+first+issue%22) labels.

If this is your first time here, we recommend reading our [guide for new contributors](https://zulip.readthedocs.io/en/latest/overview/contributing.html) before getting started.

LoopThrough-i-j:
So, the documentation for this is left I hope.

timabbott:
[...]

PR #666, opened 2021-03-10
zulip: Add support for additional  args for call_on_each_*.

LoopThrough-i-j:
Fixes part of #647

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -617,17 +617,17 @@ class Client:
         callback: Callable[[Dict[str, Any]], None],
         event_types: Optional[List[str]] = None,
         narrow: Optional[List[List[str]]] = None,
+        **kwargs: object,
     ) -> None:
         if narrow is None:
             narrow = []
 
         def do_register() -> Tuple[str, int]:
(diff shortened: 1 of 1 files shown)


zulip/zulip/__init__.py
```diff
@@ -617,17 +617,17 @@ def call_on_each_event(
         callback: Callable[[Dict[str, Any]], None],
         event_types: Optional[List[str]] = None,
         narrow: Optional[List[List[str]]] = None,
+        kwargs: Dict[str, object] = {},
```
andersk, line 620: It’s dangerous to specify a mutable type for a parameter with a default argument, because the default argument is shared between all callers; use `Mapping`, not `Dict`.

However, enclosing these extra arguments in a dictionary is inconsistent with the `register` API. I think we should use `**kwargs`.

LoopThrough-i-j:
@timabbott and @andersk I think this is all we need.

timabbott:
Merged, thanks @LoopThrough-i-j!  Once we update python-zulip-api in zulip/zulip, we can add documentation for how to do this to the API docs.

PR #668, opened 2021-03-10
zephyr: Attempt to fix types

andersk:
The mirror has some chance of running on Python 3 now, once the python-zephyr patch is rebased on 0.2.1, though it’s untested.

Cc @timabbott


zulip/integrations/zephyr/zephyr_mirror_backend.py
```diff
@@ -446,17 +462,6 @@ def process_notice(notice: Any, log: Optional[IO[Any]]) -> None:
...
-def decode_unicode_byte_strings(zeph: Dict[str, Any]) -> Dict[str, str]:
-    # 'Any' can be of any type of text that is converted to str.
-    for field in zeph.keys():
-        if isinstance(zeph[field], str):
-            try:
-                decoded = zeph[field].decode("utf-8")
-            except Exception:
-                decoded = zeph[field].decode("iso-8859-1")
-            zeph[field] = decoded
-    return zeph
```
timabbott, line 458: I don't remember the history of this block at all, but it certainly looks like it's for dealing with Zephyr clients using incorrect iso-8859-1 encoding in their messages, which must have been a real situation N years ago.  I'm fine with dropping it if you're up for doing whatever support is required related to this change, though this should at least be called out in the commit message as a functional change (or in its own commit).

timabbott:
I think you should be able to test this on any Athena dialup by running it oneself (basically the /mit/tabbott/for_friends/ method), at least at the level of starting up and delivering a PM?  But if that is nontrivial, let me know and I can test.

timabbott:
(In general, master is nonfunctional for this right now anyway, so I'm happy to merge things that are a step forward as long as we think they're right).

timabbott:
BTW if you have a good way to get in touch with broder, we should arrange a python-zephyr release; I just don't think I have bits to do so.

zulip/integrations/zephyr/zephyr_mirror_backend.py
andersk: Thinking of this as a “functional change” is complicated by the observation that the previous version is guaranteed non-functional in Python 3: a `str` cannot be decoded. That’s the case with most of this diff.

One of the upstream changes in python-zephyr 0.2.1 seems to be that the bytes have already been decoded to strings (ebroder/python-zephyr@48f73c8d38c52b53084d2334ced0b2d5f0ea00b6).

timabbott:
Great, had forgotten that bit of Zephyr context.  Merged, thanks @andersk!  

Remind me to test this transition tomorrow or something.

PR #669, opened 2021-03-11
api: Add sorting for API responses.

MSurfer20:
Currently the API response keys were sorted in random order, making
the responses unreadable. Added JSON dumps to sort the response keys.
Fixes https://github.com/zulip/zulip/issues/17428

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -609,8 +609,8 @@ class Client:
             if v is not None:
                 marshalled_request[k] = v
         versioned_url = API_VERSTRING + (url if url is not None else "")
-        return self.do_api_query(marshalled_request, versioned_url, method=method,
-                                 longpolling=longpolling, files=files, timeout=timeout)
+        return json.dumps(self.do_api_query(marshalled_request, versioned_url, method=method,
+                                 longpolling=longpolling, files=files, timeout=timeout), sort_keys=True)
 
     def call_on_each_event(
         self,
--- a/test_function.py
+++ b/test_function.py
@@ -1,0 +1,10 @@
+#!/usr/bin/env python3
+
+import zulip
+
+# Pass the path to your zuliprc file here.
+client = zulip.Client(config_file="~/zuliprc")
+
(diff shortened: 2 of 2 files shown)


zulip/zulip/__init__.py
```diff
@@ -609,8 +609,8 @@ def call_endpoint(self, url: Optional[str] = None, method: str = "POST", request
             if v is not None:
                 marshalled_request[k] = v
         versioned_url = API_VERSTRING + (url if url is not None else "")
-        return self.do_api_query(marshalled_request, versioned_url, method=method,
-                                 longpolling=longpolling, files=files, timeout=timeout)
+        return json.dumps(self.do_api_query(marshalled_request, versioned_url, method=method,
```
timabbott, line 612: We definitely don't want to do JSON dumps here -- a program accessing this wants the Python dictionary.

What I think we want to do here is sorting in #663 and in all the tools in `examples/`, so the main PR to do here would be to adjust `examples/` appropriately.

MSurfer20: @timabbott Can we instead change this to 
```return json.loads(json.dumps(self.do_api_query(marshalled_request, versioned_url, method=method,
                                 longpolling=longpolling, files=files, timeout=timeout), sort_keys=True))```
since dictionaries are ordered since Python 3.7?

MSurfer20: @timabbott Should I instead go with the modifying `examples/` approach?

LoopThrough-i-j: @MSurfer20 I think we should go for changing `examples/`, but am not sure if it is worth solving the problem.

LoopThrough-i-j: I don't think this will help if someone is running zulip in an env where `python <= 3.6`.
The only way is updating `examples/` but I am not sure if it is worth the amount of work.

timabbott: I think after thinking about this more, I'm not sure there's anything we want to do client-side -- the extra code in the examples is unlikely to be worth it.

Issue #673, opened 2021-04-03
Incrementor bot is broken when the time limit for editing messages has expired


PIG208:
In the production environment of Zulip, calls to `update_message` might not actually update the message if the bot doesn't have access to edit the message. E.g.: when the time limit is expired, or message editing is disabled, etc. The implementation for incrementor should take respect of the error message from the server and send a new message when necessary.

PR #674, opened 2021-04-03
zulip-bots: Fix incrementor to handle edit time limit error.

PIG208:
#673 

This allows the incrementor to regard the error message from the server (such as having the edit time limit expired). It also adds a test for this feature. 
If the API returns an error message when the incrementor attempts to update a message, the incrementor will send a new message instead.


timabbott:
Merged, thanks @PIG208!

PR #676, opened 2021-04-06
sync-public-streams: Rewrite using an event queue

andersk:


timabbott:
Merged, thanks @andersk!

PR #680, opened 2021-04-26
zulip: Fix call_on_each_event.

LoopThrough-i-j:
Fix Improper assignement of narrow variable outside its scope.


timabbott:
Merged; I think it's worth unbreaking master immediately and then adding tests in a follow-up PR.  (Easy to verify the tests work by reverting this).

PR #681, opened 2021-04-27
zulip: Replace GET realm/filters with realm/linkifiers.

abhijeetbodas2001:
In zulip/zulip@3947b0c80a26b9126540ada996f027aac03063b7, we replaced
the legacy endpoint to use the term "linkifier" instead of
"filter" and to return the data in a dictionary format.

None of the official clients currently actually use this
endpoint.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -900,15 +900,25 @@ class Client:
             method='DELETE',
         )
 
-    def get_realm_filters(self) -> Dict[str, Any]:
+    def get_realm_linkifiers(self) -> Dict[str, Any]:
         '''
             Example usage:
 
-            >>> client.get_realm_filters()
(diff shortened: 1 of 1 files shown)


abhijeetbodas2001:
Do I need to do anything for backwards compatibility? I changed the name of the function here, so it could break any existing application which uses that function. But not changing would break it too, because the data returned would be different than what the client expected.

timabbott:
If we wanted to be fancy, what we'd do for backwards compatibility would be to have the `client` connect process do a `GET /server_settings` to verify the Zulip server when you initialize the client, and check the `feature_level` for which endpoint to call, etc.

I think given this endpoint's low use level and the fact that we haven't yet built said infrastructure, we can just merge this; folks can always run an old version of python-zulip-api as needed.

PR #682, opened 2021-04-29
zulip: Add tests for API functions.

LoopThrough-i-j:


zulip/setup.py
```diff
@@ -65,6 +65,7 @@ def recur_expand(target_root: Any, dir: Any) -> Generator[Tuple[str, List[str]],
 
 setuptools_info = dict(
     install_requires=['requests[security]>=0.12.1',
+                      'responses',
```
timabbott, line 68: We should make this a development-only dependency, like mypy.

zulip/tests/test_api_functions.py
```diff
@@ -0,0 +1,71 @@
...
+            json={'result': 'success', 'msg': '', 'events': [{'id': 123}]},
+            status=200
+        )
+        # For add_reaction
+        self.responses.add(
+            responses.POST,
+            url=re.compile("{}v1/messages/([0-9]*)/reactions".format(self.base_url)),
+            json={'result': 'success', 'msg': ''},
+            status=200
+        )
```
timabbott, line 35: I think we want to do this individually in each test, for just the endpoints we expect that test to call.  

Additionally, I would configure it for the precise URL that the test is going to use, and assert what parameters were actually present in the HTTP request -- the goal here is to use this to verify that the API bindings made the precise request that we expected, not simulate a server.

LoopThrough-i-j:
I was thinking of simulating an event queue for testing `call_on_each_event` but restricted myself from doing that thinking of the previous comments and from the prospects of setting up an integration test suite.

LoopThrough-i-j:
@timabbott a review on this if possible would be great.

zulip/tests/test_api.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @responses.activate
+    def test_add_reaction(self) -> None:
+        def request_callback(request: Any) -> Tuple[int, Dict[str, str], str]:
+            params = {}
+            for param in request.body.split("&"):
+                key, value = param.split("=")
+                params[key] = urllib.parse.unquote(value)
+            assert "emoji_name" in params or "emoji_code" in params
+            return (200, {}, json.dumps({'result': 'success', 'msg': ''}))
```
timabbott, line 23: I don't understand this function -- why do we need this parsing logic?

zulip/tests/test_api.py
```diff
@@ -0,0 +1,95 @@
...
+        # request with emoji code
+        request = {
+            "message_id": 10,
+            "emoji_code": "1f419",
+        }
+        result = client.add_reaction(request)
+        self.assertEqual(result, {'result': 'success', 'msg': ''})
+
+    @responses.activate
+    def test_call_on_each_event(self) -> None:
```
timabbott, line 48: I think this test could benefit from a comment or docstring explaining what we're trying to test here.

LoopThrough-i-j:
@timabbott thanks for the review I will update the PR shortly, with required comments and explanations as best as I can.

zulip/tests/test_api.py
LoopThrough-i-j: This function needs some work as well as I understand. This test is primarily for testing whether the payload provided is properly parsed, the URL is correct, and all the It accepts and sends a proper response, as a part of the normal API workflow.

This test makes sure all intermediate functions for an API call like `do_api_query` and `call_endpoint`, work correctly.

PR #685, opened 2021-05-11
botserver: Return a valid JSON that is acceptable to outgoing webhooks.

PIG208:
In zulip/zulip@b998138d3ab8ca0ea212c00469042349f8c3fe53, we introduce a check for responses from outgoing webhooks that require them to be a dictionary. This fixes the return value of the botserver view function to accommodate with the change from the serverside.


PIG208:
image
(No more "Invalid Message(s)" error)

PR #688, opened 2021-05-13
Implement react for TerminalBotHandler

### LINKED ISSUE #686: Implement react for TerminalBotHandler.
Just like `ExternalBotHandler` or other implementations, we should make `TerminalBotHandler` follow the `BotHandler` protocol. Otherwise even the `helloworld` bot cannot be run properly due to the fact that it doesn't support reactions. (We can simply mock it by just printing out the emoji name).

PIG208:
Fixes #686.

```
Enter your message: j

Reply from the bot is printed between the dotted lines:
-------
beep boop
-------
The bot reacts to message #1: wave
Enter your message: p

Reply from the bot is printed between the dotted lines:
-------
beep boop
-------
The bot reacts to message #3: wave
```

### DIFF
--- a/zulip_bots/zulip_bots/simple_lib.py
+++ b/zulip_bots/zulip_bots/simple_lib.py
@@ -40,10 +40,10 @@ class SimpleMessageServer:
         return dict(result='success', msg='', uri='https://server/user_uploads/{}'.format(uuid4()))
 
 class TerminalBotHandler:
-    def __init__(self, bot_config_file):
+    def __init__(self, bot_config_file, message_server):
         self.bot_config_file = bot_config_file
         self._storage = SimpleStorage()
-        self.message_server = SimpleMessageServer()
+        self.message_server = message_server
 
     @property
     def storage(self):
--- a/zulip_bots/zulip_bots/simple_lib.py
+++ b/zulip_bots/zulip_bots/simple_lib.py
@@ -53,7 +53,10 @@ class TerminalBotHandler:
         return BotIdentity("bot name", "bot-email@domain")
 
     def react(self, message, emoji_name):
-        raise NotImplementedError
+        print("""The bot reacts to message #{}: {}""".format(message["id"], emoji_name))
+        return self.message_server.add_reaction(dict(message_id=message['id'],
+                                                     emoji_name=emoji_name,
+                                                     reaction_type='unicode_emoji'))
 
     def send_message(self, message):
         if message['type'] == 'stream':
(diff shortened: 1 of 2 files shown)


zulip_bots/zulip_bots/simple_lib.py
```diff
@@ -39,15 +39,22 @@ def update(self, message):
     def upload_file(self, file):
         return dict(result='success', msg='', uri='https://server/user_uploads/{}'.format(uuid4()))
 
-class TerminalBotHandler:
+class TerminalBotHandler(BotHandler):
     def __init__(self, bot_config_file):
```
LoopThrough-i-j: I don't think `TerminalBotHandler` needs to inherit `BotHandler`, by default all `BotHandler`s follow the protocol.

PIG208: Fixed that.

LoopThrough-i-j:
I had a look and things look good. @neiljp would you mind reviewing?

zulip_bots/zulip_bots/simple_lib.py
timabbott: Hmm, I kinda feel like this should inherit from AbstractBotHandler once we do that rename.

timabbott:
This lgtm, merged, thanks @PIG208 and @LoopThrough-i-j!  

I think we should do a round of adding docstring/comment documentation to TerminalBotHandler explaining what's going on -- e.g. the `send_message` function looks like it's doing something twice now, basically because the context isn't clear that it's recording the data in the `SimpleMessageServer` mock server.  

(We might also want to rename `SimpleMessageServer to `MockZulipServer`, for clarity).

zulip_bots/zulip_bots/simple_lib.py
LoopThrough-i-j: If we inherit a protocol, all properties of the protocol now come into the class even if the class does not implement it. Since the implementation of abstract functions is not a mandate. I would rather recommend creating functions and not inheriting them so that we can come back to them later. Inheriting a Protocol also skips a lot of mypy issues like the ones we faced here which gets caught, if we keep them separate.

neiljp: I've not looked deeply at these recent changes but if this is intended to be a `Protocol` (~ static duck typing), then I agree with @LoopThrough-i-j re not inheriting. It may make it clearer to add a `Protocol` suffix to `BotHandler` to indicate that.

Issue #679, opened 2021-04-26
Add a context manager to manage storage


PIG208:
In the documentation, we have:
> Since each access to bot_handler.storage will involve a round-trip to the server, we recommend writing bots so that they do a single bot_handler.storage.get at the start of handle_message, and a single bot_handler.storage.put at the end to submit the state to the server. We plan to offer a context manager that takes care of this automatically.

The goal is to implement a context manager for `bot_handler.storage.put` and `bot_handler.storage.get`.

PIG208:
@zulipbot claim

PR #689, opened 2021-05-15
Implement context manager for bot storage

PIG208:
The context manager is implemented based on a newly added storage calledCachedStorage. It is buffered storage that doesn't sync with the database until it's flushed. With CachedStorage, we can implement the context manager easily by loading all the data on __enter__ and just flush all the modified (dirty) data on __exit__. This approach can help the user minimize the number of round-trips to the server almost invisibly (despite the fact that they need to use it with "with").

https://zulip.com/api/writing-bots#bot_handlerstorage

Manually tested with the incrementor bot.

Possible upcoming work: adding documentation and tests

#679

### DIFF
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -8,7 +8,8 @@ import time
 import re
 
 
-from typing import Any, Optional, List, Dict, IO, Text
+from contextlib import contextmanager
+from typing import Any, Iterator, Optional, List, Dict, IO, Set, Text
 from typing_extensions import Protocol
 from zulip import Client, ZulipError
 
@@ -83,6 +84,50 @@ class BotStorage(Protocol):
     def contains(self, key: Text) -> bool:
         ...
 
+class CachedStorage:
+    def __init__(self, parent_storage: BotStorage, init_data: Dict[str, Any]) -> None:
+        # CachedStorage is implemented solely for the context manager of any BotHandler.
+        # It has a parent_storage that is responsible of communicating with the database
+        #   1. when certain data is not cached;
+        #   2. when the data need to be flushed to the database.
+        # It can be initialized with the given data.
+        self._parent_storage = parent_storage
+        self._cache = init_data
+        self._dirty_keys: Set[str] = set()
+
+    def put(self, key: Text, value: Any) -> None:
+        # In the cached storage, values being put to the storage is not flushed to the parent storage.
+        # It will be marked dirty until it get flushed.
+        self._cache[key] = value
+        self._dirty_keys.add(key)
+
+    def get(self, key: Text) -> Any:
+        # Unless the key is not found in the cache, the cached storage will not lookup the parent storage.
+        if key in self._cache:
+            return self._cache[key]
+        else:
+            value = self._parent_storage.get(key)
+            self._cache[key] = value
+            return value
+
+    def flush(self) -> None:
+        # Flush the data to the parent storage.
+        # Data that are not marked dirty will be omitted.
+        # This should be manually called when CachedStorage is not used with a context manager.
+        while len(self._dirty_keys) > 0:
+            key = self._dirty_keys.pop()
+            self._parent_storage.put(key, self._cache[key])
+
+    def flush_one(self, key: Text) -> None:
+        self._dirty_keys.remove(key)
+        self._parent_storage.put(key, self._cache[key])
+
+    def contains(self, key: Text) -> bool:
+        if key in self._cache:
+            return True
+        else:
+            return self._parent_storage.contains(key)
+
 class StateHandler:
     def __init__(self, client: Client) -> None:
         self._client = client
@@ -111,6 +156,16 @@ class StateHandler:
     def contains(self, key: Text) -> bool:
         return key in self.state_
 
+@contextmanager
+def use_storage(storage: BotStorage, keys: List[Text]) -> Iterator[BotStorage]:
+    # The context manager for StateHandler that minimizes the number of round-trips to the server.
+    # It will fetch all the data using the specified keys and store them to
+    # a CachedStorage that will not communicate with the server until manually
+    # calling flush or getting some values that are not previously fetched.
+    data = {key: storage.get(key) for key in keys}
+    cache = CachedStorage(storage, data)
+    yield storage
+    cache.flush()
 
 class BotHandler(Protocol):
 
--- a/zulip_bots/zulip_bots/lib.py
+++ b/zulip_bots/zulip_bots/lib.py
@@ -67,6 +67,21 @@ class RateLimit:
         logging.error(self.error_message)
         sys.exit(1)
 
+class BotIdentity:
+    def __init__(self, name: str, email: str) -> None:
+        self.name = name
+        self.email = email
+        self.mention = '@**' + name + '**'
+
+class BotStorage(Protocol):
+    def put(self, key: Text, value: Any) -> None:
+        ...
+
+    def get(self, key: Text) -> Any:
+        ...
+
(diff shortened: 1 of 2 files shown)


PIG208:
For some reason, I need to explicitly add `BotStorage` to `SimpleStorage` and `CachedStorage` to make mypy happy. I think it might be the problem with the `@contextmanager` decorator in the protocol. But I'm not sure what is the correct way to define such context manager in a protocol.

zulip_bots/zulip_bots/lib.py
```diff
@@ -68,7 +68,28 @@ def show_error_and_exit(self) -> None:
...
+
+    def contains(self, key: Text) -> bool:
+        ...
+
+class BotStorage(BaseStorage):
+    @contextmanager
+    def use(self, keys: List[Text]) -> Iterator[BaseStorage]:
+        ...
+
+class StateHandler(BotStorage):
```
LoopThrough-i-j, line 92: I don't think `StateHandler` needs to import `BotStorage`

zulip_bots/zulip_bots/lib.py
```diff
@@ -68,7 +68,28 @@ def show_error_and_exit(self) -> None:
...
+    def put(self, key: Text, value: Any) -> None:
+        ...
+
+    def get(self, key: Text) -> Any:
+        ...
+
+    def contains(self, key: Text) -> bool:
+        ...
+
+class BotStorage(BaseStorage):
```
LoopThrough-i-j, line 87: I think We can Include the methods of `BaseStorage` in `BotStorage`, am not sure why we need a separate class for `BaseStorage`

zulip_bots/zulip_bots/simple_lib.py
```diff
@@ -1,10 +1,11 @@
 import configparser
+from contextlib import contextmanager
 import sys
 
-from zulip_bots.lib import BotIdentity
+from zulip_bots.lib import BotIdentity, BotStorage, CachedStorage
 from uuid import uuid4
 
-class SimpleStorage:
+class SimpleStorage(BotStorage):
```
LoopThrough-i-j, line 8: I think we don't need to inherit `BotStorage`.

zulip_bots/zulip_bots/lib.py
```diff
@@ -115,13 +114,73 @@ def get(self, key: Text) -> Any:
...
+        while len(self._dirty_keys) > 0:
+            key = self._dirty_keys.pop()
+            self._parent_storage.put(key, self._cache[key])
+
+    def flush_one(self, key: Text) -> None:
+        # Flush data with the specified key to the server.
+        self._dirty_keys.remove(key)
+        self._parent_storage.put(key, self._cache[key])
+
     def use(self, keys: List[Text]) -> Iterator[BaseStorage]:
```
LoopThrough-i-j, line 175: I am not sure why do we need this since the `CachedStorage` does not follow any `Protocol`. And am not sure if we explicitly need a `Protocol` for this.

zulip_bots/zulip_bots/lib.py
PIG208: If I don't do that, there will be a mypy error.

zulip_bots/zulip_bots/lib.py
PIG208: Because the return value of `use` needs to be BaseStorage. If it's in `BotStorage`, I can't annotate it to return a `BotStorage` as it's not yet defined.

zulip_bots/zulip_bots/lib.py
PIG208: That makes sense. perhaps I can simply delete it.

zulip_bots/zulip_bots/simple_lib.py
PIG208: Somehow mypy goes wrong here as well if I don't put it here. I'm not really sure why that happens.

PIG208: This is what the error looks like: https://github.com/zulip/python-zulip-api/runs/2590519387?check_suite_focus=true

PIG208:
Thank you for leaving the comments, I'm still trying to figure out what leads to the mypy errors, though everything works fine when I print-tested it. It would be the best if we don't have to inherit the protocol explicitly.

LoopThrough-i-j:
Yeah, I am investigating the problem as well, will let you know if I find the solution.

zulip_bots/zulip_bots/lib.py
```diff
@@ -93,24 +114,64 @@ def get(self, key: Text) -> Any:
         self.state_[key] = marshalled_value
         return self.demarshal(marshalled_value)
 
+    @contextmanager
+    def use(self, keys: List[Text]) -> Iterator[BaseStorage]:
```
LoopThrough-i-j, line 118: I would remommend `-> Iterator[CachedStorage]:` instead of `-> Iterator[BaseStorage]:`

LoopThrough-i-j: and have only one `BotStorage` and skip `BaseStorage`

LoopThrough-i-j:
Can you remove the Inheritance of protocols and let the build fail here? This would help us investigate better.

zulip_bots/zulip_bots/lib.py
PIG208: The second piece of advice is not achievable as I have to set the return type of `use` for the protocol.

PIG208: The dependency of these classes is now `BaseStorage` > `BotStorage` > `CachedStorage` > `StateHandler`. But I think it might not be necessary to have `StateHandler`'s `use` function to return a CachedStorage as the return type because only `BaseStorage` will be visible to the users as they can only see the protocol.

LoopThrough-i-j: Yeah, I understood the circular dependency issue, and `BaseStorage` is probably necessary. Thinking of a work around.

PIG208: I just found that there is a way to define recursive protocol so that we won't need `BaseStorage`.

LoopThrough-i-j:
I think typing has a `ContextManager` return type you can refer to (this)[https://stackoverflow.com/a/58349659].

PIG208:
It appears that I have fixed the problem by using the recursive definition. `Iterator[BotStorage]` works just fine. Thank you for your help!

LoopThrough-i-j:
No, problem. I pulled it to local and checked it myself.  Looks great.

zulip_bots/zulip_bots/lib.py
PIG208: BTW, I later found out that this is required as the `CachedStorage` will otherwise not satisfy the `BotStorage` protocol.

andersk:
Changing the `BotStorage` protocol is a backwards-incompatible change because the server has its own implementation for `EmbeddedBotHandler`. We could do that, but I don’t think it’s justified in this case. You could just as easily have users write `with CachedStorage(bot_handler.storage()) as storage:` without changing the protocol.

PIG208:
I have fixed the backward compatibility problem. But there is still space for optimization, like allowing the context manager to get and put all key-value pairs all at once instead of looping through the list with `bot_storage.get`. I couldn't think of a way to implement this improvement without changing the protocols.

timabbott:
This makes sense, merged, thanks @PIG208!  We can certainly change the protocol to do a single bulk operation down the line, we just need to coordinate carefully.  (The `EmbeddedBotHandler` is AFAIK not in production use yet).

It's definitely worth doing tests and documentation updates to recommend this.  Thanks @PIG208!

PIG208:
Opened #693 for this

PR #691, opened 2021-05-20
Update Slack bridge dependency and make it work on macOS.

rht:


zulip/integrations/bridge_with_slack/run-slack-bridge
```diff
@@ -123,8 +123,8 @@ if __name__ == "__main__":
         try:
             sb = SlackBridge(config)
 
-            zp = mp.Process(target=sb.zulip_client.call_on_each_message, args=(sb.zulip_to_slack(),))
-            sp = mp.Process(target=sb.run_slack_listener, args=())
+            zp = threading.Thread(target=sb.zulip_client.call_on_each_message, args=(sb.zulip_to_slack(),))
+            sp = threading.Thread(target=sb.run_slack_listener, args=())
```
neiljp, line 129: I'm not sure about the previous `Process` case, but my understanding is that exceptions cannot be propagated across threads; are these targets safe from that? What is the Exception block below expected to handle?

zulip/integrations/bridge_with_slack/run-slack-bridge
```diff
@@ -6,10 +6,12 @@ import os
 import argparse
 import traceback
 import threading
+
+from typing import Any, Dict, Callable
```
neiljp, line 10: Is this a manual reorder or per linting? (there's no note in the commit text re this)
How did this pass CI linting in the first place? Is there a linter for this?
(isort in ZT wouldn't have the blank line above, which makes me query this too)

neiljp (commented):
@rht Good to see the bridge being worked on - just a few queries from reading it after seeing the discussion on czo.

zulip/integrations/bridge_with_slack/run-slack-bridge
rht: I did the reordering manually. And the order has been that way for a long time, which means isort is not configured to run with this repo yet.

rht: I have reordered with isort and force pushed.

zulip/integrations/bridge_with_slack/run-slack-bridge
rht: I can't make a definitive answer to your question without testing how the code behaves.

timabbott:
Merged, thanks @rht!  Can you follow up with the original issue reporter to confirm whether this fixes it for them on their macOS system?

rht:
@chrisbobbe said it worked on his macOS system.

PR #700, opened 2021-06-10
twitter: remove requirement of two internal file and support multiple searches and users

### LINKED ISSUE #697: Duplicating same tweets when following more users with Twitter-bot
When following multiple users on Twitter with Twitter-bot using crontab, the same tweets will be will be importing again and again everytime crontab call the twitter-bot script. Using
```
*/5 * * * * /usr/local/share/zulip/integrations/twitter/twitter-bot --twitter-name="@TwitterUsername1" --config-file=~/zuliprc
*/5 * * * * /usr/local/share/zulip/integrations/twitter/twitter-bot --twitter-name="@TwitterUsername2" --config-file=~/zuliprc
*/5 * * * * /usr/local/share/zulip/integrations/twitter/twitter-bot --twitter-name="@TwitterUsername3" --config-file=~/zuliprc
```
will repeat to import last 5 tweets from two TwitterUsernames from the list every five minutes. Only one TwitterUsername will behave as expected and his tweets will not be duplicated.

@strifel proposed this solution:
You could try changing line 114 in /usr/local/share/zulip/integrations/twitter/twitter-bot from
```
CONFIGFILE_INTERNAL = os.path.expanduser("~/.zulip_twitteruserrc_fetchuser")
```
to
```
CONFIGFILE_INTERNAL = os.path.expanduser("~/.zulip_twitteruserrc_fetchuser_" + opts.twitter_name.replace('@', ''))
```

It solved the problem so I suggest to embed this code to twitter-bot if this don't affect using twitter-bot with only one following TwitterUser.

strifel:
This PR removes the necessity to have to internal config files in twitterbot, by using different keys for searches and twitter_names in the same file.
This does also fix #697 in the same way.
See https://chat.zulip.org/#narrow/stream/127-integrations/topic/Twitter/near/1198859 for discussions.

As I do not have access to the Twitter API I could not test this PR.

I could see an possible error at 
```
SINCE_ID_NAME = "since_id_" + opts.search_terms
```
as search_terms (or twitter_name) contains special characters like `,` or `@`


zulip/integrations/twitter/twitter-bot
```diff
@@ -43,7 +44,7 @@ that will process tweets every 5 minutes is:
 
 == Setting up Twitter authentications ==
 
-Run this on a personal or trusted machine, because your API key is
+Run this on a personal or trusted machine, because your APIx key is
```
timabbott, line 47: This is presumably a typo.

zulip/integrations/twitter/twitter-bot
```diff
@@ -106,10 +107,10 @@ if all([opts.search_terms, opts.twitter_name]):
     parser.error("You must only specify either a search term or a username.")
 if opts.search_terms:
     client_type = "ZulipTwitterSearch/"
-    CONFIGFILE_INTERNAL = os.path.expanduser("~/.zulip_twitterrc_fetchsearch")
+    SINCE_ID_NAME = "since_id_search_" + opts.search_terms
 elif opts.twitter_name:
     client_type = "ZulipTwitter/"
-    CONFIGFILE_INTERNAL = os.path.expanduser("~/.zulip_twitteruserrc_fetchuser")
+    SINCE_ID_NAME = "since_id_name_" + opts.twitter_name
```
timabbott, line 113: I was trying to think of a better way to serialize the parameters to a particular instances of the bot, but didn't come up with anything great.  One idea would be to have the configuration file support having a block for each search, and use the names defined there.  E.g. 

```
[twitter_search_foo]
search_terms = ...

[twitter_search_bar]
twitter_name = ...
```

zulip/integrations/twitter/twitter-bot
strifel: Yes it is.
I will remove the x again.

Thanks

PR #702, opened 2021-06-18
matrix: fix bounce-back when bot not on matrix.org

steadfasterX:
this requires having the matrix name properly set in matrix_bridge.conf (e.g. @zulip-bot:myhomserver.com)

### DIFF
--- a/zulip/integrations/bridge_with_matrix/matrix_bridge.py
+++ b/zulip/integrations/bridge_with_matrix/matrix_bridge.py
@@ -77,7 +77,7 @@ def matrix_to_zulip(
         """
         content = get_message_content_from_event(event, no_noise)
 
-        zulip_bot_user = "@{}:matrix.org".format(matrix_config["username"])
+        zulip_bot_user = "{}".format(matrix_config["username"])
         # We do this to identify the messages generated from Zulip -> Matrix
         # and we make sure we don't forward it again to the Zulip stream.
         not_from_zulip_bot = event["sender"] != zulip_bot_user


zulip/integrations/bridge_with_matrix/matrix_bridge.py
```diff
@@ -77,7 +77,7 @@ def _matrix_to_zulip(room: Any, event: Dict[str, Any]) -> None:
         """
         content = get_message_content_from_event(event, no_noise)
 
-        zulip_bot_user = "@{}:matrix.org".format(matrix_config["username"])
+        zulip_bot_user = "{}".format(matrix_config["username"])
```
timabbott, line 80: Do we want to check if there's a `:` in the string and use that to determine whether to append `:matrix.org`?  I don't know the Matrix world's conventions to know if they prefer to have everyone specify the full name everywhere or use a `:matrix.org` default, but we should probably follow whatever patterns they have.

steadfasterX: yea it might look like a "too wide" approach but the syntax of a mxid (matrix account id) is:

`@<username>:<homeserver.tld[:port]>`

so `@username` is clear. then a `colon` as a delimiter to the `homeserver`. the homeserver can have a _custom port_ optionally too. 

here some examples just for the homeserver part:

```
foo.com
matrix.foo.com
foo.com:8383
matrix.org
```
matrix.org is just 1 homeserver (the most known one ofc) but "the" matrix is not matrix.org. matrix is a decentral network where everyone can have its own homeserver and chats/rooms can be synced between. 

another approach would be to let the user not write the mxid but the username in one key and the homeserver and port in another like

```
[matrix]
username = foo
homeserver = foo.com:8111
```

but that is very unusal in the matrix world and I don't think thats needed. KISS ;)

TL;DR
I would never ever hard code `matrix.org`  anywhere there and let the format be as free as possible.

rht: The most unambiguous way is to just force user to set `matrixid` in the `[matrix]` section in the configuration file. So no need to guess if there is ':' or not.

rht: As in, we need to rename `username` field to `matrixid` for clarity.

steadfasterX: I would vote for `mxid` as that is is the official name for it :) 
https://matrix.org/faq/#what-is-a-mxid%3F

rht: `mxid` sounds good. You can change that in this PR, because if it were a separate PR, there would be a merge conflict.

timabbott: Migrating to `mxid` sounds good to me.

timabbott: (And I think the only reason to assume `matrix.org` is if that was a convention in the Matrix community; since it sounds like that isn't, we should not do so either).

rht: A lot of people self-host their Matrix instance.

rht: @steadfasterX any update?

rht: I will just do the `mxid` conversion myself. Also, this is not the only place where matrix.org is hardcoded. You need to grep for others.

rht: @steadfasterX  this is my change: https://github.com/rht/python-zulip-api/commit/aed94d1ee9cd119dc7ed184344f9c8a792d7c2c7. Feel free to pull --rebase from it.

I kept getting this error
```
  File "matrix_bridge.py", line 333, in main
    matrix_login(matrix_client, matrix_config)
  File "matrix_bridge.py", line 42, in matrix_login
    matrix_client.login_with_password(matrix_config["mxid"], matrix_config["password"])
  File "/home/rht/code/venv/lib/python3.8/site-packages/matrix_client/client.py", line 249, in login_with_password
    return self.login(username, password, limit, sync=True)
  File "/home/rht/code/venv/lib/python3.8/site-packages/matrix_client/client.py", line 280, in login
    self._sync()
  File "/home/rht/code/venv/lib/python3.8/site-packages/matrix_client/client.py", line 558, in _sync
    for presence_update in response['presence']['events']:
KeyError: 'presence'
```

How did you fix this?

zulip/integrations/bridge_with_matrix/matrix_bridge.py
```diff
@@ -77,7 +77,7 @@ def _matrix_to_zulip(room: Any, event: Dict[str, Any]) -> None:
         """
         content = get_message_content_from_event(event, no_noise)
 
-        zulip_bot_user = "@{}:matrix.org".format(matrix_config["username"])
```
rht, line 80: This needs to be just `zulip_bot_user = matrix_config["username"]`. No need for format.

zulip/integrations/bridge_with_matrix/matrix_bridge.py
steadfasterX: Sorry RL happened. There are several changes needed which i need to split from all my other changes and fixes first.

rht:
This PR can be closed. #734 was merged.

PR #705, opened 2021-06-28
zulip.Client.call_endpoint: Add retry_on_rate_limit_error.

ro-i:
Hi :)

I'd like to suggest a new option for the `call_endpoint` method of `zulip.Client`. It's something I added to [my bot](https://github.com/ro-i/tumcsbot/blob/0ca58024b83c8ec0353217c7fd89a984d1d4b675/src/tumcsbot/client.py#L62). @timabbott suggested to also consider this for the API, see https://chat.zulip.org/#narrow/stream/378-api-design/topic/Rate.20limits/near/1217048 for the discussion.

Short description of the patch:
If the `call_endpoint` method is called with the `retry_on_rate_limit_error` parameter set to true, wait and retry automatically on rate limit errors.


zulip/zulip/__init__.py
```diff
@@ -697,24 +697,41 @@ def call_endpoint(
...
+            result = self.do_api_query(
+                marshalled_request,
+                versioned_url,
+                method=method,
+                longpolling=longpolling,
+                files=files,
+                timeout=timeout,
+            )
+            if not retry_on_rate_limit_error or result["result"] == "success":
+                break
```
timabbott, line 724: I'd do this as two conditionals, first the `success` one, and then the other one.  We like to do this so that code coverage tools can tell you whether we have tests for each independent condition.

zulip/zulip/__init__.py
```diff
@@ -697,24 +697,41 @@ def call_endpoint(
...
+            elif "code" in result and result["code"] == "RATE_LIMIT_HIT":
+                secs = result["retry-after"]
+            elif "X-RateLimit-Reset" in result:
+                secs = float(result["X-RateLimit-Reset"])
```
timabbott, line 728: So, I think the `retry-after` property is there since the beginning of time for these errors; just not `code`. So I think you could do 
```
if 'retry-after' not result:
    break
 
wait_before_retry_secs = result["retry-after"]
logger.warning(...)
time.sleep(...)

ro-i:
@timabbott Thanks for the feedback! :+1: 
I updated the code (the layout of the if-statement with the closing bracket on its own line has been done by `./tools/lint` btw. ;) ).
And you would also think that `retry-after` has always been a float, not a string? I'm not sure how that has been handled by older Zulip versions. And I'm not sure what happened to the capitalized `Retry-After` I found [here](https://github.com/zulip/zulip/blob/master/zerver/tests/test_external.py#L121)...

PR #709, opened 2021-07-30
zulip_bots: Add a script for creating Zulip bots.

PIG208:
Following support to running bots from entry points in #708, we
implement this `create-zulip-bot` tool to simplify the process of
creating new bots. The user will be able to directly install the package
with pip and run the bot with `zulip-run-bot`, or use it to quickly set
up a git repository.

Note that the boilerplate generated by this script does not contain
`tests.py` yet. We need to figure out the right pattern for integrating
unittests for such packaged bots.


zulip_bots/zulip_bots/create_bot.py
```diff
@@ -0,0 +1,139 @@
...
+"""
+
+SETUP_TEMPLATE = """import {bot_name}
+from setuptools import find_packages, setup
+
+package_info = {{
+    "name": "{bot_name}",
+    "version": {bot_name}.__version__,
+    "entry_points": {{
+        "zulip_bots.registry": ["{bot_name}={bot_name}.{bot_name}"],
```
neiljp, line 23: Does this working assume the generating directory (default `.`) is actually equal to the bot name that is supplied? If so, it seems cleaner to make the base folder and the nested one relative to the current location?

PIG208: The name of the base directory doesn't matter. We assume that the nested directory and the bot module have the same name as the bot. This is the same practice we have for bots under the `zulip_bots/zulip_bots/bots` directory.

neiljp: Right you are, this was from a quick read rather than test.

neiljp:
@PIG208 This looks like it needs a rebase and tidy, but otherwise would progressing this be easier if the two commits were split between PRs? The first commit isn't necessary for the second, right?

zulip_bots/zulip_bots/create_bot.py
```diff
@@ -0,0 +1,139 @@
...
+def main() -> None:
+    args = parse_args()
+
+    handler_name = f'{args.bot.title().replace("_", "")}Handler'
+
+    bot_path = Path(args.output, args.bot)
+    bot_module_path = Path(bot_path, args.bot)
+
+    try:
+        os.mkdir(bot_path)
```
neiljp, line 109: Reading this again, is there a reason not to use the pathlib functions instead, if we're already using Path? That is, can we drop `import os`?

Issue #683, opened 2021-05-06
Add support for using ZULIP_FEATURE_LEVEL to determine how to encode parameters


timabbott:
As part of https://github.com/zulip/zulip/issues/18035

We should implement this as follows:
* A commit should make a `Client` object do a `GET /server_settings` and store the `feature_level` and server version as properties.
* We can then write functions to encode parameters differently depending on the feature level. for `update_user` as a sample case, in preparation for the API migration there.
* We can figure out a test suite strategy to ensure that these actually work across versions.  E.g. maybe we have CI to run the zulip/zulip `tools/test-api` tests for 2.1, 3.0, 4.0, and master against the Python bindings.
* Once we've figured out the infrastructure for this, we can make a master issue to make everything in our [API changelog](https://zulip.com/api/changelog) be supported across versions by this project.

@ligmitz FYI.

timabbott:
Also @msurfer @orientor @LoopThrough-i-j FYI, since if we implement this correctly, it would adjust and hopefully improve the workflow for making API changes.

LoopThrough-i-j:
Exactly something I was looking for and couldn't find. This could create a door for writing integration tests for the API clients and much more.

timabbott:
@PIG208 FYI as well; this is probably one of the more important features for us to add to the Python API bindings.

@neiljp also FYI, since this may have some utility for ZulipTerminal as well (if you can transparently handle some API upgrades).

neiljp:
We already fetch the feature-level and adjust some behavior based on it. It's not clear to me precisely what's involved here, but we can likely handle this via a `zulip` dependency upgrade or via feature-level conditionals?

Is this required (ie a ZT bug as of now) or optional (ie. ZT api migration, or subject->topic)?

timabbott:
The goal as stated here is just to have things written with the python-zulip-api transparently handle how to call endpoints (E.g. if we rename an endpoint or we change how a parameter to send is encoded).  I think ZT will still need to have all the logic for parsing event formats, etc., though it would certainly be reasonable to consider as a follow-up having a shared function here that converts any old-format events into newer-format events.

Relatedly, we probably want `python-zulip-api` to pass various `client_capabilities` by default?  Not sure.

timabbott:
[...]

PR #710, opened 2021-08-02
Use `ZULIP_FEATURE_LEVEL` to call enpoints ensuring compatibility

PIG208:
#683

I experimentally implement a decorator for this purpose.

With this PR, we can sort of "overload" the functions calling endpoints regarding `ZULIP_FEATURE_LEVEL`. However, I'm not pretty sure about what role `client_capabilities` would play here.


PIG208:
One possible limitation of this approach to be noted: the decorated function should have a function signature compatible with any other alternatives.

eeshangarg:
I haven't worked with the API bindings in a long time, so there is a lot of context I may be missing here. Having said that, I am really not sure if a decorator is necessarily the right idiom to use here due to what you said about function signatures having to be the same as the alternatives.

The freedom to be flexible and less stringent with function signatures seems like a good goal to have here, so I wonder if there is a simpler implementation approach here. Perhaps a dispatcher class of some sort that every call goes through before settling on which endpoint to call?

zulip/zulip/__init__.py
```diff
@@ -486,6 +486,11 @@ def __init__(
 
         self.has_connected = False
 
+        server_settings = self.get_server_settings()
+        self.zulip_version: Optional[str] = server_settings.get("zulip_version")
+        self.feature_level: int = server_settings.get("zulip_feature_level", 0)
```
timabbott, line 512: Do we want to call this `zulip_feature_level`, or just `feature_level`?  The `zulip_` prefix might be more appropriate for analogy with `zulip_version`.

zulip_botserver/tests/test_server.py
```diff
@@ -25,6 +25,11 @@ class MockLibModule:
         def handler_class(self) -> Any:
             return BotServerTests.MockMessageHandler()
 
+    def setUp(self) -> None:
+        super().setUp()
+        self.patch = mock.patch("zulip.Client.get_server_settings", return_value=mock.Mock())
+        self.patch.start()
```
timabbott, line 31: Can you add a comment explaining the motivation for this?

zulip/zulip/__init__.py
```diff
@@ -1320,6 +1320,14 @@ def reactivate_user_by_id(self, user_id: int) -> Dict[str, Any]:
...
         )
 
+    def update_user_by_id_compatible(self, user_id: int, **request: Any) -> Dict[str, Any]:
+
+        for key, value in request.items():
+            request[key] = json.dumps(value)
+
+        return self.call_endpoint(url=f"users/{user_id}", method="PATCH", request=request)
+
+    @compatible(update_user_by_id_compatible, 100, 100)
```
timabbott, line 1340: Let's use 10000 with a comment for this sort of demonstration value.

zulip/zulip/__init__.py
```diff
@@ -1320,6 +1320,14 @@ def reactivate_user_by_id(self, user_id: int) -> Dict[str, Any]:
             method="POST",
         )
 
+    def update_user_by_id_compatible(self, user_id: int, **request: Any) -> Dict[str, Any]:
```
timabbott, line 1323: I think it's worth having functions like this come with a block comment explaining the API change.  Potentially just copy-pasting the relevant section of the API documentation would be appropriate.

zulip/zulip/__init__.py
```diff
@@ -326,6 +328,25 @@ def validate_boolean_field(field: Optional[str]) -> Union[bool, None]:
         return None
 
 
+FuncT = Callable[..., object]
+
+
+def compatible(
```
timabbott, line 334: It'd be great to add a comment/docstring here.  I also think it might be appropriate to make the `min_feature_level` / `max_feature_level` parameters mandatory kwargs (i.e. `alternative: FuncT, *, max_feature_level: int=...`.

zulip/zulip/__init__.py
```diff
@@ -1322,6 +1322,16 @@ def reactivate_user_by_id(self, user_id: int) -> Dict[str, Any]:
...
 
+        """
+        Example usage:
+
+        >>> client.update_user_group_members({
+            'delete': [4, 8, 15],
+            'add': [16, 23, 42],
+        })
+        {'msg': '', 'result': 'success'}
+        """
```
timabbott, line 1333: This docstring seems misplaced?

timabbott:
@PIG208 thanks for working on this!  I think the decorator strategy is a good one.  Ideally, what we want to be doing is presenting the same/modern API in the Python bindings regardless of server version (where possible), and having the `@compatible` functions take care of rewriting the requests/responses to match the format used by the modern version.  For that reason, I don't think the requirement that the functions have a similar signature is likely to be problematic.

timabbott:
It might make sense to pull the "Zulip CI" piece to a separate PR that we can iterate on independently, if it isn't already.

PIG208:
Thank you for the feedback! I will get back to this when I get time.

andersk:
As discussed [in #api design](https://chat.zulip.org/#narrow/stream/378-api-design/topic/Compatible.20binding.20design), I think we should use simple `if` statements instead of complicating this with decorators.

zulip/zulip/__init__.py
PIG208: I think `feature_level` will be more concise and it corresponds to what we have in `version.py`

zulip/zulip/__init__.py
PIG208: Yes, I think we will turn to the `if` approach after the discussion.

zulip/zulip/__init__.py
PIG208: I think in any cases, we will need to have comments explaining why we will do a feature_level comparison. Otherwise the feature level just looks like a magic number.

zulip/zulip/__init__.py
```diff
@@ -1728,3 +1741,51 @@ def hash_util_decode(string: str) -> str:
...
+    def update_user_group_members(
+        self, user_group_id: int, group_data: Dict[str, Any]
+    ) -> Dict[str, Any]:
+        ...
+
+    @overload
+    def update_user_group_members(self, group_data: Dict[str, Any]) -> Dict[str, Any]:
+        ...
+
+    def update_user_group_members(self, *args: Any, **kwargs: Any) -> Dict[str, Any]:
```
andersk, line 1766: There is no reason to use an overload here, and even if you did need to use an overload, that would not be a reason to use the wildly type-unsafe `*args: Any, **kwargs: Any` signature.

```python
    def update_user_group_members(
        self, user_group_id: Optional[int] = None, group_data: Dict[str, Any]
    ) -> Dict[str, Any]:
```

zulip/zulip/__init__.py
```diff
@@ -1728,3 +1741,51 @@ def hash_util_decode(string: str) -> str:
...
+        """
+        # This interface was removed in 4d482e0ef30297f716885fd8246f4638a856ba3b
+        return self.call_endpoint(
+            url="realm/filters",
+            method="GET",
+        )
+
+
+if USE_LEGACY_INTERFACE:
+    Client = LegacyInterfaceClient  # type: ignore # We patch the original Client class to make it support older clients
```
andersk, line 1791: I’m confused—is this about older servers or older clients?

timabbott:
OK, we can rebase this now.

ligmitz:
I don't have a pull request yet. Can you brief me a little about the steps we are going to take so that I can quickly work a PR ?

PR #711, opened 2021-08-06
CI: Add test suites from zulip server.

PIG208:
WIP. Testing the test setup on Github.


timabbott:
@PIG208 this is awesome!  I see that we're failing CI with the 18.04/3.2 test case.

The problem is kinda subtle: the API documentation that comes with 3.2 expects an old version of the python-zulip-api library's interface.  I guess the bug is that we made a client-side API change that is not backwards-compatible, which was an intentional interface change, but does mean the 3.2 documentation isn't correct anymore.

If there was a magical way to do this, I guess what we want to promise to users and thus want to test is that the modern `master` API documentation, when run against the module `python-zulip-api` and an old server version will work correctly.  I doubt it's particularly feasible to do that, since the `test-api` suite requires a bunch of code from the current server version for test database setup (etc.).

I think the next best thing we can do is try to hack around it -- e.g. figuring out a way to have a bit of configuration that lives in `python-zulip-api` that provides some way to do a compatibility hook for the tests.  E.g. we have some mechanism through which if a certain environment variable is set, we rename a `LegacyInterfaceClient` class to `Client`, and that class overrides a few methods with thin wrappers to translate the interface for interface migrations.  E.g. for the `update_members_by_id` function, it'd be something like this:
```
class LegacyInterfaceClient(Client)
    def update_user_group_members(self, group_data: Dict[str, Any]) -> Dict[str, Any]:
        modern_group_data = group_data.copy()
        group_id = group_data['group_id']
        del modern_group_data['group_id']
        return super().update_user_group_members(group_id, modern_group_data)

if USE_LEGACY_INTERFACE:
     Client = LegacyInterfaceClient  # type: ignore
```

PIG208:
Originally, I was thinking of making the exposed function signature to be something like `def update_user_group_members(args, kwargs)` and let the decorator handle the actual signature. This would make it easier to let us make the api compatible with arbritrary signatures, bypassing the below restriction:

> One possible limitation of this approach to be noted: the decorated function should have a function signature compatible with any other alternatives.

However, type-checking (for dynamic function signature in terms of `FEATURE_LEVEL`) will be impossible with this approach.
Then I guess hacking around will be the optimal solution here. But probably having an option to let the `compatible` decorator forcibly turn the decorated modern api to a legacy interface (which does not type-check, but works during runtime) would be a nicer hack here, since we can reuse some functionality defined in #P710.

timabbott:
For type-checking, one can use `@override` to define a couple compatible approaches.

That said, I think I'm increasingly convinced we want the library to just have the modern interface, and to the extent that we need to do hacks to make CI work, those should be contained in the CI, via the `LegacyInterfaceClient` approach I suggested above.

PIG208:
This and #710 are ready for review.

zulip_botserver/tests/test_server.py
```diff
@@ -25,6 +25,13 @@ class MockLibModule:
         def handler_class(self) -> Any:
             return BotServerTests.MockMessageHandler()
 
+    def setUp(self) -> None:
+        # Since initializing Client invokes `get_server_settings` that fails in the test
+        # environment, we need to mock it to pretend that there exists a backend.
+        super().setUp()
+        self.patch = mock.patch("zulip.Client.get_server_settings", return_value=mock.Mock())
+        self.patch.start()
```
timabbott, line 33: It would likely be better for this to return fixture content that we fetch once from a real Zulip server and store in a file in the codebase.

zulip/zulip/__init__.py
```diff
@@ -486,6 +486,11 @@ def __init__(
 
         self.has_connected = False
 
+        server_settings = self.get_server_settings()
+        self.zulip_version: Optional[str] = server_settings.get("zulip_version")
+        self.feature_level: int = server_settings.get("zulip_feature_level", 0)
+        assert self.zulip_version is not None
```
timabbott, line 493: Does `zulip_version` need to use `.get()`?  I think `zulip_version` at least has been present since essentially the beginning of time, and it should probably be an error for it to not be present.

zulip/zulip/__init__.py
```diff
@@ -1740,3 +1741,51 @@ def hash_util_decode(string: str) -> str:
...
+        """
+        # This interface was removed in 4d482e0ef30297f716885fd8246f4638a856ba3b
+        return self.call_endpoint(
+            url="realm/filters",
+            method="GET",
+        )
+
+
+if USE_LEGACY_INTERFACE:
+    Client = LegacyInterfaceClient  # type: ignore # We patch the original Client class to make it support older clients
```
timabbott, line 1791: So, I'm thinking that we may need multiple legacy interface versions for different Zulip major releases if we want the tests to run against multiple versions; maybe we want to encode this as `LEGACY_INTERFACE_VERSION==3`, with the versions being the Zulip major release versions (E.g. 4 or 2.1 when the numbering scheme was different)?

timabbott:
This looks great @PIG208!  I posted a few small comments, but this looks like something we'll be very happy to merge.

Does it make sense to skip the "api: Make `update_user_by_id` compatible. " commit for this PR (moving it to a separate one) so that we can merge this PR fully?  And then we can just open a new PR with just that commit once this is merged.

PIG208:
Yes, I rebase this PR onto the feature level branch just to test it out.

PIG208:
OK. I will come back to this on weekend.

timabbott:
This looks great; I did some updates to the documentation and set the `legacy_client_interface` parameter to `4.0` for the last two tests.

timabbott:
Merged, thanks @PIG208!

PR #712, opened 2021-08-16
zulip: Deprecate methods whose names don't match with OperationIDs an…

MSurfer20:
…d add newer methods.


MSurfer20:
I have manually audited, and these are all the bindings whose names don't match the operationIDs as defined in `zulip.yaml`.

timabbott:
Thanks for doing this audit @MSurfer20!  I started https://chat.zulip.org/#narrow/stream/378-api-design/topic/python.20bindings.20function.20names/near/1255079 for discussion.

zulip/zulip/__init__.py
```diff
@@ -951,6 +951,10 @@ def remove_reaction(self, reaction_data: Dict[str, Any]) -> Dict[str, Any]:
         )
 
     def get_realm_emoji(self) -> Dict[str, Any]:
+        logger.warning("get_realm_emoji() is deprecated." " Please use get_custom_emoji() instead.")
```
andersk, line 954: Deprecation warnings should be flagged with [`warnings.warn`](https://docs.python.org/3/library/warnings.html#warnings.warn).

```suggestion
        warnings.warn(
            "get_realm_emoji() is deprecated. Please use get_custom_emoji() instead.",
            DeprecationWarning,
            stacklevel=2,
        )
```

zulip/zulip/__init__.py
```diff
@@ -1078,6 +1119,12 @@ def reorder_realm_profile_fields(self, **request: Any) -> Dict[str, Any]:
         )
 
     def update_realm_profile_field(self, field_id: int, **request: Any) -> Dict[str, Any]:
+        logger.warning(
+            "update_realm_profile_field() is deprecated." " Please use update_linkifier() instead."
```
gnprice, line 1123: Is this the right pairing?

PR #718, opened 2021-08-25
zephyr_mirror_backend: Add missing parens for calling lower()

andersk:
Found by trying to enable mypy `strict_equality`. @timabbott FYI.


timabbott:
Merged, thanks @andersk!  Probably won't try to deploy until I'm home.

PR #720, opened 2021-08-26
finder: Support older importlib.metadata API

andersk:


timabbott:
The CI failure seems to be on all the older Python versions, so likely real.

andersk:
Yeah, this is still WIP.

timabbott:
I think this is no longer relevant, as those Python versions are EOL.

PR #722, opened 2021-09-17
Slack bridge: Explicitly require aiohttp.

rht:
It is required by the Slack library, but strangely not installed. So we
explicitly specify it.


timabbott:
Merged, thanks @rht!

andersk:
This was already fixed upstream (slackapi/python-slack-sdk#1000), why don’t we upgrade instead?

PR #730, opened 2021-10-18
Slack bridge: Bump slack-sdk to 3.11.2.

rht:
We also upgrade the RTM client API from v1 to v2. This is so that we no longer require aiohttp. If we use v1, it would still require aiohttp.


timabbott:
Merged, thanks @rht !

PR #723, opened 2021-09-21
[WIP] Enhanced matrix bridge.

ro-i:
Hi :)

As discussed [here](https://chat.zulip.org/#narrow/stream/127-integrations/topic/Matrix.20.3C-.3E.20Zulip.20Bridge), I'm now creating the pull request for the new Matrix <-> Zulip bridge.
See [here](https://chat.zulip.org/#narrow/stream/127-integrations/topic/Use.20upload_file.20with.20in-memory.20IO-object) for further discussions about uploading an in-memory file-object to a Zulip server.

Feedback would be greatly appreciated! :)

PR #734, opened 2021-11-02
Clarify Matrix bridge terminology.

rht:
#702 hasn't been updated for a long time. #723 will take a while to finish. This PR remakes #702 so that the current bridge config can be more easily understood.


timabbott:
Merged, thanks @rht!

PR #736, opened 2021-11-19
bots: Migrate script name from zulip-terminal to zulip-bot-shell.

neiljp:
This change is intended to reduce confusion between zulip-bot-shell
(test bots interactively without a server) and the zulip/zulip-terminal
project and its associated command (zulip-term).

Latest reference at https://chat.zulip.org/#narrow/stream/127-integrations/topic/Migrating.20bot.20zulip-terminal.20script

Docs change in main repo at zulip/zulip#20310


timabbott:
Merged, thanks @neiljp!  

Do you think we should also rename the actual directories in this codebase?

PR #737, opened 2021-11-20
api: Add subset of Literal types used in zulip-terminal

neiljp:
See https://chat.zulip.org/#narrow/stream/378-api-design/topic/python.20bindings.20and.20constants

The first commit is useful as it integrates with the API directly; the second doesn't integrate with the API as the method parameter data types are unspecified dicts right now (see discussion).

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -28,6 +28,7 @@ from typing import (
 
 import distro
 import requests
+from typing_extensions import Literal
 
 __version__ = "0.8.1"
 
@@ -43,6 +44,38 @@ assert LooseVersion(requests.__version__) >= LooseVersion("0.12.1")
 requests_json_is_function = callable(requests.Response.json)
 
 API_VERSTRING = "v1/"
+
+# An optional parameter to `move_topic` and `update_message` actions
+# See eg. https://zulip.com/api/update-message#parameter-propagate_mode
+EditPropagateMode = Literal["change_one", "change_all", "change_later"]
+
+# Generally a `reaction_type` is present whenever an emoji is specified:
+# - Optional parameters to actions: `add_reaction`, `remove_reaction`
+# - Events: "user_status", "reaction", "message", "update_message"
+# - Inside each reaction in the `reactions` field of returned message objects.
+EmojiType = Literal["realm_emoji", "unicode_emoji", "zulip_extra_emoji"]
+
+# Message flags which may be directly modified by the current user:
+# - Updated by `update_message_flags` (and for the `read` flag, also
+#   the `mark_all_as_read`, `mark_stream_as_read`, and
+#   `mark_topic_as_read` actions.
+# - User is notified of changes via `update_message_flags` events.
+# See subset of https://zulip.com/api/update-message-flags#available-flags
+ModifiableMessageFlag = Literal["read", "starred", "collapsed"]
+
+# All possible message flags.
+# - Generally present in `flags` object of returned message objects.
+# - User is notified of changes via "update_message_flags" and `update_message`
+#   events. The latter is important for clients to learn when a message is
+#   edited to mention the current user or contain an alert word.
+# See https://zulip.com/api/update-message-flags#available-flags
+MessageFlag = Literal[
+    ModifiableMessageFlag,
+    "mentioned",
+    "wildcard_mentioned",
+    "has_alert_word",
+    "historical",
+]
 
 
 class CountingBackoff:
@@ -1638,7 +1671,7 @@ class Client:
         topic: str,
         new_topic: Optional[str] = None,
         message_id: Optional[int] = None,
-        propagate_mode: str = "change_all",
+        propagate_mode: EditPropagateMode = "change_all",
         notify_old_topic: bool = True,
         notify_new_topic: bool = True,
     ) -> Dict[str, Any]:
--- a/zulip/setup.py
+++ b/zulip/setup.py
@@ -71,6 +71,7 @@ setuptools_info = dict(
         "matrix_client",
         "distro",
         "click",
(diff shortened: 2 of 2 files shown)


zulip/zulip/__init__.py
```diff
@@ -1634,7 +1649,7 @@ def move_topic(
         topic: str,
         new_topic: Optional[str] = None,
         message_id: Optional[int] = None,
-        propagate_mode: str = "change_all",
+        propagate_mode: EditPropagateMode = "change_all",
```
gnprice, line 1652: Can this perhaps just have the `Literal` type directly, inline? That'd be ideal from the perspective of someone looking at this function signature and wanting to understand how to use it.

If that's inconvenient for some reason, then a one-line comment on the `EditPropagateMode` definition, mentioning it's for this method, would be helpful for the reader looking at that definition and wondering what it means.

zulip/zulip/__init__.py
```diff
@@ -44,6 +45,20 @@
...
 
+EditPropagateMode = Literal["change_one", "change_all", "change_later"]
+
+EmojiType = Literal["realm_emoji", "unicode_emoji", "zulip_extra_emoji"]
+
+# Limited flags which could be present in update_message_flag events.
+ModifiableMessageFlag = Literal["read", "starred", "collapsed"]
+MessageFlag = Literal[
+    ModifiableMessageFlag,
+    "mentioned",
```
gnprice, line 56: For each of these, I'd encourage having a one- or two-line comment mentioning what part of the API it's for. Just a link to the relevant bit of API docs may be enough -- that way there's no need to try to duplicate any prose describing the semantics.

gnprice (commented):
Cool! Glad to see more types appearing here. Small comments below.

zulip/zulip/__init__.py
PIG208: These Literal types look a bit vague regarding how exactly the users are supposed to use them. For update_message_flags specifically, do you think that defining a TypedDict making use of these types will be a good idea?

zulip/zulip/__init__.py
PIG208: I agree, since we aren't using this type anywhere else.

PIG208 (commented):
Another thing that might be valuable to add is to have mypy check these type bindings. Right now we don't really have a way to guarantee that everything typed here is "correct" regarding the documentation, especially for those literal types that are added without being used here.

zulip/zulip/__init__.py
neiljp: I think having comments relating it to which API calls would be useful, and for other types too, when we introduce those.

For discussion on whether it's worth introducing them as named types, see the discussion thread for details - but in this case the argument is that it's useful to track the proposed mode in an application before sending it, and also mapping it to user-displayed strings.

zulip/zulip/__init__.py
neiljp: I referred to the commenting aspect in another response above.

Currently these types aren't directly used in the API, at least via other types, so we'd have to use TypedDicts or migrate the API to allow alternative function signatures.

neiljp:
As per the discussion in #**api design**, this is now updated and ready for review.

It looks like we'll squash commits when this is done, so I just dropped the new comment updates in a separate commit for now.

zulip/zulip/__init__.py
```diff
@@ -44,6 +45,31 @@
...
+# All possible message flags
+# - generally present in `flags` object of returned message objects
+# - notified to the user via "update_message_flag" events
+MessageFlag = Literal[
+    ModifiableMessageFlag,
+    "mentioned",
+    "wildcard_mentioned",
+    "has_alert_word",
+    "historical",
+]
```
gnprice, line 71: I'd encourage you to add a link here to where these are described in the API docs:
https://zulip.com/api/update-message-flags#available-flags

That's good for the reader looking at this library and trying to understand what the semantics are -- in particular, what these flags mean -- because it helps them find where we describe that.

It's also good for a future reader seeking to update the library (perhaps yourself) by providing a quick way to compare against the then-current docs and see if anything's changed.

We try to make a similar practice in the API bindings within zulip-mobile. Discussion here:
https://github.com/zulip/zulip-mobile/blob/main/docs/style.md#api-doc-link

zulip/zulip/__init__.py
```diff
@@ -44,6 +45,31 @@
...
+# Generally a `reaction_type` parameter whenever an emoji is specified:
+# - optional parameters to actions: `add_reaction`, `remove_reaction`
+# - events: "user_status", "reaction", "message", "update_message"
+# - generally inside each reaction in the `reactions` field of returned message objects
+EmojiType = Literal["realm_emoji", "unicode_emoji", "zulip_extra_emoji"]
+
+# Personally modifiable message flags:
+# - updated by `update_message_flags`, `mark_all_as_read`, `mark_stream_as_read`, `mark_topic_as_read` actions
+# - notified to the user via "update_message_flag" events
+ModifiableMessageFlag = Literal["read", "starred", "collapsed"]
```
gnprice, line 60: This part would actually especially benefit from a link to the API docs:
https://zulip.com/api/update-message-flags#available-flags

because it's otherwise a little bit mysterious what "personally modifiable" means, but the prose in that table makes it clear: these are the flags that can be changed by the user directly in an update-message-flags request.

gnprice (commented):
Thanks! This looks quite helpful. A pair of quick further comments below.

neiljp:
@gnprice Thanks for feedback; pushed an update.

gnprice (approved):
Looks great! Thanks for adding these.

timabbott:
Merged after a bit of copy-editing, huge thanks for doing this @neiljp!

PR #738, opened 2021-11-20
bots: Migrate source file for zulip-bot-shell to similar name.

neiljp:
Follow-up to #736 based on comment by Tim.

On noticing an exclusion of the file for type-checking, added a prep commit to do this, though based on this `zulip-bot-shell` doesn't cover the cases like the standard run script does, which we should unify to improve consistency and deduplicate.


zulip_bots/zulip_bots/bot_shell.py
```diff
@@ -32,15 +32,21 @@ def parse_args():
     return args
 
 
-def main():
+def main() -> None:
     args = parse_args()
 
-    bot_path, bot_name = resolve_bot_path(args.bot)
+    # NOTE: Use of only this implies bots from eg. registry cannot be explored in this way
```
timabbott, line 38: Let's open an issue for fixing this to support everything that the other bot interfaces support.

timabbott:
Merged, thanks @neiljp!

PR #739, opened 2021-11-28
Slack bridge: Implement multiple channels bridges.

rht:


rht:
@PIG208 thank you for the review!

zulip/integrations/bridge_with_slack/run-slack-bridge
```diff
@@ -37,14 +43,17 @@ class SlackBridge:
         self.zulip_config = config["zulip"]
         self.slack_config = config["slack"]
 
+        self.slack_to_zulip_map = dict(config["pairs"])
```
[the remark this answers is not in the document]
rht: The only use seems to be to ensure that `config["pairs"]` is indeed a dict. It is unlikely that people will screw up with the syntax and schema if they modify from the template config, so I'm removing this casting.

rht:
I have implemented the requests in the 3rd commit.

zulip/integrations/bridge_with_slack/run-slack-bridge
```diff
@@ -18,23 +18,25 @@ ZULIP_MESSAGE_TEMPLATE = "**{username}**: {message}"
 SLACK_MESSAGE_TEMPLATE = "<{username}> {message}"
 
 
-def check_zulip_message_validity(
+def get_zulip_message_slack_channel(
     msg: Dict[str, Any], zulip_to_slack_map: Dict[str, Any], bot_email: str
-) -> Tuple[bool, str]:
+) -> Union[str, None]:
```
PIG208, line 23: We can use `Optional[str]` here.

PIG208 (commented):
Except for a minor issue and the CI failure, it LGTM! Thanks.

zulip/integrations/bridge_with_slack/run-slack-bridge
rht: I intentionally didn't use `Optional` because this is different from the situation with `arg=None` in function arguments. It'd be better with `str | None`, but this is not supported in Python <3.10 unless there is `from __future__ import annotations`. See https://peps.python.org/pep-0604/.

zulip/integrations/bridge_with_slack/run-slack-bridge
```diff
@@ -18,23 +18,25 @@ ZULIP_MESSAGE_TEMPLATE = "**{username}**: {message}"
 SLACK_MESSAGE_TEMPLATE = "<{username}> {message}"
 
 
-def check_zulip_message_validity(
+def get_zulip_message_slack_channel(
```
neiljp, line 21: This may be a difference in how (human) languages name variables, but I'd choose something closer to `slack_channel_for_zulip_message`.

zulip/integrations/bridge_with_slack/run-slack-bridge
```diff
@@ -18,23 +18,25 @@ ZULIP_MESSAGE_TEMPLATE = "**{username}**: {message}"
...
+) -> Union[str, None]:
     is_a_stream = msg["type"] == "stream"
     if not is_a_stream:
-        return False, ""
+        return None
 
     stream_name = msg["display_recipient"]
     topic_name = msg["subject"]
-    stream_topic = stream_name + "-" + topic_name
+    stream_topic = f"{stream_name}-{topic_name}"
```
neiljp, line 30: The f-string formatting is an improvement, but reading this I wondered if you'd considered using a stream-topic tuple in your dicts instead?

zulip/integrations/bridge_with_slack/run-slack-bridge
```diff
@@ -142,7 +142,8 @@ if __name__ == "__main__":
         exit(1)
 
     print("Starting slack mirroring bot")
-    print("MAKE SURE THE BOT IS SUBSCRIBED TO THE RELEVANT ZULIP STREAM")
+    print("MAKE SURE THE BOT IS SUBSCRIBED TO THE RELEVANT ZULIP STREAM(S)!")
+    print("MAKE SURE THE BOT IS SUBSCRIBED TO THE RELEVANT SLACK CHANNEL(S)!")
```
neiljp, line 148: Good to include, but I think as a user, seeing these two lines 'shouting' at me with a lot of common text, I might not notice that they're subtly different. Maybe an (expanded) formatting on three lines relating to 1) common subscribe text 2) zulip stream(s) 3) slack channel(s) ?

zulip/integrations/bridge_with_slack/bridge_with_slack_config.py
```diff
@@ -3,12 +3,17 @@
...
-        "stream": "test here",
-        "topic": "<- slack-bridge",
     },
     "slack": {
         "username": "slack_username",
         "token": "xoxb-your-slack-token",
-        "channel": "C5Z5N7R8A -- must be channel id",
+    },
+    # Mapping between Slack channels and Zulip stream-topic's.
+    # You can specify multiple pairs.
```
neiljp, line 12: I'd be inclined to give a 'multiple pair' as an example, so the format is obvious. Or maybe more than one "pairs" section, with some commented out, as alternative examples.

Then maybe even multiple examples of how it could be used. eg. multiple slack channels -> many topics in one stream, etc

neiljp (commented):
@rht Always interested to see more bridge extensions :+1:

The other review pushed it onto my radar, this is only from the perspective of reading the code.

zulip/integrations/bridge_with_slack/run-slack-bridge
rht: OK I will just do what is done in zulip/zulip.

zulip/integrations/bridge_with_slack/run-slack-bridge
rht: I have combined the 2 lines into 1 sentence.

zulip/integrations/bridge_with_slack/run-slack-bridge
rht: Converted to tuple.

rht:
I have implemented the requests and tidied the commits.

neiljp:
@rht I'm not sure how much detail is expected here before merging, but other than the checks above (one needing fixing - static-analysis), has this been tested and if so how?

rht:
The static analysis test has been fixed.

>  has this been tested and if so how?

I tested this on a live Slack workspace and CZO. I have retested just now, just in case if the refactor causes any regression.
See https://chat.zulip.org/#narrow/stream/7-test-here/topic/.3C-.20slack-bridge.20general and https://chat.zulip.org/#narrow/stream/7-test-here/topic/.3C-.20slack-bridge.20general

zulip/integrations/bridge_with_slack/bridge_with_slack_config.py
```diff
@@ -3,12 +3,17 @@
...
+    "pairs": {
```
timabbott, line 13: Maybe call this `channel_mapping`?

zulip/integrations/bridge_with_slack/bridge_with_slack_config.py
```diff
@@ -3,12 +3,17 @@
...
+        "C5Z5N7R8A -- <Slack channel; must be channel id>": {
```
timabbott, line 14: Let's just have `C5Z5N7R8A` here, and document what it should be in a comment on the line above; I think there's a good chance some users will have `-- <` in their value otherwise :).

zulip/integrations/bridge_with_slack/bridge_with_slack_config.py
```diff
@@ -3,12 +3,17 @@
...
+            "stream": "test here",
+            "topic": "<- slack-bridge",
```
timabbott, line 16: Is there a way to specify "Use the slack channel name as the topic" here?

zulip/integrations/bridge_with_slack/bridge_with_slack_config.py
rht: Fixed.

zulip/integrations/bridge_with_slack/bridge_with_slack_config.py
rht: Fixed.

zulip/integrations/bridge_with_slack/bridge_with_slack_config.py
rht: > Could be nice to support substitution of the Slack channel name with a syntax like "slack: {channel_name}.

For a few number of pairs, users can specify that pattern by themselves explicitly. Adding this feature adds more complication to what the config can be, and requires more documentation

I haven't encountered >40 Slack channels pairing use case. The original use case is to bridge Julia Slack to Julia Zulip for content preservation, and they wanted to do so for only a few of the channels they are comfortable with being preserved. If this is needed soon, I can do it a follow-up PR.

zulip/integrations/bridge_with_slack/bridge_with_slack_config.py
```diff
@@ -3,12 +3,20 @@
...
     "slack": {
         "username": "slack_username",
         "token": "xoxb-your-slack-token",
-        "channel": "C5Z5N7R8A -- must be channel id",
+    },
+    # Mapping between Slack channels and Zulip stream-topic's.
+    # You can specify multiple pairs.
+    "channel_mapping": {
+        # Slack channel; must be channel ID
+        "C5Z5N7R8A>": {
```
timabbott, line 15: Is the `>` here part of the Slack channel name? I wouldn't have expected that to be present.

rht: Fixed. I probably copied that from a `<C5Z5N7R8A>` string and accidentally included it.

timabbott:
Merged, thanks @rht!

Issue #483, opened 2018-10-26
twisted support?


glyph:
I'd love a version of `call_on_each_event` that would not block, and instead register a callable with a reactor.

This could probably be combined with fairly general / framework agnostic async support by having the existing implementation be converted into a coroutine which `await`s different kinds of things (`treq` requests in the Twisted case, synchronously-returning coroutines that just call requests in the default / existing case, `aiohttp` Futures for asyncio).

neiljp:
Would it be accurate to describe this as asking for non-blocking support in the python API?

timabbott:
I think @kcr's https://github.com/1ts-org/snipe/blob/master/snipe/zulip.py#L127 has some work that could be borrowed/extracted for this (since `snipe` is a Python cross-protocol IM client that uses asyncio).

glyph:
> Would it be accurate to describe this as asking for non-blocking support in the python API?

Yep; any form of modern asyncio/tornado/Twisted support could ultimately be converted into this, and could therefore close this ticket.

PR #735, opened 2021-11-07
zulip: Reraise exceptions in do_api_query. 

eeshangarg:
There are cases where the call to get_server_settings returns a JSON response with a connection error. In the case of such an unsuccessful response, we should raise an appropriate exception instead of parsing the response as though it was successful.

@timabbott Could you please take a look at this? Not sure if this is the right fix. Also, wondering if I should put a test for this somewhere in `zulip_botserver`?

Issue #672, opened 2021-04-02
Raise exceptions for `{"result": "error"}` in `zulip` apis.


LoopThrough-i-j:
Currently, `zulip` APIs don't handle `{"result": "error"}` and sends send back comes from the server. We would want to handle these errors by raising exceptions in the API layer whenever there is a `{"result": "error"}`.

alexzw7:
@zulipbot claim

zulipbot:
Hello @alexzw7!

Thanks for your interest in Zulip! You have attempted to claim an issue without the labels "help wanted", "good first issue". Since you're a new contributor, you can only claim and submit pull requests for issues with the [help wanted](https://github.com/zulip/python-zulip-api/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22help+wanted%22) or [good first issue](https://github.com/zulip/python-zulip-api/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22good+first+issue%22) labels.

If this is your first time here, we recommend reading our [guide for new contributors](https://zulip.readthedocs.io/en/latest/overview/contributing.html) before getting started.

alexzw7:
@zulipbot add "good first issue"

alexzw7:
@zulipbot claim

zulipbot:
Welcome to Zulip, @alexzw7! We just sent you an invite to collaborate on this repository at https://github.com/zulip/python-zulip-api/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

LoopThrough-i-j:
@zulipbot remove "good first issue"

PIG208:
A brief note for this one here: we might need to add a common exception class called `BotError` and specify a set of sub errors like AccessDenied etc. (we need to discuss the naming). I think we can have a protocol that can be shared with embedded bots as well. 
[...]

PR #740, opened 2021-12-17
asyncio API and discord mirror (seeking feedback)

dehnert:
This PR adds support for an asyncio Zulip API (#483) and a discord mirroring script using it. It's definitely not ready to merge, but I think it'd be useful to have somebody look it over and provide some more feedback before I get it mergeable. The async client is mostly a pretty direct port of the original client, though obviously some of the lower-level functions had more changes.

So things I'd like guidance on / places that seem suboptimal that maybe there's a better solution for:
* As you can see, it hasn't been properly rebased. I ported a big chunk of it in one go, and then made some tweaks (eg, the iterator version of `call_on_each_event`, `RandomExponentialBackoff`) in separate commits, and then various later commits cleaning up lint errors. What degree of commit splitting seems useful? Presumably the lint fixup commits should be merged to be earlier, but is it useful to leave any of the early commits separately, or should I just have one "add an async API" commit?
* Ditto, but for the Discord mirror -- I feel a bit better with the current "first the actual mirror, then the docs and feedback" situation, but I could combine things. I may also add some more features soon-ish, and it'd be good to know if I should squash those in too.
* I'm not super clear on what the typing should be like for async functions -- is `async def call_endpoint(...) -> Awaitable[Dict[str, Any]]:` correct? The `Awaitable` feels sorta unwieldy and feels a little like it should be inferable from the `async`, but maybe that's the right thing?
* `mypy` doesn't like async RandomExponentialBackoff because `fail` changes from a function returning `None`, to a coroutine returning `None`. It seems to work, and seems a pity to duplicate `CountingBackoff`, but I'm not sure there's a good way to appease the type-checker (or if there's a subtle error).
* Does the existing API client have tests? The async API is very similar, so it's possible that any existing tests could be trivially or programmatically converted to run against the async API.
* Relatedly, *most* of the async API is basically "take the sync API, add `async` before `def`, and `await` after `return`". It's conceivable there's a good way to automate this? I didn't bother to, but I also didn't bother converting all the methods...
* I pretty directly ported `do_api_query`, which has a lot of error handling code. I tried to update it appropriately, but haven't tested all the ways things could fail. It's possible it makes more sense to delete a bunch of the error handling, and re-add it as people discover issues, rather than have untested error handling that I don't especially understand? Dunno.


timabbott:
@andersk would you be up for doing the review on this?

andersk:
Not sure if I’ll have time for anything like a full review soon, but some quick thoughts:

* An `async def` should not be annotated `-> Awaitable[…]`. The purpose of `Awaitable` is for typing *non*-`async` definitions that will be `await`ed ([documentation](https://mypy.readthedocs.io/en/stable/more_types.html#typing-async-await)).
  ```python
  async def a() -> int:
      return 1

  b: Callable[[], Awaitable[int]] = a

  def c() -> Awaitable[int]:
      return a()
  ```

* The new `aiohttp` requirement needs to be declared.

* Let’s not propagate the mistake from the sync API of returning server errors as if they were normal values; those should raise exceptions. See
  * #672
  * #735

* `asynch` is a weird name. I know this is a workaround for `async` being a keyword, but consider `aio` instead.

* I’m not excited about the maintenance cost of duplicating all our functions between the sync and async APIs. We should think about whether we want to deprecate the sync API and/or replace it with a wrapper around the async API.

merlinz01:
I would be interested in an async version of the client to improve overall performance in an async server environment.

Have you thought about generating both sync and async clients based on [the OpenAPI specification](https://github.com/zulip/zulip/blob/main/zerver/openapi/zulip.yaml)? Although I don't have experience with doing so, I have heard it recommended as a best practice.

timabbott:
@dehnert is this ready for review? You don't appear to have posted a comment when pushing to this branch after @andersk's comments, which results in GitHub not having notified us about your push. In any case, it'd be great to understand whether you believe you've addressed the feedback.

andersk:
No, we were notified about the push. It added two Discord-related commits and addressed none of my comments.

PR #741, opened 2022-01-05
tools: Delete the release-packages script.

eeshangarg:
This script has been outdated for a long time now and has been
obsoleted by some recent changes in how packages should be built
and distributed. In general, the release process is now too
complicated to automate, so we are better off just making the
release manually.


eeshangarg:
Merging after builds passed.

PR #743, opened 2022-01-13
zulip: Fix principals default for remove_subscriptions

andersk:
The documented API for `DELETE /api/v1/users/me/subscriptions` is that `principals` should be omitted to remove a subscription for the calling user.  A call with `principals=[]` should have a different meaning, but a server bug currently conflates this with a call omitting `principals`. Avoid relying on this bug.


timabbott:
Merged, thanks @andersk!

PR #744, opened 2022-02-10
Remove legacy distutils support

andersk:


timabbott:
Merged, thanks @andersk!

PR #747, opened 2022-02-24
add --topic arg to rss-bot script

dimisjim:


zulip/integrations/rss/rss-bot
```diff
@@ -51,6 +51,13 @@ stream every 5 minutes is:
 parser = zulip.add_default_arguments(
     argparse.ArgumentParser(usage)
 )  # type: argparse.ArgumentParser
+parser.add_argument(
+    "--topic",
+    dest="topic",
+    help="The Stream Topic to which to send RSS messages.",
+    default="stream events",
```
neiljp, line 58: This changes the default behavior which currently exists; is there a reason for this?

zulip/integrations/rss/rss-bot
```diff
@@ -241,7 +248,7 @@ for feed_url in feed_urls:
             # entries in reverse chronological order.
             break
 
-        feed_name = data.feed.title or feed_url  # type: str
+        feed_name = opts.topic
```
neiljp, line 251: While this change sets the topic, this doesn't preserve the existing behavior, when a fixed topic is not specified.

neiljp (commented):
@dimisjim Thanks for submitting this feature :+1:

I expect existing users could benefit from this, but would also appreciate the option of retaining the existing behavior.

In addition, it would be useful if you could adjust the documentation to take into account any new option.

zulip/integrations/rss/rss-bot
dimisjim: should be addressed via: https://github.com/zulip/python-zulip-api/pull/747/commits/4de095269de917feb2409c54bd5171bcbcd8f79c

zulip/integrations/rss/rss-bot
dimisjim: I've removed the default value via: https://github.com/zulip/python-zulip-api/pull/747/commits/4de095269de917feb2409c54bd5171bcbcd8f79c

Is my assumption correct that it will now fallback to `data.feed.title or feed_url` if no --topic arg is specified?

zulip/integrations/rss/rss-bot
```diff
@@ -42,6 +42,10 @@ You can customize the location on the feed file and recipient stream, e.g.:
 
 /usr/local/share/zulip/integrations/rss/rss-bot --feed-file=/path/to/my-feeds --stream=my-rss-stream
 
+optionally, you can also specify the name of the stream topic, e.g.:
+
+/usr/local/share/zulip/integrations/rss/rss-bot --feed-file=/path/to/my-feeds --stream=my-rss-stream --topic=my-topic-name
```
timabbott, line 47: Maybe specify an example topic containing spaces, e.g. `--topic="topic name"`, since that's more typical naming style.

zulip/integrations/rss/rss-bot
```diff
@@ -248,7 +251,7 @@ for feed_url in feed_urls:
             # entries in reverse chronological order.
             break
 
-        feed_name = opts.topic
+        feed_name = opts.topic or data.feed.title or feed_url  # type: str
```
timabbott, line 254: I would expect `feed_url` to often be too long; do we want to clamp too long topics here in the client code?

zulip/integrations/rss/rss-bot
dimisjim: indeed: https://github.com/zulip/python-zulip-api/pull/747/commits/3574ed6493c00e12933c8ccf7256e2d5228c8f22

zulip/integrations/rss/rss-bot
dimisjim: not sure what your suggestion is here really.

`data.feed.title or feed_url` was already here before my PR

i-ky:
The feature has been implemented in #785.

Issue #748, opened 2022-03-09
Cert Bundle might also be a directory


e7p:
When using the requests API with the `certificate` parameter, it is also possible to enter e.g. `/etc/ssl/certs` as a certificate bundle and it will result in verifying against all certificates lying in this path.

Though, I would recommend also including `os.path.isdir()` in the statement.

https://github.com/zulip/python-zulip-api/blob/e6dff1b5b6a9c6851d6c6699c2c52f345a4f486e/zulip/zulip/__init__.py#L492

e7p:
Can be linked to #749

PR #749, opened 2022-03-09
Cert Bundle might also be a directory

e7p:
So I changed the cert bundle verification to also support supplying other paths to the requests API by using `exists` instead of `isfile`. Now for me the zulip API works again.


e7p:
Can be linked to #748

PR #756, opened 2022-06-04
Matrix bridge: Print reminder to subscribe the bots on startup.

rht:
This problem has happened far too frequently. See e.g. https://chat.zulip.org/#narrow/stream/9-issues/topic/.E2.9C.94.20Matrix.20Zulip.20integration.20error/near/1388040.

### DIFF
--- a/zulip/integrations/bridge_with_matrix/matrix_bridge.py
+++ b/zulip/integrations/bridge_with_matrix/matrix_bridge.py
@@ -317,6 +317,11 @@ def main() -> None:
     zulip_config = config["zulip"]
     matrix_config = config["matrix"]
 
+    print(
+        "IMPORTANT: Make sure that the bot accounts have been"
+        " subscribed to the relevant Matrix room / Zulip stream"
+    )
+
     # Initiate clients
     backoff = zulip.RandomExponentialBackoff(timeout_success_equivalent=300)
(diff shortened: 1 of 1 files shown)


dehnert:
From the [README](https://github.com/zulip/python-zulip-api/tree/main/zulip/integrations/bridge_with_matrix), it sounds like configuring the room and stream is required. Would it make sense to grab subscriptions from Matrix (if possible) and Zulip and ensure the expected room+stream are present? Or even do the subscription in the code? (Not sure if that'd really be desirable.)

An unconditional warning still seems maybe useful, but warning only when the config seems wrong is probably more useful (at least sometimes I'd expect people to *think* they've subscribed, but not actually done so).

rht:
#723 already does the auto-subscription. I'm deciding not to spend more time into the existing Matrix bridge because it is going to be replaced by #723 anyway.

> From the [README](https://github.com/zulip/python-zulip-api/tree/main/zulip/integrations/bridge_with_matrix), it sounds like configuring the room and stream is required

But looks like people still miss the description.

timabbott:
Merging this, since it seems like a useful stopgap pending #723 being completed. (And it's probably not worth spending a bunch of time making the error handling details here nicer given that plan).

PR #765, opened 2022-08-25
integrations: Enhanced matrix bridge. (Fix #723)

ro-i:
Hi :) This PR aims to fix #723 for Windows.
As discussed there, python-magic seems to be problematic on Windows. However, I noticed in the output of the CI that `python-magic-bin` doesn't get installed as requested. (It's needed to provide the Windows library support as described [here](https://pypi.org/project/python-magic/)). Thus, my first attempt to fix the Windows support is trying to get `python-magic-bin` properly installed. If this doesn't work, I disable the magic-support in Windows completely.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,3 +15,4 @@ types-python-dateutil
 types-pytz
 types-requests
 gitlint>=0.13.0
+-r ./zulip/integrations/bridge_with_matrix/requirements.txt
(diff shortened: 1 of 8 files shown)


ro-i:
Ok, this does not seem to work. Unfortunately without useful error message in the logs :/
So I'll completely disable python-magic on Windows...

ro-i:
The Windows tests now succeed, just the code coverage is not yet happy...

rht:
Can you highlight the change you made that makes it work on Windows? I couldn't spot any difference. The requirements.txt looks the same.

rht:
Strange. The coverage should have been identical to https://github.com/zulip/python-zulip-api/pull/723/commits/12b5e7a2507bcac9c6b6db794158c03ef7ff7173.

timabbott:
Merged, thanks @ro-i!

PR #767, opened 2022-10-16
api: Update add_realm_filter to support url_template.

PIG208:
This adds support to adding linkifiers with the new URL template syntax.

This PR relies on https://github.com/zulip/zulip/pull/23125 to determine the feature level that changes the endpoint.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -1040,20 +1040,24 @@ class Client:
             method="GET",
         )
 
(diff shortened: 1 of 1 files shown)


timabbott:
I'm not sure it's particularly coherent to have the same function pass different formats, but ultimately this is an administrative API and the user will likely be looking at the Zulip settings UI when using it, so I think it's fine in practice. (And better than, say, writing code to convert a URL template into a format string).

PR #780, opened 2023-05-22
Fix recent linter errors

PIG208:
Newer pull requests are getting linter errors to reformat files due to updates in the upstream.
This reformats the affected files using black, and pin it in `requirements.txt`; fixes a mypy error for a missing return type annotation.
However, I think we need to look into options like Poetry to do dependencies versioning properly in future follow-ups.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 crayons
 twine
-black
+black==23.3.0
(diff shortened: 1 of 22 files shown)


timabbott:
Merged, thanks @PIG208!

PR #781, opened 2023-05-27
Do not send heartbeat events to clients.

showell:
I tested this with https://github.com/showell/zulip-api-examples/blob/main/heartbeat_bug.py.

See https://chat.zulip.org/#narrow/stream/137-feedback/topic/api.20client.20silent.20failure for more context.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -819,6 +819,15 @@ class Client:
 
             for event in res["events"]:
                 last_event_id = max(last_event_id, int(event["id"]))
+
+                if event["type"] == "heartbeat":
+                    # Heartbeat events are sent to clients regardless
+                    # of the client's requested event types, and are
+                    # intended to be an internal part of the Zulip
+                    # longpolling protocol, not something that clients
+                    # need to handle.
+                    continue
+
                 callback(event)
 
     def call_on_each_message(


zulip/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+## [0.9.0] - 2023-05-26
+## Changed
+- Do not send heartbeat events to clients.
```
timabbott, line 8: Let's do the release in a separate commit; see fec8cc50c42f04c678a0318f60a780d55e8f382b for an example.

zulip/zulip/__init__.py
```diff
@@ -823,7 +823,8 @@ def do_register() -> Tuple[str, int]:
 
             for event in res["events"]:
                 last_event_id = max(last_event_id, int(event["id"]))
-                callback(event)
+                if event["type"] != "heartbeat":
+                    callback(event)
```
timabbott, line 827: I think I'd prefer an early return style -- doing a `continue` if it is that type; it'll be nicer when this loop inevitably starts doing some new thing in addition to calling the callback.

zulip/CHANGELOG.md
timabbott: (Or a separate PR -- we need to actually do the pypi upload too when we do it)

timabbott:
I took care of the comment above, added some explanatory code comments, and pushed the changes back here.

PR #785, opened 2023-06-07
rss-bot: Custom topic configuration

danteu:
rss-bot selects the topic of its RSS notification messages based on the topic of the RSS feed. Monitoring a large number of RSS feeds therefore leads to a large number of topics in the stream. Also, the user has no option to customize the topic names.

This patch adds a new `--subject` argument that replaces the topic for all RSS feed notifications with the provided string. If no custom topic is provided, the bot uses the default behaviour described above.

### DIFF
--- a/zulip/integrations/rss/rss-bot
+++ b/zulip/integrations/rss/rss-bot
@@ -59,6 +59,13 @@ parser.add_argument(
     action="store",
 )
 parser.add_argument(
+    "--topic",
+    dest="topic",
+    help="A fixed topic to use for RSS messages, overriding the default of the feed title/URL",
+    default=None,
+    action="store",
+)
+parser.add_argument(
     "--data-dir",
     dest="data_dir",
     help="The directory where feed metadata is stored",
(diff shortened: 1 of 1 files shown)


timabbott:
Let's change it to `--topic`; otherwise lgtm.

(`subject` is a legacy API internals name that we don't want to expose to users as we're trying to get rid of it).

danteu:
> Let's change it to `--topic`; otherwise lgtm.
> 
> (`subject` is a legacy API internals name that we don't want to expose to users as we're trying to get rid of it).

Done. Also shortened the assignment in line 193.

timabbott:
I pushed back here after adjusting the `--help` text to update the defaults. I also wonder whether `zulip/integrations/rss/requirements.txt` is supposed to have `feedparser` in it -- I needed to install that module to run this with `--help`.

danteu:
> I also wonder whether `zulip/integrations/rss/requirements.txt` is supposed to have `feedparser` in it -- I needed to install that module to run this with `--help`.

Yes, I had to manually install `feedparser` as well. Should I add that in a commit or open a seperate PR? Other integrations such as `google-calendar` also seem to have incomplete requirement files.

timabbott:
Merged. A separate PR would be great, thanks!

PR #787, opened 2023-06-12
zulip-integrations: Add missing dependencies to requirement files.

danteu:
Adds missing items to the requirement files for the zulip integrations in `zulip/integrations/*`.


timabbott:
Merged, thanks for the contribution @danteu!

PR #793, opened 2023-10-01
IRC: Add option for SASL authentication.

rht:
This reverts to using sync IRC client, because upstream https://github.com/jaraco/irc only supports it for the sync client.


rht:
cc: @asdf8dfafjk

timabbott:
Looks good, but looks like it has a merge conflict. @andersk feel free to rebase and merge if it helps you upgrade dependencies.

rht:
Rebased.

andersk:
What’s the motivation here? Adding SASL support sounds good, but dropping non-SASL support sounds limiting?

rht:
I will add back the PLAIN authentication, so as not to restrict user's choice.

rht:
Rebased and the SASL authentication is now optional.

zulip/integrations/bridge_with_irc/irc_mirror_backend.py
```diff
@@ -1,16 +1,16 @@
...
 import sys
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 import irc.bot
 import irc.strings
 from irc.client import Event, ServerConnection, ip_numstr_to_quad
-from irc.client_aio import AioReactor
 
+logging.basicConfig(level=logging.DEBUG)
```
andersk, line 10: This looks like a temporary debugging measure that should be removed.

zulip/integrations/bridge_with_irc/requirements.txt
```diff
@@ -1 +1 @@
-irc==18.0
+irc==20.3.0
```
andersk, line 1: Should this be `>= 20.3.0` or `~= 20.3`?

zulip/integrations/bridge_with_irc/irc_mirror_backend.py
rht: This is the log when I didn't specify the SASL password to Libera.Chat. This message wouldn't show up unless I set the level to `logging.DEBUG`.
```
DEBUG:irc.client:FROM SERVER: :sodium.libera.chat NOTICE zz_zulip :*** Notice -- SASL authentication to a NickServ account wi
th a verified email address is required to connect from your current network. Please see https://libera.chat/guides/sasl for
configuration assistance.
...
DEBUG:irc.client:FROM SERVER: ERROR :Closing Link: rein.zulipdev.org (SASL authentication to a NickServ account with a verifi
ed email address is required to connect from your current network. Please see https://libera.chat/guides/sasl for configurati
on assistance.)
```

zulip/integrations/bridge_with_irc/requirements.txt
rht: I changed it to `~=20.3`.

zulip/integrations/bridge_with_irc/irc_mirror_backend.py
andersk: Well that’d be a side effect of logging [all data sent from the server](https://github.com/jaraco/irc/blob/v20.3.0/irc/client.py#L335).

Try adding an `on_error` handler to `IRCBot`.

rht: I added `on_error` that raises `Exception` for all errors. I'm not sure why there is no default `on_error` on the `IRCBot` that reports the errors.

Currently testing by running the bridge for some period of time, to see that there are errors that are recoverable.

zulip/integrations/bridge_with_irc/irc_mirror_backend.py
```diff
@@ -144,3 +144,6 @@ def on_dccchat(self, c: ServerConnection, e: Event) -> None:
             except ValueError:
                 return
             self.dcc_connect(address, port)
+
+    def on_error(self, c: ServerConnection, e: Event) -> None:
+        raise IRCError(e.target)
```
andersk, line 149: Don’t steal someone else’s exception class like this. It’s intended to be raised by the library, so its constructor interface belongs to the library and could change at the library’s discretion. For example, its constructor could gain more required arguments, like the corresponding `irc.server.IRCError` [already has](https://github.com/jaraco/irc/blob/v20.3.0/irc/server.py#L71).

andersk:
Merged without the error handling change since that needs more work and should arguably be separate anyway. Feel free to open a new PR for it.

PR #794, opened 2023-10-18
Convert type comments to Python ≥ 3.6 variable annotations

andersk:


neiljp:
This is via pyupgrade? It'd be great to see this change 👍 

Reading some of this, it reminds me it'd be great to have a typeddict for more of the interface. As a first step that'd be easiest with something like `PrivateComposition`/`StreamComposition` (as currently in zulip-terminal), extending potentially into `Message` - which is so much more complex.

PR #810, opened 2023-11-09
ruff: Fix S108 Probable insecure usage of temporary file or directory

andersk:


timabbott:
Merged, thanks @andersk!

PR #817, opened 2023-12-16
api: Add get_stream_email_address().

rsashank:
Python wrapper function for the new endpoint introduced in ZFL 226 to fetch a stream's email address. This could enhance the ease of making calls within zulip terminal and potentially other clients.

### DIFF
--- a/zulip/zulip/__init__.py
+++ b/zulip/zulip/__init__.py
@@ -1488,6 +1488,15 @@ class Client:
         """
         return self.call_endpoint(url=f"users/me/{stream_id}/topics", method="GET")
 
+    def get_stream_email_address(self, stream_id: int) -> Dict[str, Any]:
+        """
+        Example usage:
+
+        >>> client.get_stream_email_address(stream_id=1)
+        {'result': 'success', 'msg': '', 'email': 'username@example.com'}
+        """
+        return self.call_endpoint(url=f"streams/{stream_id}/email_address", method="GET")
+
     def get_user_groups(self) -> Dict[str, Any]:
         """
         Example usage:


zulip/zulip/__init__.py
```diff
@@ -1488,6 +1488,13 @@ def get_stream_topics(self, stream_id: int) -> Dict[str, Any]:
         """
         return self.call_endpoint(url=f"users/me/{stream_id}/topics", method="GET")
 
+    def get_stream_email_address(self, stream_id: int) -> Dict[str, Any]:
+        """
+        Example usage: client.get_stream_email_address(stream_id)
+        """
+        url = f"streams/{stream_id}/email_address"
+        return self.call_endpoint(url=url, method="GET")
```
neiljp, line 1496: This looks functional enough; have you manually tested this?

When contributing to a codebase it's useful to look at nearby code to match the style. My thoughts would be:
- the example doesn't use literals, which others nearby do?
- some nearby examples use a repl-like format, and show an idea of an expected successful response
- most nearby functions inline the url

A question for someone else: does this need backwards-compatibility for servers that don't have this endpoint?

rsashank: Thanks for reviewing. I've updated the PR with the given feedback. 

I have tested this manually by putting it into the ZT model. Are there alternative methods for manual testing that I might not be familiar with?

zulip/zulip/__init__.py
```diff
@@ -1488,6 +1488,15 @@ def get_stream_topics(self, stream_id: int) -> Dict[str, Any]:
         """
         return self.call_endpoint(url=f"users/me/{stream_id}/topics", method="GET")
 
+    def get_stream_email_address(self, stream_id: int) -> Dict[str, Any]:
+        """
+        Example usage:
+
+        >>> client.get_stream_email_address(stream_id=1)
+        {'result': 'success', 'msg': '', 'email': '...@chat.zulip.org'}
```
timabbott, line 1496: Please use `username@example.com` as the example email address; that's the RFC defined standard for example email addresses.

rsashank: Thanks for the review; updated the PR.

zulip/zulip/__init__.py
neiljp: Within ZT should work fine; I didn't know what kind of testing you'd performed :)

I just tested this manually in a REPL environment:
```python3
import zulip
c = zulip.client(config_file="...")
s_id = c.get_stream_id("zulip-terminal")
print(c.get_stream_email_address(s_id))
```
This produces an output which matches what I'd expect for the example in the code, and matches the default output by the web client for me.

neiljp:
@timabbott LGTM based on manual test, and given lack of old-server fallbacks and inconsistent nearby styling to compare to (see below).

It would be good to get this out, unless more API calls would be beneficial to add based on new changes/features in 7.5+/8.0?

It looks like we don't note/handle lack of features in older servers in `zulip`, ie. on an older server this call would simply fail with this call - the only use of `feature_level` is for method parameters, not the presence of the feature.

Reviewing adjacent code, we're not particularly consistent with style:
- absence/presence of responses to examples
- regular/unicode strings in example responses (eg. `u'result': u'success'` vs dropping `u`)
- example layout (blank lines, or no lines, from 'Example usage' to code)
- historically I thought we had call_endpoint calls split over lines (possibly just black?)

zulip/zulip/__init__.py
rsashank: Oh that makes sense, thanks for letting me know :)

timabbott:
Merged, thanks @rsashank!

Issue #761, opened 2022-07-30
Improve error handling when a request fails due to the persistent connection being expired 


timabbott:
Some networks drop long-lived inactive TCP connections, which can lead to connection errors when using the Zulip API bindings for some persistent jobs. We should ideally either enable a keep-alive, or perhaps a transparent retry for this class of connection error.

See https://github.com/zulip/zulip/commit/ff647dff03f178c329240c0e938d3fff8fbc5ab0 for an example.

ryansong13:
Hi @timabbott, I'm a first time contributor and I was wondering if this is this bug still open. If so, do you think it would be a good first bug to handle?

Beanpen:
@zulipbot claim

zulipbot:
Welcome to Zulip, @Beanpen! We just sent you an invite to collaborate on this repository at https://github.com/zulip/python-zulip-api/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:

- Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
- [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

Beanpen:
Hi @timabbott I wonder if you mean to remove all the alias of "client = zulip.Client(...)" and instead directly call "zulip.Client(...)" in all files. Thank you!

zulipbot:
@Beanpen We noticed that you have not made any updates to this issue or linked PRs for 10 days. Please comment here if you are still actively working on it. Otherwise, we'd appreciate a quick `@zulipbot abandon` comment so that someone else can claim this issue and continue from where you left off.

If we don't hear back, you will be automatically unassigned in 4 days. Thanks!

ryansong13:
@zulipbot claim

zulipbot:
Welcome to Zulip, @ryansong13! We just sent you an invite to collaborate on this repository at https://github.com/zulip/python-zulip-api/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:

[...]

PR #819, opened 2024-01-05
bridge_with_slack: Do not use a stale Zulip client for send_message.

rht:
This is based on https://github.com/zulip/zulip/commit/ff647dff03f178c329240c0e938d3fff8fbc5ab0, from #761.


timabbott:
Seems reasonable, let's just make sure to come back to this when we fix #761.

PR #828, opened 2024-08-13
zmirror: Drop empty zulip messages.

alexmv:
Zulip will reject sending these, so there is no need to construct them.


timabbott:
Merged, thanks @alexmv!

PR #838, opened 2024-11-21
bots: Rename BotHandler to AbstractBotHandler.

### LINKED ISSUE #690: Rename BotHandler to AbstractBotHandler.
To prevent confusion about its usage. To cement the fact that it is an abstract class used only for typing annotation purpose.
See discussion at https://chat.zulip.org/#narrow/stream/127-integrations/topic/deploying.20bots.20on.20heroku.3F.3F.3F.20--.20BotHandler.

Still need discussion: whether we should additionally use `NotImplementedError` instead of `...` to proactively tell the library users if they accidentally use the class.

Niloth-p:
Fixes #690.

### DIFF
--- a/packaged_helloworld/packaged_helloworld/packaged_helloworld.py
+++ b/packaged_helloworld/packaged_helloworld/packaged_helloworld.py
@@ -2,7 +2,7 @@
 from typing import Any, Dict
 
 import packaged_helloworld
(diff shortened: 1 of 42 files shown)


timabbott:
Merged, thanks @Niloth-p! (CI failure is present in `main`).

PR #839, opened 2024-11-21
fix: importlib-metadata as a dependency on Python>=3.10

hr1shu:
### Issue Summary
The zulip_bots package attempts to import importlib_metadata regardless of the Python version, even though importlib-metadata is not listed as a dependency in the package. This leads to an ImportError on Python versions >= 3.10, where importlib.metadata is included in the standard library, but the fallback to importlib_metadata is still attempted.

### Root Cause
The error occurs because the code in zulip_bots does not correctly handle the conditional import for importlib-metadata vs. the standard library's importlib.metadata. The setup.py configuration ensures the dependency is installed only for Python < 3.10, but the runtime code doesn't account for the difference in module naming.

### Solution
To fix the issue, I modified the code in zulip_bots/finder.py to conditionally import the appropriate module based on the Python version.

### Testing
Verified behavior on Python 3.9 and Python 3.10+ environments:
Python >= 3.10: Uses importlib.metadata.
Python < 3.10: Falls back to importlib_metadata.
Added unit tests to validate both paths with mock objects.

### Impact
This fix ensures that:

The package correctly resolves metadata imports regardless of Python version.
Dependencies are installed appropriately based on the Python version.

### DIFF
--- a/zulip_bots/zulip_bots/finder.py
+++ b/zulip_bots/zulip_bots/finder.py
@@ -2,13 +2,17 @@ import importlib
 import importlib.abc
 import importlib.util
 import os
+import sys
 from pathlib import Path
 from types import ModuleType
 from typing import Any, Optional, Tuple
 
 current_dir = os.path.dirname(os.path.abspath(__file__))
 
-import importlib_metadata as metadata
+if sys.version_info >= (3, 10):
+    from importlib.metadata import metadata  # Python 3.10+ standard library
+else:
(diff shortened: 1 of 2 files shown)


hr1shu:
Hi @timabbott, although the fix is done and it's working fine locally, can anyone help me understand why checks are failing?

timabbott:
Closing as this was fixed in a56304dc3e29efeab2ce28ef73d903b9d16a7108; @Niloth-p FYI.

PR #845, opened 2024-12-19
Added a hash to the end of the feed_file

anon1a7d:
This allows for more than one feed per domain name (netloc). Previously, only a single feed could be associated with a domain name as the feed_file kept in the cache (data-dir) was simply the domain name (netloc). Using shake128, added an 8 digit hash (4 bytes in hex) of the full URL to the end of the feed_file name.

Now can support `https://example.com/site_one/feed.rss` and `https://example.com/site_two/feed.rss`, etc.

The files in the data-dir for these two URLs will be as follows.
```
example.com-7778e2bd
example.com-52cfce4c
```

**How did you test this PR?**

Ran the modified code manually from a tmp environment (including a unique data-dir). Confirmed cached feed_files were as expected. Messages successfully sent to test stream under proper topics.

### DIFF
--- a/zulip/integrations/rss/rss-bot
+++ b/zulip/integrations/rss/rss-bot
@@ -209,7 +209,8 @@ client: zulip.Client = zulip.Client(
 first_message = True
 
 for feed_url in feed_urls:
-    feed_file = os.path.join(opts.data_dir, urllib.parse.urlparse(feed_url).netloc)  # Type: str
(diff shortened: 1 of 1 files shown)


anon1a7d:
> Is there an issue this PR is solving? Please link it if so.

There is not. It's something I noticed when I tried to use rss-bot to pull in two different feeds from a single domain.

timabbott:
@Niloth-p can you review this?

Issue #847, opened 2025-01-02
Update the Google Calendar integration


Niloth-p:
The `oauth2client` library used in the `google-calendar` [integration scripts](https://github.com/zulip/python-zulip-api/blob/main/zulip/integrations/google/) has become deprecated. So, the script needs to be updated, to replace `oauth2client` with other libraries that Google is now using.

Refer to the quickstart example in the [Google docs](https://developers.google.com/calendar/api/quickstart/python) for what to use in the place of `oauth2client`.

Make sure that you update [`requirements.txt`](https://github.com/zulip/python-zulip-api/blob/main/zulip/integrations/google/requirements.txt).

**Set Up:**
Refer to [this integration doc](https://zulip.com/integrations/doc/google-calendar#configuration-options) for instructions to set up the integration. It's currently in the process of being updated as part of [this PR](https://github.com/zulip/zulip/pull/32783), so until then you could refer to a screenshot of the updated version in that PR description.

**Bonus tasks:**
It would be nice to clean up the scripts as part of this.
And add a conventional example screenshot to the integration doc.

**References:**
For more on integrations, refer to these docs - [Integrations Overview](https://zulip.com/api/integrations-overview), and [Non-webhook integrations](https://zulip.com/api/non-webhook-integrations).

Niloth-p:
@zulipbot add "good first issue" "help wanted" "bug"

vats004:
@zulipbot claim

zulipbot:
Welcome to Zulip, @vats004! We just sent you an invite to collaborate on this repository at https://github.com/zulip/python-zulip-api/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:

- Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
- [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

theofficialvedantjoshi:
@octocamocoder47 Can I still contribute to this issue?

Niloth-p:
@theofficialvedantjoshi - Yes, the issue is still open.
[...]

PR #850, opened 2025-01-08
zulip-integrations: Update google oauth and reminders for google calendar integration.

theofficialvedantjoshi:
The `oauth2client` library used in the google-calendar [integration scripts](https://github.com/zulip/python-zulip-api/blob/main/zulip/integrations/google/) has become deprecated. This PR updates the script to replace `oauth2client` with other libraries that Google is now using, specifically `google-auth-httplib2` and `google-auth-oauthlib`.

Fixes: #847 

**How did you test this PR?**

- Generated a `client_secret.json` file from a test Google Cloud application that uses the calendar API.
- Verified the authentication process and the creation of `google-credentials.json` in the home directory.
- Tested each flag in the argument parser by specifying the hostname and port.
- Tested the no-local-sever-auth flow in the console after obtaining an authentication code from my application's redirect URL.

<details>
<summary>Self-review checklist</summary>

- [x] [Self-reviewed](https://zulip.readthedocs.io/en/latest/contributing/code-reviewing.html#how-to-review-code) the changes for clarity and maintainability
      (variable names, code reuse, readability, etc.).

Communicate decisions, questions, and potential concerns.

- [x] Explains differences from previous plans (e.g., issue description).
- [ ] Highlights technical choices and bugs encountered.
- [ ] Calls out remaining decisions and concerns.
- [ ] Automated tests verify logic where appropriate.

Individual commits are ready for review (see [commit discipline](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html)).

- [x] Each commit is a coherent idea.
- [x] Commit message(s) explain reasoning and motivation for changes.

Completed manual review and testing of the following:

- [x] Visual appearance of the changes.
- [x] Responsiveness and internationalization.
- [x] Strings and tooltips.
- [ ] End-to-end functionality of buttons, interactions and flows.
- [x] Corner cases, error conditions, and easily imagined bugs.
</details>


theofficialvedantjoshi:
@Niloth-p review. 
All the mypy tests pass when I run them locally.
image

theofficialvedantjoshi:
@timabbott could you please review this PR?

alya:
@Niloth-p It would be great to get your review on this one.

theofficialvedantjoshi:
@Niloth-p any updates?

theofficialvedantjoshi:
@timabbott @Niloth-p anything pending in this pr?

Niloth-p:
Hi, @theofficialvedantjoshi, apologies for the delay in getting back.

Thank you for the PR!
The new libraries definitely work.
Only on testing the integration, I discovered that the integration needs dozens more fixes and clean ups, apart from the library replacement, before it's ready for use. So, I have a local branch in progress, but it's unfortunately taking longer than expected, due to other priorities. I'll be absorbing this PR as part of that and give you co-author credits for this commit.

If you're interested in working more on this integration, it would be helpful if you could help improve the message being generated, by adding separate commit(s) on top of this PR. Here are some ways we can improve the message:

1. Adding more details to the messages generated would be essential. It currently just sends `<event_name> starts at <time>`, that's it. We need to be adding more details to make the integration up to par with all the other notification modes, by using the below fields that Google Calendar provides:
	- the event description
	- the `htmlLink` - to the event in the Google Calendar Web UI
	- time the event ends
	- additional info fields when present - formatted into a sentence (Use other Google Calendar notifications for inspiration):
		- `status`
		- `location`
		- `eventType`
		- `hangoutLink`
		- `organizer.displayName` (if `organizer.self` is not set, i.e., the organizer is not the owner of the calendar)
2. The script currently sends reminders only x minutes before, and x applies to all events. But, Google Calendar has a property `reminders.overrides.minutes` that tracks the duration before which the user has requested a notification for. This needs to be used whenever available.
3. The script sends reminders on every single event. Google Calendar supports `reminders.overrides` to turn off notifications for particular events. We should too.
4. Events without a title do not get a notification currently. It should instead create events with the title as "(No title)", like Google Calendar typically does.

theofficialvedantjoshi:
Hi @Niloth-p, thanks for the review.

I am interested in working more on this integration. I'll set things up locally and add the features mentioned above.
Will keep updating this thread if I have any doubts.

Niloth-p:
Glad to hear that.
Questions like these are better discussed and answered in https://chat.zulip.org/.

Contributions to this repo do count as much as any other Zulip repo.
But, have you taken a look at [the proposed project ideas](https://zulip.readthedocs.io/en/latest/outreach/gsoc.html)? In general, it would help to have some contributions in the same part of the codebase as the project you plan to propose for.

Though I anticipate being involved with any GSoC contributors working on Zulip integrations, and the set of integrations here in this repo could do with a lot of improvements, the priorities of GSoC contributions and project proposals is not my domain, the [GSoC channel](https://chat.zulip.org/#narrow/channel/14-GSoC) is a much better place to get info regarding that.

theofficialvedantjoshi:
Added the following details to the reminder messages being sent:

- time interval
- `status`
- `location`
- `hangoutLink`
- `organizer.displayName` (if `organizer.self` is not set, i.e., the organizer is not the owner of the calendar)
- `htmlLink`

image
image

theofficialvedantjoshi:
Will be working on features related to `reminder.overrides` next.

theofficialvedantjoshi:
Fixed reminder logic to respect Google Calendar's `reminders.overrides`

- Implemented event-specific reminder times by leveraging `reminders.overrides.minutes` when available.
- Added support for `reminders.useDefault`, pulling default calendar reminders if event overrides are absent.
- Introduced an `--override` option to force a uniform reminder time for all events.
- Modified event handling to prevent reminders for past events.
- Optimized event processing by sorting and iterating based on reminder time.

theofficialvedantjoshi:
@Niloth-p @alya could you please review the updated code?

alya:
@Niloth-p Tagging for your review, but I'm not following the details here, so please feel free to change the status on this PR if needed.

theofficialvedantjoshi:
@Niloth-p any updates?

zulip/integrations/google/google-calendar
```diff
@@ -40,13 +40,14 @@ class Event(TypedDict):
     description: str
     organizer: str
     hangout_link: str
+    reminder: int
```
Niloth-p, line 43: Let's add a comment explaining this field here.

zulip/integrations/google/google-calendar
```diff
@@ -40,13 +40,14 @@ class Event(TypedDict):
...
 
 
 # Our cached view of the calendar, updated periodically.
 events: List[Event] = []
 
-# Unique keys for events we've already sent, so we don't remind twice.
-sent: Set[Tuple[int, datetime.datetime]] = set()
+# Unique keys for reminders we've already sent, so we don't remind twice.
+sent: Set[Tuple[int, datetime.datetime, int]] = set()
```
Niloth-p, line 50: How about instead of adding a separate reminder field, we actually switch the start_time field in the tuple to instead use the reminder_time?

zulip/integrations/google/google-calendar
```diff
@@ -73,12 +74,11 @@ google-calendar --calendar calendarID@example.calendar.google.com
 
 
 parser.add_argument(
-    "--interval",
```
Niloth-p, line 65: Let's not change this.
This is not meant to be an override, this is the default value for the interval we're going to be using.
This shouldn't be a required argument, but an optional one with the default value as 30 mins.

zulip/integrations/google/google-calendar
```diff
@@ -125,13 +125,12 @@ def get_credentials() -> Credentials:
...
     service = build("calendar", "v3", credentials=creds)
-    now = datetime.datetime.now(pytz.utc).isoformat()
     feed = (
         service.events()
         .list(
             calendarId=options.calendarID,
-            timeMin=now,
-            maxResults=5,
+            timeMin=datetime.datetime.now(pytz.utc).isoformat(),
+            timeMax=datetime.datetime.now(pytz.utc).isoformat().split("T")[0] + "T23:59:59Z",
```
Niloth-p, line 133: `maxResults=5` is definitely restricting and needs to be updated for the new algorithm, but I'm not sure we want to pick only events for the same day.
For example, what if the user wants to send reminders like "7 days to go!", "3 days to go", etc. We should be able to support those too.
So, I'm not yet sure what we can do here, to avoid fetching too many events every single call, especially since we do `events.clear()` each time.

There are other options like not clearing our cache, and getting notifications from Google, on the calendar being updated instead. But again, I haven't looked into the feasibility of it, and whether that's necessary or optimal. And that would warrant more commits.

zulip/integrations/google/google-calendar
```diff
@@ -162,6 +161,9 @@ def populate_events() -> Optional[None]:
             # of the tzinfo base class.
             start = calendar_timezone.localize(start_naive)
             end = calendar_timezone.localize(end_naive)
+        now = datetime.datetime.now(tz=start.tzinfo)
+        if start < now:
```
Niloth-p, line 165: Not sure why this is required.
We're using `timeMin = now`, so why would start be < now?

zulip/integrations/google/google-calendar
```diff
@@ -176,7 +178,25 @@ def populate_events() -> Optional[None]:
             else event["organizer"].get("displayName", event["organizer"]["email"])
         )
         hangout_link = event.get("hangoutLink", "")
-        events.append(
+        reminders = event["reminders"]
+        # If the user has specified an override, we use that for all events.
```
Niloth-p, line 182: Let's actually do this instead.

Let's ignore the calendar's default reminders value.
The only value we'll use is `event["reminders"]["overrides"]["minutes"]`.

We'll use 30 mins as the default interval.
If the user wants something else, they can pass it in with --interval.

And for particular events with reminders set for x intervals, we override using the interval.

So, something like the below.
```suggestion
if event.get("reminders", {}).get("overrides"):
    for override in event["reminders"]["overrides"]:
        reminder_interval = override["minutes"]
        event["reminder"] = event["start"] - datetime.timedelta(minutes=reminder_interval)
		# Call events.append() for each override
```

Note that I'm suggesting setting `event["reminder"]` not to `event["reminders"]["overrides"]["minutes"]`, but actually the time it should be called, as a replacement for `event["start"]`.

And I know I had previously mentioned not sending notifications if it's turned off for particular events, after taking a look at the API. But, `event["reminders"]["useDefault"]` seems to return False by default, and I haven't yet figured out how to set it to True through the Google Calendar UI. So, unless there's an easy way to do that, and its commonly used (and you can figure it out), I think we can drop that, at least for now.

zulip/integrations/google/google-calendar
```diff
@@ -218,21 +240,30 @@ def event_to_message(event: Event) -> str:
...
-    now = datetime.datetime.now(tz=pytz.utc)
-
-    for event in events:
-        dt = event["start"] - now
-        if dt.days == 0 and dt.seconds < 60 * options.interval:
-            # The unique key includes the start time, because of
-            # repeating events.
-            key = (event["id"], event["start"])
+    # Sort events by the time of the reminder.
+    events.sort(
```
Niloth-p, line 246: Switching to using "reminder" instead of "start" can now simplify this to:
```suggestion
events.sort(key=lambda event: (event["reminder"]))
```

zulip/integrations/google/google-calendar
```diff
@@ -218,21 +240,30 @@ def event_to_message(event: Event) -> str:
...
+        key=lambda event: (event["start"] - datetime.timedelta(minutes=event["reminder"])),
+        reverse=True,
+    )
+    # Iterate through the events and send reminders for those whose reminder time has come or passed and remove them from the list.
+    # The instant a reminder's time is greater than the current time, we stop sending reminders and break out of the loop.
+    while len(events):
```
Niloth-p, line 252: Switching to using "reminder" instead of "start" can now simplify this to using the previous logic itself, so this section of changes wouldn't be needed.

Niloth-p (commented):
I've successfully absorbed the first 3 commits into my branch. Thanks for the work!

I haven't had the time to implement the last commit. So, I'm leaving some review comments regarding the implementation idea I have in mind.
I haven't checked for edge cases, so you'll have to build upon the suggested idea, and check for them.

zulip/integrations/google/google-calendar
theofficialvedantjoshi: Will do.

zulip/integrations/google/google-calendar
theofficialvedantjoshi: This would be better will make this change

zulip/integrations/google/google-calendar
theofficialvedantjoshi: Can you explain this argument properly? What is it supposed to signify?

zulip/integrations/google/google-calendar
theofficialvedantjoshi: We can deal with this later I guess. Will try to think of a feasible solution.

zulip/integrations/google/google-calendar
theofficialvedantjoshi: Yes makes sense will do.

zulip/integrations/google/google-calendar
theofficialvedantjoshi: Fetching events with timeMin = now also fetches events that are ongoing for some reason and we want to ignore those events since they have already started. Hence a small check to ignore these events.

Example:

- This event is ongoing:

    image

- When I print the feed of events received (`print(feed)`) I see this event popping up:

    image

Niloth-p:
Hey, thanks for adding that RefreshError commit addressing that, but that's actually not the fix for that issue. It's not the client-secret key that needs the refresh. I've added the fix for this in #856.
You could actually pull that #856 branch using `./tools/fetch-pull-request <number>`, and add your commits on top of it, and push to your own fork, and ping me, I'll update #856 with your commits when I can. That would be easier for both of us, because otherwise I would need to rebase your changes onto the PR, and there may be lots of conflicts. And it might take a while before #856 gets merged.

Also, if you're interested, could you help me add automated tests for the google calendar integration?
We should at least have tests covering the basic behavior - the tokens file being generated given a valid client secret file, and the messages being posted given event data.

theofficialvedantjoshi:
Alright, I am interested in adding automated tests Will do these once the above issues are addressed.
Just need to clarify somethings. Will I push to the branch on your fork or how does it work exactly?

theofficialvedantjoshi:
Reviewing my fix for `google.auth.exceptions.RefreshError` I understand the `client_secret` file doesn't need to be recreated instead the `google-credentials` file does. Is a valid fix just asking the user to remove the existing `google-credentials` and rerun the `get-google-credentials` ?

Niloth-p:
Both answers are already mentioned. I've added more details now.

1. "Will I push to the branch on your fork or how does it work exactly?"

> push to your own fork, and ping me

Pull the branch from #856 using the command above.
Create a new branch on top of it, and push that to your fork, and ping me here or on CZO with the link to your branch in your fork.

2. "Is a valid fix just asking the user to remove the existing google-credentials and rerun the get-google-credentials ?"

> I've added the fix for this in https://github.com/zulip/python-zulip-api/pull/856.

It's part of [this commit](https://github.com/zulip/python-zulip-api/pull/856/commits/445ee9b8c9e78d7f94d45fa496ea47bf799c135e).

3. Also, the tests would actually help both our changes get merged faster, because they'd require a less thorough review. Just letting you know about their relative importance, over adding more commits with fixes.
4. Feel free to review #856 too. It would definitely help to have another pair of eyes take a look, and check for any overlooked edge cases. But just make sure that you've gone through all the commits, commit messages, and the PR description before you raise any concerns, to ensure that they have not already been addressed somewhere.

theofficialvedantjoshi:
I have added the reminder logic to #856 # in my [branch](https://github.com/theofficialvedantjoshi/python-zulip-api/tree/review-original-856) keeping in mind the above resolutions. 

To clarify 

> But, event["reminders"]["useDefault"] seems to return False by default, and I haven't yet figured out how to set it to True through the Google Calendar UI.

There is a way to change the default by going into calendar settings and changing the Event notifications. In my original commit I had changed this default and tested this scenario.

image

Please review my changes and let me know what else needs to added. Also I still do not understand the point of `--interval` if we fetch reminders directly from the calendar why would we ever use `--interval`.

theofficialvedantjoshi:
@Niloth-p 
I have created automated tests to test the `get-google-credentials` logic - [commit](https://github.com/theofficialvedantjoshi/python-zulip-api/commit/faaa7a59ecc93a3f002964b4c0e84c4a8ca15574)

This test includes the following flows:
- Getting credentials provided a tokens file with valid credentials.
- Getting credentials provided expired credentials that must be refreshed.
- Getting credentials when there is no valid tokens file and the user must authenticate with local server.

Do review the code and let me know if any structural changes have to made in the way tests are written.
I will proceed to write tests for the main integration with mock event data.

theofficialvedantjoshi:
Thanks for the review.

Update:

- The bug fix and linting commit have been squashed.
- I had a `google-tokens.json` generated when I ran the code previously however when I ran the code after a couple of days with the same tokens the `creds.refresh(Request())` raised a refresh error so I had to delete the tokens and generate new ones from the script.
- Have removed docstrings and used minimal comments.
- Older commits had fixed intermediate variables.
- Can you elaborate on the tests? Mock credentials will be used to test the main message logic in `google-calendar`. To test the generation of credentials I had created mock tokens and client secrets. I will add the running of tests on running the `get-google-credentials` directly. Also could you elaborate what scenarios would require to be tested in the `google-calendar`. Should I create a dummy events payload and add that as a fixture to test different cases?

PR #903, opened 2026-02-18
chore: remove unused images from EncryptBot

RoyalOughtness:
This removes unused images from EncryptBot that take up about 1MB of space in the bundled library.


RoyalOughtness:
Closing in favor of excluding all of these from the project

